PR #17, opened 2019-11-28
Re-adds whitelist and changes an interface name to remove errors.

ProbablyCarl:


ProbablyCarl:
Travis gives me literal conniptions.

PR #31, opened 2019-11-30
3x Message/Emote Papper and Book Lenths 

Trilbyspaceclone:
## What this does

Messages are now 3078 from 1024 max
Papers are now 9478 from 6144
Books are now 27696 from 18432

## Why
People are hitting the cap really quickly and are having to use 3 emotes at a time. This should fix that


Trilbyspaceclone:
Triv is failing things I have not touched

PR #36, opened 2019-12-01
Fixs not all for me this time

Armyguy:
Makes it so any race can play ghost rolls outside the race locked ones
And makes it so you can pick armalis


StellarTheAwoo:
Looks good. PR is a fix to an issue that was needing addressed.

PR #70, opened 2019-12-07
Reverts bay no.26841

Trilbyspaceclone:
Reverts Bays PR


Trilbyspaceclone:
the revert is line to line pared with old files and commits. Its failer to compile is beyond me

PR #74, opened 2019-12-08
unfucks collectors

Andromeda-K22:
🆑 Unfucks collectors and removes the idiotic radiation overload feature. 🆑


Armyguy:
I have yet to play engi at all but i still +1 one this shit

PR #86, opened 2019-12-10
alters security skills default max level and increases cost

Andromeda-K22:
🆑 Combat Skills have had their max level for all jobs raised to 'Experienced', and cost has been accordingly increased. 🆑


StellarTheAwoo:
Considering this was talked about yesterday and we have a few players that actively play as marine ranks and would love this change in the game, I will go ahead and merge this. Thanks for the edit, by the way!

PR #102, opened 2019-12-12
Antiboarding Turret Changes

Andromeda-K22:
🆑 Antiboarding turret control implemented - works via ID tag and not area so we can avoid segmenting a single area to control all the turrets. 🆑


ProbablyCarl:
Suppose I'll have to go through and remove the areas.

PR #103, opened 2019-12-12
[SHITPOST] *oorah, but its more diverse than liberal's wet dream

NopemanMcHalt:
:cl: Anonymous
wip: Renamed SMC into UMCO, so controversial it also allows most known non-humans and modified humans in.
/:cl:

I just wanna die and be reborn as the fucking beautiful butterlord of Swadia.


Andromeda-K22:
speedmerge this

NopemanMcHalt:
tbh I would add Unathi OG under recruitables but I'm too incompetent ~~right now~~ almost always to even slightly modify OG army utility, service and dress of SCG for 'em.

PR #104, opened 2019-12-12
Amalgamation of terror

ProbablyCarl:
- - -
Tweaks:
 - Anti-Boarding turrets now controlled via the previously pushed controller by the wonderful Mazian.
 - Permabrig now has a much more defined outer lobby.
 - Yacht no longer has hostile mobs aboard.
- - -
Bugfix:
 - Yacht Solars no longer has a broken controller spot to bait new players.


ProbablyCarl:
Some additional edits.
- - -
 - Tweaks:
 - The Maintenance Jack has had its force raised from 17 to a whopping 24. This is to represent the fact that it's little more than a glorified sledgehammer with a pry-bar handle.
 - Emergency Forcing Tools now function similarly to ghetto tomahawks. They've had their throwing range reduced by one, but the force has been increased, and they're now considered sharp objects.

ProbablyCarl:
Further edits.
- - -
Tweaks:
 - Anti-Boarding control given a hyper APC.
- - -
Bugfix:
 - All turrets are given a continuous area so as to make Travis stop having a stroke. Gameplay wise, this means they all run off the local area grid that they're attached to.
 Amalgamation_of_Terror (#104)

PR #119, opened 2019-12-14
A fix for sergals dieing

Armyguy:
Aka makes it so Neural laces work


ProbablyCarl:
This affects all species with the no mechanical flag. 
It's known, and being addressed.

PR #125, opened 2019-12-14
adds space to the engine room, also runs mapmerge

Andromeda-K22:
🆑 see title 🆑 Massive diffs because I ran mapmerge.


Andromeda-K22:
wee conflict

PR #153, opened 2019-12-17
Adds fluid logging

Trilbyspaceclone:
Huh This butten just makes water and I can press it a bit!
Why isnt this logged in the first place for timing and other legit debug stuff.
Makes cat accountable when flooding bathrooms as a joke.


Trilbyspaceclone:
It was a good meme flooding xenobio cells, and bathrooms that auto drain

PR #167, opened 2019-12-18
Certain. CERTAIN somebody complained they cant be SEA just because they are not Human. What the fuck is happening with this life?

NopemanMcHalt:


ProbablyCarl:
This will be a temporary merge, unless it turns out to be proper. 
We'll see.

PR #187, opened 2019-12-21
Fixes some travis stuff partially

kevinz000:
**Red X's on Travis are a pet peeve so here we go.**

Fixes a bunch of small unit tests

Makes global access not auto-normalized by git. Reason for this is because it's checked for md5 hash, and the difference between LF and CRLF was setting off the unit test fail case.

Changes buncha crap to on_update_icon per baystation standard.

Changes color hex in vorestation crap to obey the "do not uppercase" standard

There's far more unit tests but the others involve runtime/etc which take a whole lot more effort, so I'm going to start with the find/replace ones for now since I'm doing this out of boredom anyways.

edit: haha nvm, global access is still broken even with the LF/CRLF change, whoop, forget the .gitattributes stuff it's something else probably unless I didn't specify the attribute right.


kevinz000:
also i'm gonna be honest idk why you guys keep uncompiled vore code unless you're planning to implement them but whatever yeah
reason i say this is because bay/most unit tests check every file and uncompiled code is honestly dead weight but idk y'all might find a use for it so i'm not going to pr removals or anything

kevinz000:
also another alternative for the "EXPECTED <X> OF Y" failures is just modify it to "expect" the new values in your codebase, considering you're just starting out and I'd suspect the tests are there to discourage people from making *more* of specific things that it's checking for, considering bay didn't exactly purge every unnecessary spawn() either.

ProbablyCarl:
Absolute mad lad.

PR #201, opened 2019-12-25
Adds in collars ported from Vore. As well as adds them to the accesso…

lukevale:
Adds in collars, as well as adds them to the loadout


code/modules/clothing/under/accessories/collars.dm
```diff
@@ -0,0 +1,244 @@
+/obj/item/clothing/accessory/collar
+	name = "you should not see this"
+	slot_flags = SLOT_TIE
+	var/writtenon = 0
+
+/obj/item/clothing/accessory/collar/silver
+	name = "Silver tag collar"
+	desc = "A collar for your little pets... or the big ones."
```
Andromeda-K22: Remove description, move to accessory/collar base type.

code/modules/clothing/under/accessories/collars.dm
```diff
@@ -0,0 +1,244 @@
...
+	icon_state = "collar_blk"
+	item_state = "collar_blk_overlay"
+	overlay_state = "collar_blk_overlay"
+
+/obj/item/clothing/accessory/collar/gold
+	name = "Golden tag collar"
+	desc = "A collar for your little pets... or the big ones."
```
Andromeda-K22: same for this one.

lukevale: Why move I to the base type?

code/modules/clothing/under/accessories/collars.dm
```diff
@@ -0,0 +1,244 @@
...
+	if(!jingled)
+		usr.audible_message("[usr] jingles the [src]'s bell.")
+		jingled = 1
+		addtimer(CALLBACK(src, .proc/jingledreset), 50)
+	return
+
+/obj/item/clothing/accessory/collar/bell/proc/jingledreset()
+		jingled = 0
+
+/* Disabled untill ticker can be replaced withsomething that works.
```
Andromeda-K22: Remove commented out code. We're not adding shock collars anyway.

lukevale: Will have to wait until I'm not on moble. Only reason it's still there.

code/modules/clothing/under/accessories/collars.dm
Andromeda-K22: Because there's no reason to have the desc a dozen times when you can just overwrite it when you need to.

lukevale: Ie like how every single collar literally already does?

Andromeda-K22: Pretty much, you don't need to overwrite the desc when it's literally the same tho.

PR #228, opened 2020-01-06
Replaced stun carbines with assault riffles in the emergency armory.

Wilenia:


ProbablyCarl:
why

Wilenia:
The suggestion was that if the emergency armory is being used then you want lethals. Stun guns are not lethals.

PR #267, opened 2020-01-16
[513 ONLY] CTRL-H "lentext" into "length"

NopemanMcHalt:
Since it's being phased out, for some reason. Compiles just fine, not sure about in-game effects.


NopemanMcHalt:
Well apparently #489 can do this better, even on 512 so uhhhh

### **FUCK**

PR #279, opened 2020-01-19
Makes changes to the Compact Smartgun

2cwldys:
- - -
Tweaks:
- Makes the Compact Smartgun pertain the ability to recharge itself.
- Adds a recharge time of Five Seconds for a single charge.

Future Tweaks, or Edits: [[1]](https://github.com/BoHBranch/BoH-Bay/pull/279/commits/39f8a7bb7d55bd54da840de24475933545670c02)
- Adds a recharge time of Eight Seconds for a single charge.
- The compact smartgun now has a maximum of three shots. Use them wisely inbetween recharges.


ProbablyCarl:
Standard munition count for the small smartgun sits at 5.
Perhaps lower it for the secure compact, given that it has the ability to recharge? Or remove the lethal capability. Or something of the sort, anyhow.

The value: 	max_shots = 5

PR #291, opened 2020-01-21
Medbay remap

zaredman:
Mostly a port of https://github.com/Baystation12/Baystation12/pull/27846

~~I have no clue why it's adding 76k lines. Tried this like six times, each one with map merger.~~
Edit: Looks like mapmerge completely reformatted _deck1.dmm. BS12 version is 72405 lines while current Hestia version is just over 3000 lines. Go figure. Compiles fine on my end.

The overall goals of this: I want people to have space to interact without constantly trampling each-other. I've attempted to make the layout reasonably central and direct with points that people would congregate at while trying to minimize voids that would separate people from line-of-sight and just overall isolate individuals with no other options.

This remap has also taken into consideration where medical players, historically, go. If there are no calls, they typically chill in the lobby or immediately adjacent to. Bringing Chem closer into the mix helps to avoid isolating them.

ETC was designed with a bit more of a focus on space to work with the most frequently used bits either closer to what they are used with or in the center of it all. OR1/OR2 have direct access off of the ETC

Staff wing (lockers/storage) really haven't been a major place that people hang out or run into each other for more than a few minutes at a time so that section has been mapped with more of a "utility" mindset in mind. And the ward gives people the option to be social or isolate private time following xyz trauma.

This PR does not make any adjustments to the Psych's office or Robotics. Hestia Flavored medbay remap also introduces a Resuscitation Bay for all your cloning needs as well as a staff lounge by the lobby.


ProbablyCarl:
While it looks fantastic as is, map diff shows some incredibly dangerous issues. 
In the pod north, there's a tile of space, and there's a missing door / window just north of the entrance.
Map diff usually hides things, so I might be incorrect, but I just want confirmation.

zaredman:
@ProbablyCarl missing tile by entrance is a big nanomed. Mapdiff doesn't like rendering those for some reason. Space tile's fixed in a second.

PR #310, opened 2020-01-24
Fixes Sabot shell holders

Knightfall5:
This should allow you to insert Sabot shells into their intended shellholder, letting you refill them or stick a shell back in if you accidentally remove it.


ProbablyCarl:
This slipped past me in my sleepy haze.
I appreciate you catching this - thanks.

PR #328, opened 2020-01-31
Fixes Travis

MarinaGryphon:
Big DNM, it's a WIP at the moment.
**Important notice: DO NOT USE `foo << "message to be displayed"`. Use `to_chat(foo, "message to be displayed")` instead. Also, a blind find-and-replace WILL NOT work to fix that, as that would screw up things like `browse()`, `link()`, `output()`, etc.**


MarinaGryphon:
WIP will be over as soon as Travis successfully finishes. At that point it can be merged because any longer and it'd fuck everything up with conflicts again.

PR #332, opened 2020-02-01
All hairstyles can be selected by both genders.

BurgerLUA:
Given the fact that you can be a super snowflake with genemods and whatever the heck, it doesn't make sense that I am not allowed to be a bald female human. This PR makes it so that gender doesn't matter anymore when you can select a hairstyle.


Keekenox:
Wait that's actually a thing on bay? What the fuck bay.

PR #336, opened 2020-02-01
Removes Ascent Whitelists

LNSmith458:
Removes the whitelist for Ascent Gyne and Ascent Alate

:cl: Wastedfate
tag: Ascent Gyne and Alate are no longer whitelisted.
/:cl:


Keekenox:
When am I going to get a big tiddy gyne gf?

code/modules/species/mantid/mantid.dm
```diff
@@ -210,7 +210,7 @@
 /datum/species/nabber/monarch
 	name = SPECIES_MONARCH_WORKER
 	name_plural = "Monarch Serpentid Workers"
-	spawn_flags = SPECIES_IS_RESTRICTED | SPECIES_NO_FBP_CONSTRUCTION | SPECIES_NO_FBP_CHARGEN
```
ProbablyCarl: Restricted doesn't impose a whitelist.
This is also a Monarch, not the Gyne.

BASIC-BIT: I think SPECIES_IS_RESTRICTED might be the correct thing to remove.  It's checked [in whitelist.dm](https://github.com/BoHBranch/BoH-Bay/blob/dev/code/game/jobs/whitelist.dm#L73)

LNSmith458:
Further testing shows that initial edit was correct, but also removes whitelist for Monarch. Further edits need to be made to keep Monarch whitelisted, but not Gyne.

PR #339, opened 2020-02-02
Overhauls how chemicals interact with plants

BurgerLUA:
This is a template PR to show people what I'm doing.

This basically reworks how chemicals interact with xenobotany, and adds some new ones that can help make some crazy plants and whatnot, including adding missing potency modifiers. More chemicals to come.


BurgerLUA:
Don't merge this yet. I need to test it with the latest upstream merge.

BurgerLUA:
It works fully, now. No errors from what I did with testing. The only issue is the travis no passing, but that's because of a map issue in dev and not a result of my PR.

PR #343, opened 2020-02-03
Converts all map files from DMM to TGM

BurgerLUA:
Alternative title: Gives me 523,923 line additions.

This converts all map files from DMM to TGM. For those who aren't aware of TGM, TGM is a better format to use that spreads changes across from lines to make it easier for map contributions to be done and compared.


Jeff-Dome:
based and red-pill coder

BurgerLUA:
Will be making this another time since there are map edits and travis errors.

PR #345, opened 2020-02-04
Fixes a few things in the uplink. Makes zombie tar require 2 traitors.

Torque4607:
<CL>
Fixes some spelling, flavortext, descriptions in the uplink. Bumped zombie black tar TC cost up to 220. Meaning that two traitors must combine their telecrystals in order to purchase it. Think of a unique gimmick instead of zombifying everybody again, mm'kay?
</CL>

### DIFF
--- a/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
+++ b/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
@@ -12,7 +12,7 @@
 
 /datum/uplink_item/item/visible_weapons/smallenergy_gun
 	name = "Small Energy Gun"
-	desc = "A pocket-sized energy based sidearm with three different lethality settings."
+	desc = "A pocket-sized energy based sidearm with three different lethality settings, stun, shock, kill."
 	item_cost = 16
 	path = /obj/item/weapon/gun/energy/gun/small
 
@@ -57,7 +57,7 @@
 
 /datum/uplink_item/item/visible_weapons/energy_gun
 	name = "Energy Gun"
-	desc = "A energy based sidearm with three different lethality settings, Stun, Shock, Kill."
+	desc = "A energy based sidearm with three different lethality settings, stun, shock, kill."
 	item_cost = 32
 	path = /obj/item/weapon/gun/energy/gun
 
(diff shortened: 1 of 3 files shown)


code/datums/uplink/highly_visible_and_dangerous_weapons.dm
```diff
@@ -57,7 +57,7 @@
 
 /datum/uplink_item/item/visible_weapons/energy_gun
 	name = "Energy Gun"
-	desc = "A energy based sidearm with three different lethality settings."
+	desc = "A energy based sidearm with three different lethality settings, Stun, Shock, Kill."
```
Roland410: Capitalization seems unnecessary for the options.

code/datums/uplink/stealthy_and_inconspicuous_weapons.dm
```diff
@@ -49,5 +49,5 @@
 	name = "Vial of Black Tar"
 	desc = "Supposedly, ingesting this would cause someone to become a little less alive, though still alive. How does it make sense? \
 	Well, it doesn't. That's the fun of it!"
-	item_cost = 120
+	item_cost = 220 //Needs two traitors collaborating, and even then, it burns a lot of your TCs. Zombies are overdone, mmkay?
```
Roland410: In my opinion 180 or 200 would be a bit better, as that way they aren't limited to just the zombie gimmick.

code/datums/uplink/highly_visible_and_dangerous_weapons.dm
Torque4607: I can quickly fix that up.

code/datums/uplink/stealthy_and_inconspicuous_weapons.dm
Torque4607: The reason I believe it should be so high, at least for right now, is that it's _highly_ overdone. Every round or two, there's been a zombie outbreak. In my opinion, if traitors want to do it, they should be damn-well certain that's all they're doing is committing to that gimmick. I'll let others weigh in on whether or not to drop is down by 20-40 TC or so.

PR #347, opened 2020-02-04
Biodegrade begone

EldritchSigma:
Removes biodegradation from bomb implants. Because rolling for RNG death is such a fun and engaging mechanic when its being used to make you compliant. Or when you roll merc, get the implant to prevent capture and it biodegrades in the middle of the Desperado. Hopefully it works because otherwise Hailey will be a really sad looking GAS and that is sad.

It should work though. It removes the malfunction proc from the explosive implant which scrambles activation phrase, frequency and all that stuff.


EldritchSigma:
Eh

PR #349, opened 2020-02-05
Vox Pariahs!

Eonoc:
:cl: Eonoc
tag: Adds Vox Pariah back in as a crew species for some low level jobs.
/:cl:
Re-adds the ol' Vox Pariahs as a crew species. Due to old hatred of the spammed smell messages, I changed their stink to just be a reskinned variant of neck markings available on examine.
Otherwise, they're essentially the same as they were. Non-vacuum resistant, shittier at melee, take increased damage from everything Vox. 
The jobs they're allowed are janitor, cargo technician, and test subjec--research assistant

### DIFF
--- a/code/__defines/mobs.dm
+++ b/code/__defines/mobs.dm
@@ -291,7 +291,7 @@
 #define SPECIES_DIONA       "Diona"
 #define SPECIES_VOX         "Vox"
 #define SPECIES_VOX_ARMALIS "Vox Armalis"
-#define SPECIES_VOXPARIAH "Vox Pariah"
+#define SPECIES_VOXPARIAH 	"Vox Pariah"
 #define SPECIES_IPC         "Machine"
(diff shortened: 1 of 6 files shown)


Andromeda-K22:
The primary issue with this is that vox are treated as _persona non grata_ by the crew in most, if not all cases, and if vox raiders show up there might be an issue with ... confusion.

Eonoc:
> The primary issue with this is that vox are treated as _persona non grata_ by the crew in most, if not all cases, and if vox raiders show up there might be an issue with ... confusion.

Personally, I think it's an interesting dynamic. Just earlier we had someone pose as Vox raiders over comms, then had a manhunt after people reported sighting of a Vox that turned out to be a Vox genemod custom species. I like the dynamic, but if people don't think it's great, we can just...not merge this.

Jeff-Dome:
Good conflict

PR #367, opened 2020-02-12
Ghetto Synth Repair

InfinitelyThinRectangles:
This PR adds a few ghetto options for specific steps of synth repair. There's actually a surprising number of synth-surgery steps that DO have ghetto options, but strangely none of the really important, commonly used ones do.

With this, it's possible to use a wrench and wirecutters to repair brute and burn respectively as long as you've already opened the access panel, but at a pitiful 35% base success rate and a near-doubling of the time taken to perform the action.

:cl: InfinitelyThinRectangles
tweak: Wrenches and Wirecutters can now be used to ghetto-repair damage to robotic limbs, provided their access panel is already open. This takes /significantly/ longer than usual and has a very low base success chance, so make sure you know what you're doing!~
/:cl:

### DIFF
--- a/code/modules/surgery/organs_internal.dm
+++ b/code/modules/surgery/organs_internal.dm
@@ -137,7 +137,7 @@
 	name = "Remove internal organ"
 	allowed_tools = list(
 		/obj/item/weapon/hemostat = 100,
-		/obj/item/weapon/wirecutters = 75,
+		/obj/item/weapon/material/hatchet = 75,
 		/obj/item/weapon/material/knife = 75,
 		/obj/item/weapon/material/kitchen/utensil/fork = 20
 	)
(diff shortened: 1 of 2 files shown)


MarinaGryphon:
I'd suggest making it like necrotic organ repair, where it decreases the max health of the limb in the future. Wirecutters can't add wiring, after all, just remove and rewire them.

InfinitelyThinRectangles:
I was considering changing the desctext to reflect that you're more just trying to re-wire with what you had rather than adding more wires, and I can still do so if there's any appetite for refluffing it a little.

As for reducing the max limb HP, I think the obvious difficulty and less than ideal nature of the task is reflected in how long it takes to do it, and how low the chance of actually managing to not only have any benefit, but to /not/ cause even further damage is.

PR #369, opened 2020-02-15
Ghetto synth surgery

InfinitelyThinRectangles:
This PR adds a few ghetto options for specific steps of synth repair. There's actually a surprising number of synth-surgery steps that DO have ghetto options, but strangely none of the really important, commonly used ones do.

With this, it's possible to use a wrench and wirecutters to repair brute and burn respectively as long as you've already opened the access panel, but at a pitiful 35% base success rate and a near-doubling of the time taken to perform the action.

- Re-upped because I am a humongous chungus and PR'd into the dev branch, not the master one, and then was surprised pikachu when they weren't reflected in game. Sorry about that, lads.

🆑 InfinitelyThinRectangles
tweak: Wrenches and Wirecutters can now be used to ghetto-repair damage to robotic limbs, provided their access panel is already open. This takes /significantly/ longer than usual and has a very low base success chance, so make sure you know what you're doing!~
/🆑


modular_mithra/code/modules/mob/inventory.dm
```diff
@@ -10,7 +10,7 @@
 			return TRUE // self destroying objects (tk, grabs)
 
 		if(W.loc != target)
-			INVOKE_ASYNC(W, /atom/movable/.proc/do_putdown_animation, target, src)
```
Andromeda-K22: ?

PR #370, opened 2020-02-15
Tweaks n' fixes!

InfinitelyThinRectangles:
This PR permits the taking of the MCU Augment for anyone the company has already trusted with body armour, at the normal points cost. As such, it's available to Security and Command.

It also de-restricts the Tool-Arm augment, which makes it a competitor for limited loadout points and may help draw people away from the combat skill boosting augs by providing a viable alternative.

It also corrects an oversight that led to pistols mags that you could get from cargo not actually fitting in the pistols you could get from cargo.

- Re-Upped because I am a humongous chungus and PR'd this to dev instead of master. Please no bulli Carl-Senpai. :(

🆑 InfinitelyThinRectangles
tweak: MCU Augment made available to Security/Command Roles.
tweak: Pistol Magazines from Cargo now fit in the Military Pistols from Cargo.
tweak: Tool-Arm Augments available to everyone!
/🆑


modular_mithra/code/modules/mob/inventory.dm
```diff
@@ -10,7 +10,7 @@
 			return TRUE // self destroying objects (tk, grabs)
 
 		if(W.loc != target)
-			INVOKE_ASYNC(W, /atom/movable/.proc/do_putdown_animation, target, src)
+			addtimer(CALLBACK(W, /atom/movable/.proc/do_putdown_animation, target, src), 0)
```
Andromeda-K22: ?

InfinitelyThinRectangles:
Fixed that anomalous line, honestly not entirely sure what happened there. I know it wasn't something I changed, I presume it was changed upstream or in another PR and I hadn't updated my fork before pushing my own changes.

PR #377, opened 2020-02-17
Radio is metaprotected yet gives 30% more

EldritchSigma:
30% more TC and basicly no drawback is not really well thought out in my opinion. Why would somebody not take this thing over any other uplink option. Sure it basicly takes storage space, but considering how much more value you get out of this. This brings them in line with the other uplinks, only giving the standard 120 TC.

### DIFF
--- a/code/datums/uplink/uplink_sources.dm
+++ b/code/datums/uplink/uplink_sources.dm
@@ -82,10 +82,10 @@ GLOBAL_LIST_INIT(default_uplink_source_priority, list(
 
 /decl/uplink_source/unit
 	name = "Uplink Unit"
-	desc = "Teleports an uplink unit to your location. Has 30% more TC."
+	desc = "Teleports an uplink unit to your location. "
 
 /decl/uplink_source/unit/setup_uplink_source(var/mob/M, var/amount)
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
It receives thirty-percent more due to being incredibly easy to find, rather than code-locked and hidden like the PDA program.
I understand the reasoning, though, given that they're meta-protected.
I'll wait for Maz to weigh in, as this does actually affect antags. For example, why would someone choose this now, over the PDA program?

Andromeda-K22:
I'm against this.

EldritchSigma:
Then remove that metaprotection. Its silly that this thing basicly gets all the advantages of a normal uplink + 30% more and effectively zero downsides. Its basicly a radio yes. But pressing any button on it would activate the uplink. I do not understand where this is balanced in any way. Theres no reason to not take this uplink if it has the same metaprotection like any other uplink.

Proposed Alternatives:
- Increase TC to bring them in line with the radio uplink so its not a blatant advantage and basicly just personal prefs.
- Turn it into a stationary uplink: Use radio, stationary uplink spawns, buy gear, hit "dismiss uplink" or something else. Uplink despawns. That way the 30% TC are not essentially freebies.

As it is now its just free resources for a traitor that bothers to set their uplink prefs to uplink device.

PR #378, opened 2020-02-17
Genemodder background truly customizeable, tested locally, works

EldritchSigma:
"Fine.. I'll do it myself." -Bane Buddy standing 8'2'' blue boi.

This hopefully makes it so genemodders can select ANYTHING ELSE than just human backgrounds for a background. Since they are supposed to be genetic modifications of existing species and being a skrell thats strong but weak to drugs being forced to take human only backgrounds is kind of silly.

Might not work. Dont understand byond code to its fullest. Dont understand baycode quite. Reverse engineered this crap on an energy drink and two citaloprams that supposed to deal with crap.

Hint it did not work. Most likely like this PR.


EldritchSigma:
Heavily WIP, hope I dont need to fucking add new calls and shit to make this work.

EldritchSigma:
The "Weh" patch made it actually compile. Nice.

EldritchSigma:
Works, tested it locally, all backgrounds show up.

MarinaGryphon:
https://github.com/BoHBranch/BoH-Bay/blob/dev/code/__defines/culture.dm
This is the file with every culture name in it. I'll check to see if you've missed any but I don't think you have. This should, probably, work.
Also not high risk. Only touches one thing, doesn't touch the DB, etc.

EldritchSigma:
Anything I need to fix on this?

Andromeda-K22:
Appears fine. Merging.

PR #380, opened 2020-02-18
Teensy Weensy Dagon Skill Edit

Karl-Franz-I:
Few changes:
1. Gives CMO trained complex devices like the physician so the MO is no longer better than the CMO.
2. I heard rumors of Carl planning to touch up security officer skills, I have given them 22 skillpoints instead of the default bay 16. This will shortly be followed by the removal of combat augments from loadout PR. This can easily be changed in the future, but as it stands they are sub optimal.
3. The passenger no longer has a much larger amounts of skillpoints than they should. After discussion I feel this choice is much more preferable to the community than enacting maximum skill restrictions down to trained on combat skills like bay has. Passengers will now be more in line with the members of service crew for skill points (janitor, chef, etc) which still allows them to specialize in any of their titles' professions.
4. Crewman ranking has been reduced to denote the fact that they are in fact not more experienced than half the professionals on the ship. This has no effect on mechanics and is merely a small patch. Also due to the astronomically higher amount of players, crewman slots have been bumped from 5 to 8. Their total skillpoints has been reduced by 2

:cl:
Tweak: Lowers skillpoints of passengers and crewman back down from ridiculous, slight increase to sec pending removal of combat augs.
Tweak: Lowers possible rank of crewman as they were misvalued.
/:cl:

### DIFF
--- a/maps/torch/job/command_jobs.dm
+++ b/maps/torch/job/command_jobs.dm
@@ -161,7 +161,8 @@
 	min_skill = list(   SKILL_BUREAUCRACY = SKILL_BASIC,
 	                    SKILL_MEDICAL     = SKILL_EXPERT,
 	                    SKILL_ANATOMY     = SKILL_EXPERT,
-	                    SKILL_CHEMISTRY   = SKILL_BASIC)
+	                    SKILL_CHEMISTRY   = SKILL_BASIC,
+						SKILL_DEVICES     = SKILL_ADEPT)
 
 	max_skill = list(   SKILL_MEDICAL     = SKILL_MAX,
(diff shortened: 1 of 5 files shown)


ProbablyCarl:
IIRC, I touched on Security skills in a recent PR, and buffed them by quite a bit.
I could be mistaken, but if that's the case, just ignore my statement here entirely.

Karl-Franz-I:
I'm on the server right now and looking at sec, they are still identical to bay security.

Andromeda-K22:
Allow security officers to take Experienced ranged/CQC and I'l merge this, given your augment removal.

PR #383, opened 2020-02-18
PR that could split space, time, and the community (Augments)

Karl-Franz-I:
Hear me now, brave men - brave Hestialanders - hear me! The Antag Wizards read the portents true. Shitsec/LRPers land on Dagon shores. We have maneuvered many miles, but there is no distance so great that I would not gladly go to face this - our ultimate foe! I have refused the honor of wielding the coding power of our forebears thus far. I was not worthy. But should we survive this PR, should we drive the min maxing tribes back across the Byond Hub - then, with your blessing - I will pick up the non-powergame security role and use it to bleed our foes - to carve my way through Hestia's enemies! Are you with me, Men of the Spess? For when the LRP trash come, they will attack hard, as their despicable gods demand. We must be ready to push them back, thrust them into the void in defense of the Dagon!

### tl;dr
I am removing the essentially free combat augments at roundstart for a good chunk of roles as, in my humble and 17% retarded opinion, it was never at any point a good idea to add significant advantages to a mostly cosmetic/minor influences loadout menu.

I am open to suggestions for (existing) augments to add to the augment category of the loadout that aren't so...impactful.

:cl:
rscdel: Removes unbalanced and significantly impactful combat augments that you get for unimportant loadout points at roundstart. 
/:cl:

### DIFF
--- a/maps/torch/loadout/loadout_augments.dm
+++ b/maps/torch/loadout/loadout_augments.dm
@@ -1,18 +1,3 @@
-/datum/gear/augmentation/mil
-	display_name = "tactical computer"
-	path = /obj/item/organ/internal/augment/boost/shooting
-	cost = 6
-	allowed_branches = MILITARY_BRANCHES
-
-/datum/gear/augmentation/mil/cqc_booster
-	display_name = "close combat reflex booster"
-	path = /obj/item/organ/internal/augment/boost/reflex
-	cost = 8
-
-/datum/gear/augmentation/mil/subdural_armor
-	display_name = "subdural armor"
-	path = /obj/item/organ/internal/augment/armor
-
 /datum/gear/augmentation/implanted_surgical
 	display_name = "surgical polytool - left arm (ROBOTIC)"
 	path = /obj/item/organ/internal/augment/active/polytool/surgical/left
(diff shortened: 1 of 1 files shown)


Roland410:
Being able to have loadout augments was a complete mistake in my opinion anyways. Being able to spawn with a surgical toolset for example and the only drawback being that your hand has to be robotic is a bit OP by itself. Same thing with the toolkit, but tools are common enough to not be such a big thing most of the time, and yes, I'm aware that they are loosely job-locked, but that doesn't change much.

Karl-Franz-I:
Fun fact this only removes the combat augments and the ones that increase combat skills, not the polytools. I for one would not be against removing the entirety of the augment loadout section, but that's not something people can get behind

Andromeda-K22:
Subdural and nanite MCU are balanced enough that they can probably stay.

Roland410:
Robotics can make them in-round, I don't see a reason to give sec even more of a roundstart advantage considering all the weaponry and shit they have access to.

Andromeda-K22:
Robotics never exists. Let's be real.

Research hardly ever does their job.

Roland410:
That is such a round then. It's quite a bit of advantage they get, while having high combat skills, pretty powerful weapons and good armor.

Andromeda-K22:
Security has to contend with some pretty powerful antags like the Ascent and Vox being shitters fairly often. It irks me that an MAA can't take experienced in CQC/Weapons Training because ??? Only the HoS or Brig Chief can.

Roland410:
http://prntscr.com/r4719j
Also the other PR apparently gives sec guards more skill points as well.

Sparrowbird611:
Really good PR. I agree that these shouldn't be roundstart items.

PR #386, opened 2020-02-18
Removes point buffs and nerfs from age.

Sparrowbird611:
What is does:

This PR removes the point removal or gain from age. All players gain eight points, regardless of age.

Why it's good for the game:

Tying points to age really only hurt people who played in the system. You could easily powergame it by simply increasing your age. Now, age is fully cosmetic, so you won't be losing out because you chose to play a slightly younger character.

🆑

Tweak: Removes point gains and losses from age. If you already used your max age points, this won't effect you at all. If you don't have the maximum age points, you get a few extra points to spend.

/🆑


Roland410:
It'd probably be more elegant to just set the point itself, also if you remove the species overrides, then even cleaner.

PR #394, opened 2020-02-19
VueUI port, modular edition

MarinaGryphon:
Same as in #365, but modular this time. Prerequisite for Aurora cooking.


MarinaGryphon:
Also, this isn't actually user-facing at all. That comes later.

Andromeda-K22:
Ready for merge?

PR #398, opened 2020-02-19
Reverts Crewman Rank Changes

Difrit:
Reverts removal of certain Crewman ranks snuck into [this PR](https://github.com/BoHBranch/BoH-Bay/pull/380). Putting multiple unrelated changes in the same PR isn't really that good conduct, and this snuck by without any feedback as a result. These crewman ranks allow people to play certain characters as Crewman outside of the department, if the usual job slot is full.

You, sneaking unrelated changes into the same PR
farquaad

Me, fixing your bad conduct
charming

🆑
Edit: Reverts Crewman rank changes snuck into a seperate PR.
/🆑

### DIFF
--- a/maps/torch/job/torch_jobs_boh.dm
+++ b/maps/torch/job/torch_jobs_boh.dm
@@ -393,10 +393,14 @@
 		/datum/mil_rank/fleet/e2_exp,
 		/datum/mil_rank/fleet/e3,
 		/datum/mil_rank/fleet/e4,
+		/datum/mil_rank/fleet/e5,
+		/datum/mil_rank/fleet/e6,
 		/datum/mil_rank/marine_corps/e1,
 		/datum/mil_rank/marine_corps/e2,
 		/datum/mil_rank/marine_corps/e3,
-		/datum/mil_rank/marine_corps/e4
+		/datum/mil_rank/marine_corps/e4,
+		/datum/mil_rank/marine_corps/e5,
+		/datum/mil_rank/marine_corps/e6
 	)
 /***/
 


Karl-Franz-I:
4. Crewman ranking has been reduced to denote the fact that they are in fact not more experienced than half the professionals on the ship. This has no effect on mechanics and is merely a small patch.
Duolingo offers good courses for English if you need it.

Difrit:
Emphasis on snuck into an unrelated PR. That's like me putting more blood bags in Medbay but also removing the Destructive Analyser from Science, doesn't matter if I mention it in the PR or not, it's bad conduct. Don't get snippy.

Karl-Franz-I:
Apologies for being snippy but I believe it’s an apt response when someone plays up malicious intent which still was not snuck in anywhere. It wasn’t a rider, it was also discussed for about an hour before I made final changes to crewman. Your example really isn’t fitting considering the difference of severity and how it indeed was related because the PR was of generalized job touch ups. I realized my title was incorrect, it should have stated skill and rank edits as my original branch did. If there was other ranks that needed changing it would have also been put in and clearly stated. Everything of substance regarding the job’s role on the server and how the change is appropriate for the job I said was ignored so I’m assuming it’s because this is github and not discord. We can always talk about the crewman role later if we’re not busy, as it’s my opinion that your reasoning behind the crewman revert differs from its actual responsibility in the round.

GrayRachnid:
I don't think it was 'snuck' into the PR. It was mentioned quite visibly and it was merged. This revert PR doesn't give a resounding argument as to why they should be re-added either, non-assigned crewman shouldn't be outranking specialized crewman.

Difrit:
Closing PR due to a similar PR being opened. Will relook at this further down the line.

PR #402, opened 2020-02-20
Improves the Firelocks, Part 1

BurgerLUA:
Firelocks no longer act and behave like shit. It will accurately tell you which areas are subject to danger and flash accordingly.

image

Reference:
Blue flashing lights means the area in that direction is cold as shit.
Orange flashing lights mean the area in that direction is hot as shit.
Yellow flashing lights means that if the airlock is opened, you will fall over and be knocked around.

Lights can be combined as well.


Keekenox:
NICE
Finally the engoodening of bay begins!

BurgerLUA:
DON'T MERGE THIS YET

BurgerLUA:
It's ready to be merged now.

PR #405, opened 2020-02-20
Round time is based on REAL time instead of time adjusted for lag.

BurgerLUA:
Draft PR.
Title is self-explanatory.

Need to test this so it's a draft.


Difrit:
May present minor non-code issues with things based on in-game timers, like the nuke, brig timers or jump/evac taking longer than the real-time timer. Worth checking how this impacts those.

GrayRachnid:
> May present minor non-code issues with things based on in-game timers, like the nuke, brig timers or jump/evac taking longer than the real-time timer. Worth checking how this impacts those.

This only impacts the vote.

BurgerLUA:
It actually impacts a few other things that I need to list here once I start testing the PR and looking at the code.

BurgerLUA:
Works fine here. The change only affects round time now.

PR #407, opened 2020-02-20
Various Rank Changes

Difrit:
**Some issues arose with my fork, going to delete the fork and recreate rather than hassle with fixing it. Comments have been noted, contractorsec section of this PR will be removed, will add O-2 Pathfinder.**

Allows XO to be O-4, as Bridge Officers have been restricted from higher ranks which allows for some variation.

Allows RD to be O-3, in keeping with other Head ranks.

Allows Crewman to be E-1, not sure why this was removed when other crew can be E-1.

Removes Contractors from Security, due to rampant misuse of the title. This is a bit of a sticky issue and I'm happy to take feedback regarding this, but the majority of Contractor Security grossly misuse the role as "I'm not in the military, you can't tell me what to do, I'm going to validhunt/by edgy", with most of the time it not being particularly ahelpable. I'd rather have the possibility of that gone until similar behaviour has settled down. This impacts the few players who play the Contractor side well, but it also means no more red flag PCRC/SAARE.

🆑
Tweak: Various joinable rank changes.
/🆑

### DIFF
--- a/code/modules/boh_misc/boxes.dm
+++ b/code/modules/boh_misc/boxes.dm
@@ -75,7 +75,7 @@
 
 /obj/item/gunboxsmall/attack_self(mob/living/user)
 	var/list/options = list()
-	options["Ballistic"] = list(/obj/item/weapon/gun/projectile/pistol/military/alt/solar,/obj/item/ammo_magazine/pistol/double)
+	options["Ballistic"] = list(/obj/item/weapon/gun/projectile/pistol/military/alt/solar,/obj/item/ammo_magazine/pistol/double/rubber)
 	options["Energy"] = list(/obj/item/weapon/gun/energy/gun/secure,/obj/item/weapon/grenade/empgrenade/low_yield)
 	var/choice = input(user,"What type of equipment?") as null|anything in options
 	if(src && choice)
--- a/maps/torch/job/torch_jobs_boh.dm
+++ b/maps/torch/job/torch_jobs_boh.dm
@@ -16,7 +16,9 @@
 		/datum/mil_branch/marine_corps = /decl/hierarchy/outfit/job/torch/crew/command/XO/marine
 	)
 	allowed_ranks = list(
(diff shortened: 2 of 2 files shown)


Sparrowbird611:
My personal opinion on the contractor thing. I see no reason to remove the contractor options from security. While I am aware there are contractors who misuse the title, I feel that this issue should be resolved by simply ahelping the players, not by removing the role. I would be incredibly surprised if removing the contractor role improved these players, instead of just having them walk off to a different rank, and being NTEF or Marine shitlers instead of contractor ones.

TL:DR Keep contractor.

Roland410:
Gonna agree with Sparrow on the contractor thing. It'll just push the people who already are shitters as contractors into being shitters under a different title/rank, while punishing legit people who want to play something different.

Difrit:
> By the way, could really expand on the changelog, as it means next to nothing.
Planned on doing so after the resolution of the Contractor feedback.

> Also don't see the point in allowing XO's to be O-4, but I couldn't care less, it's just easier to know if they are the commander or in the CO's case, captain, and not be like LCMDR.
With Bridge Officers now being dropped a few ranks, it means there's some progression for XOs that want that, as well as some variation in not having a LtCol for the seventeenth round in a row. Role > Rank

OolongCow:
Contractor security was always a bad idea. Move them to the Representative's asset protection role instead. That makes way, way, way more sense.

Sparrowbird611:
If you mean corp rep, that already exists. Contractor SAARE/PCRC asset protection units are rather common.

Glitterhoof:
I like the changes but woud it also be possible to add an O-2 pathfinder since the chaplain can be an O-2?

Difrit:
Some issues arose with my fork, going to delete the fork and recreate rather than hassle with fixing it, will recreate this PR. Comments have been noted, contractorsec section of this PR will be removed, will add O-2 Pathfinder.

PR #409, opened 2020-02-20
Limits the amount of implants per type you can implant in an external limb to 1.

BurgerLUA:
Basically prevents duplicate implants in the same limb. I mean, sure you can implant 8 microbomb implants, one in each limb, but at least it isn't 50.

### DIFF
--- a/code/game/objects/items/weapons/implants/implant.dm
+++ b/code/game/objects/items/weapons/implants/implant.dm
@@ -45,11 +45,23 @@
 /obj/item/weapon/implant/proc/implanted(var/mob/source)
 	return TRUE
 
-/obj/item/weapon/implant/proc/can_implant(mob/M, mob/user, var/target_zone)
+/obj/item/weapon/implant/proc/can_implant(mob/M, mob/user, var/target_zone, var/obj/item/weapon/implant/I)
 	var/mob/living/carbon/human/H = M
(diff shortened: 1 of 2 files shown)


MemedHams:
okay, so: Why not simply hardcap the number of microbombs that can be implanted in a single body to 8? This also nerfs implants in general, and I'd be willing to have that conversation, but that requires it's own PR, and maybe a vote if we do those.

Chaznoodles:
This would impact traitor counselors/mentalists/psionics in general who want to up their psionic levels but can only have one amp. Set a hard limit to 5 or 8 or something.

Anyone dedicated enough to blow all their points on micros deserves some payoff.

BurgerLUA:
> okay, so: Why not simply hardcap the number of microbombs that can be implanted in a single body to 8? This also nerfs implants in general, and I'd be willing to have that conversation, but that requires it's own PR, and maybe a vote if we do those.

That doesn't make much sense code wise and realism wise. It's much saner to make it so that each limb has a limit of implants so you don't become a god.

Sparrowbird611:
Ok so, I rarely use implants so correct me if I'm wrong on my logic when I suggest this, but I think you should probably be able to have like three head implants. Otherwise, you may just have people implanting the death alarm, tracking implant, or literally anything else to prevent mindslaving.

PR #414, opened 2020-02-21
Removes Vox Armalis Nab Grab

EldritchSigma:
Saw this somewhere on the discord that this is to be added or rather removed.

Removes the Vox Armalis nab.


EldritchSigma:
Should work since its a simple removal.

ProbablyCarl:
While a simple change, this is something we've been discussing. 
It was added a long while back, due to the species being whitelisted, and filling a very niche role with one slot.
It no longer really has a reason to exist, and just fuels powergame type plays.
I'll ultimately leave it up to Mazian.

MemedHams:
I'd argue that, in-game, an armalis -should- be more powerful then most other races at CQC, considering that they are war-bred bio-mechanical monstrosities. Maybe reducing the number of armalis spawns might help, though I wouldn't be sure about that either. A competent armalis should, IC, be a dreaded threat that any normal human would be extremely cautious about facing.

Andromeda-K22:
I haven't seen this abused, really. I've only seen Armalis use to meatshield with the crew and such, so uh, i'm gonna close this.

PR #416, opened 2020-02-21
Changes the Supermatter Cascade announcement

ShadeAware:
### What this does:
A very simple edit to the announcement for a Universal Collapse to be less universal. Because a fair majority of people agree that a ship running off of a _reality ending superweapon_ is a bit much.

Still leaves it up to interpretation if you still want it to be universal I guess.

:cl:
Edit: Supermatter Mega Cascades are no longer universal in scale, atleast when it comes to the announcement.
/:cl:

### DIFF
--- a/code/game/gamemodes/endgame/supermatter_cascade/universe.dm
+++ b/code/game/gamemodes/endgame/supermatter_cascade/universe.dm
@@ -60,18 +60,14 @@ var/global/universe_has_ended = 0
 	new /obj/singularity/narsie/large/exit(pick(endgame_exits))
 	spawn(rand(30,60) SECONDS)
 		var/txt = {"
-A galaxy-wide electromagnetic pulse has been detected. All systems across space are heavily damaged and many personnel have died or are dying. We are currently detecting increasing indications that the universe itself is beginning to unravel.
+Attention [station_name()]. Anomalous energy readings have been detected emanating from your vessel; long range scanners seem to be having some trouble getting an exact readout.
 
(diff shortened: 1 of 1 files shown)


ShadeAware:
Fuck phone typing

Roland410:
This is literally an admin event though, I don't see the point of changing this as you can't achieve it in-round.

ShadeAware:
You can with a loose black hole and SM shard afaik

Roland410:
Shouldn't be hard to test, but if you go into secrets, the SM cascade button triggers this exact one.

Andromeda-K22:
The supermatter cascade is supposed to be something within the realm of theory and not actually known to happen in-character. It should be considered the territory of theoretical physics, as the supermatter is already not very well understood; nothing good could come of tossing a hunk of it into a artifical singularity.

I don't see the need for this.

The bluespace rift generated by the cascade also expands infinitely; therefore, it would literally mean the end of the universe.

ShadeAware:
Whether or not its universal or not is left up to interpretation in the actual announcement proper. No matter how the event plays out, the previous version was a little silly in its wording and description.

Andromeda-K22:
"A galaxy-wide electromagnetic pulse has been detected. All systems across space are heavily damaged and many personnel have died or are dying. We are currently detecting increasing indications that the universe itself is beginning to unravel."

That's pretty unambiguous.

Andromeda-K22:
This seems unnecessary - Closing.

PR #417, opened 2020-02-21
Balances IPCs and FPB damage.

BurgerLUA:
Okay. I did some investigating and it turned out that the IPC species was heavily reliant on RNG on whether or not they take damage to their internal organs.

Since they don't take pain or anything, an IPC can only get fucked if their battery is damaged. This is a very rare event because the chance to hit a battery is around the 5%-10% chance range compared to a human which is about a 80-110% chance range.

I balanced it so that their batteries have a much higher chance to be hit given that they lack other organs inside their chest, as well as gave them a 10% weakness to brute damage and a 50% chance weakness to burn damage, as burn damage has a 50% lower chance to damage their organs.

Also increased the chance from 5% to 10% because Bear is a little bitch.

### DIFF
--- a/code/modules/organs/internal/eyes.dm
+++ b/code/modules/organs/internal/eyes.dm
@@ -110,6 +110,7 @@
 
 /obj/item/organ/internal/eyes/robot
 	name = "optical sensor"
+	relative_size = 10
 
 /obj/item/organ/internal/eyes/robot/New()
 	..()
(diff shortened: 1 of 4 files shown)


Difrit:
As an IPC player, these are good changes. As it currently stands, you can wade through pretty much anything as a IPC. This means a lot less of that happening.

PR #418, opened 2020-02-21
Improves slurring, stuttering, and intoxication.

BurgerLUA:
Slurring, stuttering, and intoxication is improved now.
Instead of stuttering one sentence and being completely fine the next, it's now more gradual.


Andromeda-K22:
Fix globals.

PR #427, opened 2020-02-23
Service Runabout

ProbablyCarl:
- - -
Tweaks:
 - Kitchen given a facelift, to bring it in line with the soon to be added Aurora cooking.
 - Chef and Barkeep jobs increased to two per job. As such, they've been given a minor prep south-east inside the mess hall.
 - Chlorine Exoplanet unloaded. Shouldn't cause issues, but knowing Baycode, it'll probably break something.
 - Barkeep and Chef given new titles.
 - Number of crewmen slots raised to ten.


MarinaGryphon:
No don't merge it

PR #431, opened 2020-02-23
Makes nuke explosions better also clears bay from ending screen

BurgerLUA:
Nuke explosions no longer hang the fucking server.

Also the nuclear ending screen was fixed.


ProbablyCarl:
debby downers want this reverted
something about poor taste
i'll show them >:(

PR #449, opened 2020-02-25
Removes Security and Engineering's Medbay Access

MarinaGryphon:
Because Sec won't stop standing in front of the fucking door on harm intent.

It is not Security's place to barge into Medical unannounced. It is not Security's place to stir shit in Medbay's lobby. It is Security's place to listen to what Medical says while they're in Medical, but it seems people can't understand that when the temptation of walking right in and shouting is there.

### DIFF
--- a/html/changelogs/MDP-secgetout.yml
+++ b/html/changelogs/MDP-secgetout.yml
@@ -3,4 +3,3 @@ delete-after: True
 
 changes: 
   - tweak: "Security no longer has access to Medical."
-  - rscdel: "Killed innocent kittens."
--- a/html/changelogs/MDP-secgetout.yml
+++ b/html/changelogs/MDP-secgetout.yml
@@ -1,0 +1,6 @@
+author: MoondancerPony
+delete-after: True
+
+changes: 
+  - tweak: "Security no longer has access to Medical."
+  - rscdel: "Killed innocent kittens."
--- a/html/changelogs/MDP-secgetout.yml
+++ b/html/changelogs/MDP-secgetout.yml
@@ -2,4 +2,4 @@ author: MoondancerPony
 delete-after: True
 
 changes: 
-  - tweak: "Security no longer has access to Medical."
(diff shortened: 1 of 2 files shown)


html/changelogs/MDP-secgetout.yml
```diff
@@ -0,0 +1,6 @@
+author: MoondancerPony
+delete-after: True
+
+changes: 
+  - tweak: "Security no longer has access to Medical."
+  - rscdel: "Killed innocent kittens."
```
MarinaGryphon:
```suggestion
```

html/changelogs/MDP-secgetout.yml
```diff
@@ -0,0 +1,5 @@
+author: MoondancerPony
+delete-after: True
+
+changes: 
+  - tweak: "Security no longer has access to Medical."
```
MarinaGryphon:
```suggestion
  - maptweak: "Security and Engineering no longer have direct access to Medical and must be allowed in."
```

ProbablyCarl:
It'd be easier to, and much more sensible to change the area access type, and restrict it to Medical.

Roland410:
I think this should be a broader range PR such as removing medical and engineering's access to security as well, considering they have no reason to be able to come and go as they wish. Not sure if other departments are affected or not, but unless major access changes happened since the Bay forking, there will be more I forgot.

MarinaGryphon:
> It'd be easier to, and much more sensible to change the area access type, and restrict it to Medical.

Ìt wasn't set to autoset, so no, it would not be easier. You would also have to remove the mapped in access vars to prevent Travis errors, meaning an extra line change and then some code changes.

MarinaGryphon:
> I think this should be a broader range PR such as removing medical and engineering's access to security as well, considering they have no reason to be able to come and go as they wish. Not sure if other departments are affected or not, but unless major access changes happened since the Bay forking, there will be more I forgot.

Out of scope. This is literally only to prevent Sec from loitering in the medbay hallway and getting patients killed, which is a pretty frequent problem.

PR #454, opened 2020-02-26
Various Bugfixes and Rebalancing for Ascent

Tupinambis:
This is a collection of bug fixes and balance changes for the Ascent. Weapon balance changes absolutely subject to suggestions and change. The main intent is to make Ascent a bit more fun to play, improve quality of life, fix shit, and hopefully make particle beams less meme tier against armored personnel. Flechettes are probably a tad OP, let me know what you think.

More mapping changes to come when I have time. 

* Ascent exosuits no longer degrade.
* Fixes up and rebalances flechette rifle (now fires 25 damage, 40-AP flechettes in 2 round bursts from an 18-round rechargable battery.) 
* Removes the equipment boxes from the armory and code.
* Replaces equipment boxes with 4 particle lances, 2 flechette rifles.
* Particle beam damage 35 -> 40. Armor pen 50 -> 15
* Small Particle beam damage (From particle projector) 25 -> 30. Armor pen 30 -> 10
* Nanoblade for gyne exosuit is now functional. It is an energy sword that does 5 extra damage per swing, but has no block.
* A modified Particle Lance replaces Particle Defender, fires 2-round bursts from an 8-round capacity battery. Slightly less bulky to use. Lets you switch between stun, shock, and lethals
* Mantid gas reactors changed to generate gas over time as intended, including captive oxygen tanks.


TheMaskedMan2:
I think the gun nerfs are overkill in my opinion. I never really had problems personally with Ascent weaponry, and AP goes a long way in baycode to making guns effective.

Alates primary is now weaker than Laser Carbines. For example, the smartgun does 40 burn against a normal plate carrier only achieved 15 burn with all 14 shots.

Yet the Laser Carbine with only 20 AP and 10 more damage manages to reliably do 15-20 a shot.

I don't think you can offset effectiveness of AP with just giving it a bit more damage, it will make the weapons absolutely useless against any armored opponent. I believe baycode armor as if all "Higher" tier weapons are supposed to have some AP to offset armor. For example if you do 60 damage with no AP it might still all be blocked, but 30 damage with 30 AP will apply all that damage through a certain tier of armor.

tl;dr - I never thought Ascent weaponry was meme-tier at all, it was pretty standard, and I fear removing most of their AP (When every higher tier weapon has AP) will result in them being terrible side effects.

code/modules/ascent/ascent_rigs.dm
```diff
@@ -1,7 +1,7 @@
 // Rigs and gear themselves.
 /obj/item/weapon/rig/mantid
-	name = "alate support exosuit"
-	desc = "A powerful support exosuit with integrated power supply, weapon and atmosphere. It's closer to a mech than a rig."
+	name = "alate utility exosuit"
+	desc = "A powerful utility exosuit with an integrated power supply, internals system, toolset, plasmacutter, and maneuvering jets. It's closer to a mech than a rig."
```
Roland410: Old description was fine, don't need to have every module in it.

code/modules/ascent/ascent_rigs.dm
```diff
@@ -47,36 +46,36 @@
...
-	icon_state = "blade"
-	interface_name = "nanoblade"
-	usable = FALSE
-	gun = null
-
 /obj/item/rig_module/mounted/flechette_rifle
-	name = "flechette rifle"
-	desc = "A flechette nanofabricator and launch system of Ascent design."
-	interface_name = "flechette rifle"
-	interface_desc = "A flechette nanofabricator and launch system of Ascent design."
```
Roland410: Fairly sure this was only used by the Gyne rig, which is fine.

code/modules/ascent/ascent_rigs.dm
```diff
@@ -215,7 +212,9 @@
 
 // Rig definitions.
 /obj/item/weapon/rig/mantid/gyne
-	name = "gyne support exosuit"
+	name = "gyne utility exosuit"
+	desc = "A massive exosuit designed for a massive Kharmaani. Has all the features of a standard Alate exosuit, along with an integrated nanoblade, particle lance, \
+			and crystal flechette rifle."
```
Roland410: Once again, don't state all the damn modules.

code/modules/ascent/ascent_rigs.dm
Roland410: Also, the Ascent got nerfed enough, no need to make their weapons use projectile bs as well now.

Roland410:
Going to mostly agree with Masked above. Ascent recently ate an armor nerf because raisins, now if you nerf their weapons and everything, then we might as well just remove them, as they will be nothing to "fear". They are meant to be a highly advanced race with superior technology, not just some purple bugs with reskinned secgear.
What I'd do honestly is just nerf sec gear, as it's absurdly OP as Masked said with the laser carbines.

TheMaskedMan2:
Just to further elaborate, I did a test earlier with Ascent gear with the current damage stats. This is all tested on a normal Sec Officer with a standard plate carrier. Not even ablative or heavy.

Mag Dump of Each

Laser Carbine (Sec Armory) - 10 Shots - 333 Burn - Organs are messed up and there is 50 Tox
Small Lance - 11 Shots (Mag Capacity is 10 but it regen'd 1 by the time i finished shooting) - 154 Burn - 10 Tox (Weaker than Laser Carbine, but the AP lets it be more useful against higher armor, and they self regenerate ammo.)
Gyne Big Lance - 19 Shots (Capacity is 18, Burstfires) - 503 Burn - 80 Tox (Argurably too strong, but you could make an argument that there is only one gyne, but even then the most i'd suggest is take her down from the 50 AP to the Small/Lances 30 AP)

Roland410:
Yes, but at the same time, sabots have like 95AP which security has available at roundstart, compared with the fact that the Ascent armors got nerfed, it literally obliterates them.

TheMaskedMan2:
Yeah I agree, especially with Ascents 1.5 Brute Mod, Gynes are the same.

Anyways, I feel like their weaponry and armor was fine as is.

Roland410:
By the way thanks for the testing, but also we can't forget that if security gets knowledge of Ascent, they will probably gear up in ablatives, which would reduce that by a LOT as well.

TheMaskedMan2:
I should mention my test wasn't with this PR. It's with what is currently live, I wasn't sure if I made that clear. Which this PR results in even lower stats than those. Here are what the stats would look like with these weapon changes. Normal Plate Carrier.

Mag Dump of Each
Laser Carbine - Unchanged
Small Lance - 12 Shots - 50 Brute only.
Big Gyne Lance - 19 Shots - 250 Brute.
(For Fun) Bulldog Frangible - 3 Shots on Alate in a Rig - 100 Brute, 40 Tox (Probably the 1.5 Brute mod at work.)

I forgot to test the flechette but considering it was also lowered to 15 AP I imagine its the same.

Roland410:
I forgot to reply to this, so basically a fully discharged small lance can't even put someone into crit and the gyne lance if all shots hit can kill someone, while the bulldog that security has can outright murder an alate with 3 shots. Balanced.. smh.

Tupinambis:
For context, I was airing on the safe side because people complained about Ascent being op. Evidentally their analysis is kinda quirky, so I'm going to try buffing AP again to reach a middle ground.

code/modules/ascent/ascent_rigs.dm
Tupinambis: Only the Flechette rifle does, and it was arguably a buff and adds a new Option for Alates to choose.

Roland410:
The Ascent's rifle should NOT be weaker than a regular laser carbine.

Tupinambis:
!!!Closing to modularize and reconsider balance Changes!!!

Tupinambis:
I shouldn't do this crap at 2am

PR #460, opened 2020-02-26
adds job whitelist

Andromeda-K22:
title


Sparrowbird611:
So this is just a ground work PR for potentially making roles like command staff whitelisted?

Chaznoodles:
No.

Whitelists create elitism and cliquey behaviour, limit player creativity and add nothing that taking IC action/talking to a player/issuing a jobban will not solve. I will always argue against them.

Roland410:
Going to agree with Chaz here. If someone is being bad at a job, just jobban them. We really don't need whitelisted jobs. One of the reasons I really liked this server in the first place is that the species didn't have one either, which doesn't mean I can't roleplay as one, but that I don't want to write up an essay on why should I be allowed to choose it.

Tupinambis:
All we need to do is implement role time checks. We literally have dedicated role slots for learning departments. Lock heads of staff behind timers in respective departments. Play some Bridge Officer before you play CO and such. 

This also means staff will have more of a chance to notice someone who is consistently garbage BEFORE they are given the power of a head.

Chaznoodles:
@Tupinambis Already exist and are active, to my knowledge.

Roland410:
@Tupinambis The thing about that is, that you'll either have to:
A. Create a character just to get the hours to be a head then discard it or promote them, which is pretty icky to me.
B. Ask admins to override your hours, giving them more work.

I may have personally not played a lot of engineering on this server, but I did on Bay for example and I know how to do shit, but if that were implemented, I'd have to do either of the above, and none of that is favorable to me. Timelocks for roles already exist right now, and if someone is a shitter, admins can jobban them.

GrayRachnid:
A whitelist for high ranking roles is a pretty good idea to me personally for one reason only. It adds importance to the role. Frankly i'd love to see this applied to the CO and SEA role, though some would disagree with the SEA choice. 

This way, these two roles would become rare and you'd have most shifts with only the XO running the ship. The XO becomes the stair step that you have to take in order to become a CO.
Apply the same concept for SEA but with BC/DC/SE as the per-requisites with more focus on their attitude than their competency in their job.

Hopefully, this adds some power behind those who play CO and SEA. Yes, I don't disagree that some elitism might sprout out but that's why the filtering of people needs to be made by the right people and bad whitelisted people should be removed quickly.

Andromeda-K22:
This is a framework.

Expect this for infantry. This may be used elsewhere.

ProbablyCarl:
This isn't intended to be added to existing roles as they are currently.

Roland410:
The problem is, as you have said, is that it's alread planned for the infantry (a whole branch/role I think is pointless unless adminspawned instead of ERT or such) and may be used elsewhere.
While knowing that currently existing jobs aren't going to be whitelisted is good, it opens up the possibility for people who get whitelisted to easily abuse their position (ala Bill Carson, I think most people know the name) and these people usually go into admin protection eventually (unless they manage to do something so dumb that the community rages about them as a whole).

All in all, I still don't think this is good for the server.

Jeff-Dome:
white-list servers nothing but a terrible way to balance around unbalanced features. The fact that this is been suggested as a change by whomever is slightly concerning. However, there is role which could be better taught, but slapping a white-list ruins interaction and dynamic between crew and command that isn't the best but likewise there is roles that could dearly need improved player knowledge but there is otherwise than the most ham-fisted way about which is white-lists, we already have time locks that could be from Job role play time where you actively have to play within the role which is much likely to have an impact than some terrible white-list and as well, admins have tools in place to remove shitters already from the role if need be. Honestly if administration is the issue, wouldn't it be wise to improve tools already in hands of mods and admins than trying or making framework for is likely the most ham-fisted way of controlling and improving all command or security or wherever this framework would be applied. 

I believe within my opinion, while this is good framework, I can't accept white-lists as a game-play balance, as it's a asinine way to balancing races, jobs or other things.

Sparrowbird611:
I'm kind of going to agree with what everyone else has said against whitelists. Very few good things come from whitelists, and it doesn't really stop shitlers too effectively.

Additionally, I think the fact that we need to even consider whitelisting infantry to get them in is a bit more than just concerning, personally.

Andromeda-K22:
This is not any job being whitelisted.

This is a framework.

Stop commenting here. Comment on a PR that actually implements the
whitelist. Thanks.

Roland410:
The issue with "this is not any job being whitelisted..." is that the framework still implies it will happen in the future, which as you can see is not really a desired thing.
You'll get the same comments on a PR whitelisting a job as you do now, but it's better to address it at its roots, as in no framework is necessary.

Andromeda-K22:
There are certain roles - like infantry - where the concept is neat, but a large portion of the server populace can't be trusted with the role. We tried that already, and policing them was a full time job; infantry was never brought to a satisfactory standard and had to be removed because of that. The job whitelist exists because of those specific cases. Do not expect to see this implemented anywhere but roles that have has this issue.

Roland410:
Then it honestly seems like an issue with infantry itself. People can behave (mostly) as security, same with engineering etc.

What's the point of infantry even? Doesn't seem like it'd add much to the server other than branch/role/job bloat. As far as I'm aware they were meant to accompany exploration, but let me know if I'm wrong. If I'm not, then either security could give people to exploration just as any other department like medical, or an exploration guard role could be added instead.

Right now, there is also no content that'd warrant the (re)addition of such in my opinion, unless exploration gets boarded or attacked by Vox or Ascent (or merc assuming that rework is in, but they'd probably focus on the Dagon).

Jeff-Dome:
the issue was with marines is that they was far too used where security would be more acceptable but white listing infantry, what I believe my opinion from information available to me is that role was believed to be more with dealing with external threats than internal threats and thus white-listing this job seem retarded because it's not addressing the issue of player game play by force of role but rather a stop-gap of a terrible system. again, I don't disagree that the framework is bad but I'm disagreeing with how the framework will be applied through jobs. I don't wish for the PR to be merged on the basics of slippery slope idea as then it's the framework and then so and so forth. 

I believe rather than using this ham-fisted method for a catch all for terrible players is general a bad idea, I believe if that's the issue with, wouldn't it be better to forward game play changes and ensure that they are more actively engaged towards their intended game play objective of securing the Dagon? or rather changing the infantry department goals to be, have X or prevent X from being blown up or long those lines.

besides, the reason infantry was such an attractive role was due to the power and the fancier equipment that security was given and how they could apply without fear of them being under much IC punishment, however wouldn't it be wise to improve the regulations and/or apply different set of regulations to the infantry and have them act completely differently rather them screaming for red alert and auth for lethals so they may help out security and due the atmosphere around the role, it created a more secure the situation as any military like deployment. Generally military level deployment means that everything has failed and your moving towards a critical situation so you commit towards an assault and thus within an IC game play, your inclined to shoot anything that moves and the lack of a visible and audible surrender button/emote makes it harder of antagging and general crew play.

code/modules/boh/job_whitelist/jobwl.dm
```diff
@@ -0,0 +1,42 @@
+#define JOBWHITELISTFILE "config/jobwhitelist.txt"
+
+GLOBAL_LIST_EMPTY(job_whitelist)
+
+/hook/startup/proc/loadJobWhitelist()
```
MarinaGryphon: not entiiirely sure why this is a hook and not done in a subsystem somewhere, like SSconfig or something, but it works i guess

code/modules/boh/job_whitelist/jobwl.dm
```diff
@@ -0,0 +1,42 @@
...
+	if(config.job_whitelist)
+		load_job_whitelist()
+	return 1
+
+/proc/load_job_whitelist()
+	GLOB.job_whitelist = file2list(JOBWHITELISTFILE)
+
+/proc/is_job_whitelisted(var/client/M, var/job)
```
MarinaGryphon: I'm not sure why this isn't a proc on `client`, but that's just a stylistic thing. Should work fine regardless.

MarinaGryphon:
Solid code-wise. As it's not actually implemented anywhere at the moment, I feel it's fine to merge right now and wait for feedback when it's actually used somewhere.

PR #461, opened 2020-02-26
Ascent Fixes! Now you can take hostages safely!

Tupinambis:
Ascent Code is a bit of an unfinished mess. This PR makes some improvements to the code that mostly consists of fixing oversights. There are no unique sprites for the nanoblade or crystal flechette cannon currently.

Additionally, the nanoblade does not currently light up when activated. This is likely because RIG modules don't actually show up in hands (Which they absolutely should for both balance and visibility reasons) so I'd rather that be changed, than make a hacky solution.

Anyways, enjoy.

* Gynes will now spawn with a filled gear harness in their belt slot.
* Alates will now spawn with the gear harness in their belt slot instead of their suit slot.
* The Gyne exosuit nanoblade module has been fixed and replaced with an integrated energy sword that does slightly more damage (35 instead of 30) at the cost of having no block chance.
* The Gyne Flechette module has been fixed to recharge properly, and now fires three round flechette bursts from an 18 round capacity "battery". 
* Reactor tanks now properly regenerate gas over time without the need for a power source.
* Tentative fix for exosuit armor degredation.

### DIFF
--- a/baystation12.dme
+++ b/baystation12.dme
@@ -1445,6 +1445,7 @@
 #include "code\modules\boh\ascent2\projectiles.dm"
 #include "code\modules\boh\ascent2\structures.dm"
 #include "code\modules\boh\ascent2\turf.dm"
+#include "code\modules\boh\ascent2\weapons.dm"
 #include "code\modules\boh\culture_descriptor\culture\cultures_sergal.dm"
 #include "code\modules\boh\culture_descriptor\faction\factions_human_boh.dm"
 #include "code\modules\boh\culture_descriptor\faction\factions_sergal.dm"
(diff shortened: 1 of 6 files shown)


Roland410:
This looks way better, but how does the flechette thing work? It's an energy weapon that fires bullets and doesn't seem to recharge, while having a firemode selector that only has one mode?

Tupinambis:
> This looks way better, but how does the flechette thing work? It's an energy weapon that fires bullets and doesn't seem to recharge, while having a firemode selector that only has one mode?

It looks weird, but from an actual coding perspective it is much simpler to code. Energy weapons are weapons that fire from batteries. Think of the bullets as being battery powered magic bullets. It does recharge. Firemode selector only having one mode is just to override the firemode of the parent type without breaking things.

TheMaskedMan2:
Just curious why the outfit for Alates was removed, won't Alates just spawn naked now?

Also I believe
armor_type = /datum/extension/armor/ablative
is related to the armor degredation mechanic, based off of the original PR for Armor Degredation. Not sure if it should be left in then.
https://github.com/Baystation12/Baystation12/pull/27624/files

Tupinambis:
Alates and Gyne share the same outfit now.

PR #465, opened 2020-02-27
Requested (Various) Changes

ProbablyCarl:
Tweaks:
 - Cadet given a light vest, rather than the medium as other Officers get by default.
 - Light vests included in the armoury vendor.
 - Extra pepperball mag in equipment kits replaced with the standard rubber mag, for backup scenarios.
 - Additional shock plating added to the accessory vending machine.
 - Loyalty kits added to the armoury.
 - Additional shock plating added to the E-Arm. Instead of one plate, there's four, as was intended originally.
 - Introduces precharged Shield Generators aboard the ship for both Deck Two and the Bridge backup Generator.
- - -
Bugfix:
 - The Cadet hud icon missing has been corrected.


Roland410:
Please don't add loyalty implants, they're a really awful item/mechanic.

Sparrowbird611:
If loyalty implant is the name for mindshield implant here, then I agree with roland.

Sparrowbird611:
If loyalty implant is the name for mindshield implant here, then I agree with roland.

ProbablyCarl:
That's fair. I'll remove them, given.

Roland410:
Loyalty inplant unless used with mindbreaker gives the written "suggestions" to the implanted person. With mindbreaker they become like laws.

Roland410:
While you're there, unless it was changed, the XO has a box of them as well, probably should be purged from anywhere else and kept as the traitor item it is now.

Roland410:
While you're there, unless it was changed, the XO has a box of them as well, probably should be purged from anywhere else and kept as the traitor item it is now.

N8-Toe:
with the compromise of having both rubbers and pepper balls, is a box of rubber magazines back in equipment storage? also I do want to echo what others have said, loyalty implants hamper RP, not encourage it

Chaznoodles:
To my knowledge, Cadets are currently limited to Basic Firearms and Trained CQC, which is wack considering you can have passengers and doctors sprinting about with Experienced. Should probably be looked at in conjunction with this PR.

Is there any benefit to having a light plate over medium? As it stands, it just looks to me to be a "Can I just get a medium plate from the armoury instead?" time delay, probably better off to just give them a medium plate like everyone else.

conduct

ProbablyCarl:
Light plates don't have a benefit, no. It's purely a flavor change, as they're legitimately worse in every fashion in comparison to medium plates. Neither have a slowdown.

I'll touch on their skills, though. They're intended to be introductory roles, similar to Engineer trainee, Research assistant, etc. They are hampered quite a heavy amount though, and it's more than just a tad.

Roland410:
Introduces precharged Shield Generators aboard the ship for both Deck Two and the Bridge backup Generator.

Doesn't seem like a security change unless we really want to go loose on the terms of security.
On the other hand if you want to change shields to be less shit, it'd be a welcome one. (Also there are off-ship shields such as the Bearcat's that should have this applied, and as such, making the base variant of the shield spawn precharged per code should solve this)

ProbablyCarl:
Security cadet now has another title, alongside providing the vendor with two spare rubber mags.

PR #467, opened 2020-02-27
Osmium Carbide Plasteel will now once again withstand up to 12000K

BlueNexus:
Note that this is used on the ship, in the burn chambers. One of your recent changes broke it so it had the same melting point as plasteel, so this deals with that.

This isn't just used off-map.


BlueNexus:
Note that 12000 was the original melting point they had when I added them, and the sensors in the burn chambers are set assuming this value.

PR #470, opened 2020-02-27
Properly incorporates Fleet E1s into all roles.

GrayRachnid:
E1s are now available for all trainee roles. Security's E2 was removed since the Cadet has the choice of E1 or E2.


Chaznoodles:
Limiting Sec E2 to Cadet feels a bit restrictive, as there's a good deal of people who don't want the weird standards that come with being a JNCO. Might be worth changing that back.

GrayRachnid:
E3 is not a JNCO, that rank doesn't exist. E1-E3 are Enlisted. E4-E6 are NCOs. E7-9 are SNCOs.

PR #477, opened 2020-02-28
Removes Paramedic from Diona

Sparrowbird611:
No diona player has ever played paramedic as far as I know, but I'm doing this any way. Have you seen the speed Diona move at? They run at a mile per ten hours, and they can't even run, so they go even slower than that. This is one of the few roles where speed is absolutely necessary, something Diona can't offer. So here we are.

### DIFF
--- a/maps/torch/job/torch_jobs.dm
+++ b/maps/torch/job/torch_jobs.dm
@@ -16,7 +16,7 @@
 		/datum/species/unathi/yeosa = list(HUMAN_ONLY_JOBS, /datum/job/representative),
 		/datum/species/skrell  = list(HUMAN_ONLY_JOBS, /datum/job/representative),
 		/datum/species/machine = list(HUMAN_ONLY_JOBS, /datum/job/representative),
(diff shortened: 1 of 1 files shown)


Roland410:
This doesn't remove Paramedic. This removes medtech, corpsman AND paramedic from dionaea.

Sparrowbird611:
………..Aren't those all the same role....?

Sparrowbird611:
I know paramedic and corpsman were both alt titles for medtech, but I assumed med tech had the same responsibilities.

Roland410:
Paramedic and corpsman are medtech alt-titles, and I don't think disallowing them from those roles are a good change. Dionaea may be slow, but they are totally viable to play medical with.

Sparrowbird611:
Yeah, I didn't disallow the rest of the roles, I just removed access to med tech because I assumed it wwas just another name for paramedic. If it's not, I'll close this.

Roland410:
Medtech is the "regular" name, paramedic is the alt-title along with corpsman. Same as counselor having the mentalist alt-title (although different mechanics apply for that).
http://prntscr.com/r96hj4

Sparrowbird611:
So med techs are just doctors then? In that case, I'll close this.

PR #478, opened 2020-02-28
Merchant Station Changes

miniusAreas:
This should make merchant a less horrifying role where you can luck out and get 4 traders who sell jack shit.

- Adds ~80 random legal items for merchants to shift through and sell
- Adds 4-6 random Illegal items for merchants with less morals than a vox
- Adds a tool closet for merchants in their engineering section

If something here's fucked please slap me so I can fix it. I'm not a coder, nor the headdev, and the last PR I had got closed for some reason that nobody's told me. I value my work and time I try to put in and would prefer to fix it rather than having it yeeted into nullspace.

### DIFF
--- a/maps/torch/z1_admin.dmm
+++ b/maps/torch/z1_admin.dmm
@@ -242,35 +242,35 @@
 /area/holodeck/source_emptycourt)
 "abw" = (
 /turf/simulated/shuttle/wall{
-	icon_state = "swall_s";
-	dir = 4
+	dir = 4;
+	icon_state = "swall_s"
 	},
 /area/centcom/evac)
 "abx" = (
 /obj/structure/shuttle/engine/propulsion{
-	icon_state = "propulsion_r";
-	dir = 1
+	dir = 1;
+	icon_state = "propulsion_r"
 	},
 /turf/space,
 /area/centcom/evac)
 "aby" = (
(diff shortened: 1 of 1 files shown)


Roland410:
I think you could easily up their credits to like 15-30k as they mostly can't make that money back that the items are sold for, since crew don't have as much money. That way you could allow merchants to supply the station with stuff, while they could lower the prices.

miniusAreas:
> I think you could easily up their credits to like 15-30k as they mostly can't make that money back that the items are sold for, since crew don't have as much money. That way you could allow merchants to supply the station with stuff, while they could lower the prices.

The issue is moreso that the traders that you get on the merchants pad are awful. It's either guns (something most crew can't get), Highly illegal stuff (that security will probably lynch you for), food, or a bunch of ranks and shoes that the crew can easily get anyways. Rarely you get the automated manufacturing/mining/goods beacons that sell 1-3 things that people would want/need to buy. Hell, I can't even buy a medkit from any of em. In reality there needs to be an overhaul of the merchants beacon stuff, but that's way out of my leauge and I'm pretty sure there's not many people that would want to put effort into it.

Roland410:
I once wanted to do one, but I agree with you on the trader stuff. Although if you bring silly things like an Admiral rank or such I'd gladly buy that in a round, and they tend to be cheap, so never disregard such.

PR #479, opened 2020-02-29
Makes Travis Stop Failing PRs

Michiyamenotehifunana:
:cl: Toriate
bugfix: Curved horns for humanathi should work again
/:cl:

This PR aims to make Travis stop failing every single PR on this repo. To that end, several steps have been taken:

1. Global Variable checks have been disabled. They are simply meant to prevent people from creating map-specific globalvars, but just seems to make Travis fail PRs every other day.
2. Indentation checks have been disabled. The current checks will fail if comments exist within lists, so they kinda suck. The amount of `<<` uses in the code quality checks have also been updated to reflect their current state.
3. Unit test failures have been resolved. The ascent freezer closet has had blank icon states added, and the aux armory has been properly exempted from the unit tests.
4. The Humanathi Curved Horn sprite accessory is referred in the code as `curved`, but in the actual .dmi containing the horn sprites, the iconstate was named `curled`. The .dmi has been amended.
5. Comments have been removed from lists. These are harmless, but the way Bay's Travis Indentation check is written causes it to automatically fail any lists with comments in them.

This should ultimately make Travis a more useful asset to this repo, and would actually make it a reliable indicator for whether or not can a PR compile.


Roland410:
I think 1-3 should be kept, and a fix pushed that solves the actual issues, which if you don't want to do, I will (once I figure out some mapping issues).
With this PR, we are just chucking the existing issues under the rug.

Michiyamenotehifunana:
There is no actual issue in global vars being changed, and having more textpaths, proc overrides, spawn proc calls and `<<` outputs than Bay does, isn't really an issue unless you're really trying to make your code more optimized than Bay's is. 

However, regarding the closet unit test, I just spent the past hour digging through closets and I think the Ascent's locker iconstates are causing the closet test failure. I will attempt to fix that. I also have a hunch regarding the vacant armory inheriting the exemption flags of vacant areas, so I'll also see if I can fix that.

Michiyamenotehifunana:
Unit tests are now restored. Missing icon states are added to the Ascent's freezer (actually just blank iconstates), and aux armory's exemption flags have been properly set to reflect its actual state.

Roland410:
Very nice, thank you. Finally Travis may mean something as long as proper conduct is followed and no speedmerges and shit.

Chaznoodles:
based

MarinaGryphon:
> 1. Global Variable checks have been disabled. They are simply meant to prevent people from creating map-specific globalvars, but just seems to make Travis fail PRs every other day.

This is because we somehow have map-specific globalvars, because we messed up our repo and need to fix Torch-only code.

> 2\. Code quality checks (including color hex checks) have been disabled. These are just arbitrary standards that Bay chose to enforce. If we're not enforcing them as well, they're pointless.

We're kiiind of enforcing them, but also kind of not. They're mostly trivial to fix except for the incorrect indentation ones, which fail any list with comments after entires, for some reason. Some exist for a reason, some... don't.

> 3\. Unit test failures have been resolved. The ascent freezer closet has had blank icon states added, and the aux armory has been properly exempted from the unit tests.

Good job. I fixed them last time but people seem to not care to update the area exemptions when they change areas...

> 4\. The Humanathi Curved Horn sprite accessory is referred in the code as `curved`, but in the actual .dmi containing the horn sprites, the iconstate was named `curled`. The .dmi has been amended.

Good work here too.

> 5\. Comments have been removed from lists. These are harmless, but the way Bay's Travis Indentation check is written causes it to automatically fail any lists with comments in them.

Personally, I'd suggest removing the indentation check instead? I thought you already did, as it was included under code quality checks.

test/run-test.sh
```diff
@@ -228,8 +228,8 @@ function run_byond_tests {
         ./install-byond.sh || exit 1
         source $HOME/BYOND-${BYOND_MAJOR}.${BYOND_MINOR}/byond/bin/byondsetup
     fi
-    run_test_ci "check globals build" "python3 tools/GenerateGlobalVarAccess/gen_globals.py baystation12.dme code/_helpers/global_access.dm"
-    run_test "check globals unchanged" "md5sum -c - <<< 'baf6134534402ae5e1b1876cd23fa604 *code/_helpers/global_access.dm'"
+#    run_test_ci "check globals build" "python3 tools/GenerateGlobalVarAccess/gen_globals.py baystation12.dme code/_helpers/global_access.dm"
+#    run_test "check globals unchanged" "md5sum -c - <<< 'baf6134534402ae5e1b1876cd23fa604 *code/_helpers/global_access.dm'" # This is just to stop people from making map specific global vars. I dunno why we should be enforcing that.
```
MarinaGryphon: This is a good idea to enforce because it keeps maps modular... I guess. The idea is that non-map-specific functionality should function regardless of the currently compiled-in map.
Ideally I'd... rather just phase out map-specific functionality entirely, though, so go ahead.

test/run-test.sh
```diff
@@ -205,11 +205,11 @@ function run_code_tests {
     run_test_fail "maps contain no plane adjustments" "grep 'plane = ' maps/**/*.dmm"
     run_test_fail "ensure nanoui templates unique" "find nano/templates/ -type f -exec md5sum {} + | sort | uniq -D -w 32 | grep nano"
     run_test_fail "no invalid spans" "grep -En \"<\s*span\s+class\s*=\s*('[^'>]+|[^'>]+')\s*>\" **/*.dm"
-    run_test "code quality checks" "test/check-paths.sh"
+#    run_test "code quality checks" "test/check-paths.sh"	# This just applies coding standards that we don't really need to follow. At all.
```
MarinaGryphon: The only qualm I have with this PR: Lots of these exist for a reason, like `spawn()` being much slower than using timers, and using `to_chat()` for output instead of << being a requirement for eventually porting Goonchat/VChat/whatever new chat version we want (which is desirable for a number of reasons). I would prefer if these were re-enabled save for the indentation check, which doesn't really work.

test/run-test.sh
```diff
@@ -205,11 +205,11 @@ function run_code_tests {
...
     run_test "indentation check" "awk -f tools/indentation.awk **/*.dm"
```
MarinaGryphon: This indentation check is silly. It checks for the regex `^ {4,}` which is just any line that begins indented with four spaces (or more). It compiles just fine, though, and appears the exact same in most editors as using tabs. Personally, I prefer using tabs, but there's no reason to enforce it with an awk that, for some reason, fails lists that have comments after entries.

MarinaGryphon: Actually, no, that's the other indentation check. This one is just a very, very slow regex that fucks up comments in lists. Either way, it should be disabled.

Michiyamenotehifunana:
Alright, the code quality checks have been restored, and both indentation checks have been disabled. I have also updated the amount of `<<` uses from 475 to 474 to reflect what Travis is currently reporting.

Michiyamenotehifunana:
Looks like #483 introduced disconnected wires and is making Travis fail.

Roland410:
I've tried fixing that, but no results so far. The window grilles are supposed to be electrified by them, but they just don't work.

PR #484, opened 2020-03-01
Security remap, cleans up CL.

Roland410:
🆑Roland410
:maptweak:Remapped the prison portion of the brig, which means a bigger and more fun perma, 2 more normal and 1 more psionic cell, and an execution room.
:maptweak:Psionic dampener implants have been added to the armory, and the psionic monitor to the Brig Chief's office.
/🆑
Cleaned up the changelogs and phased out lentext while I was there (really should have made a different PR, but it should have no issues as they serve the same purpose).
Mostly up to date picture (fixing some issues etc.):
image


Roland410:
DNM shit's fucked.

Roland410:
Ready to merge now, unless someone finds any issues.

Andromeda-K22:
Waiting for atomization of lentext removal, since length is 513 only as far as i'm aware.

MarinaGryphon:
I would also like to suggest removing the excessive number of burn/brute kits from the lockers on the left. They essentially invalidate Medical, IMO, and result in a mentality of "why should we even bother with Medical?" from Sec regarding both prisoners and themselves.

Roland410:
@Andromeda-K22 Already removed lentext stuff, only the changelog cleanup and the area and unit test changes are there that were necessary.
@MarinaGryphon Those lockers only have one advanced brute and burn kit just like the MAA ones that have it right now (unless you were referring to both) which is only used for minor booboos.

MarinaGryphon:
> @MarinaGryphon Those lockers only have one advanced brute and burn kit just like the MAA ones that have it right now (unless you were referring to both) which is only used for minor booboos.

They should have gauze and ointment instead. They're *advanced* brute and burn kits for a reason. It completely devalues doctors by allowing security to just treat themselves, run back into battle, and inevitably get killed due to an infection because they've never heard of hand-washing.

GrayRachnid:
Why cut off the catwalks? Extend them to the rest of the new brig area.

Roland410:
@MarinaGryphon In my opinion security has way more to deal with in general than to run with every booboo to medical ans changing those kits would just make them steal a medkit from somewhere or get medical to give them one.

@GrayRachnid I absolutely despise the plated catwalks, and didn't support Cakey's PR in the first place (normal ones are good tho). If it were up to me, I'd outright delete them from the code. They add about exactly zero value to the map while always looking out of place. They don't cover the entire decks from start to finish, only every few tiles.

ProbablyCarl:
Can we get an updated changelog, to reflect edits?

Roland410:
I just merged dev into the branch to get rid of lentext. I should have time tomorrow, and I'll fuck with the wiring. As far as I'm aware that's the only problem with this right now, as noone reported anything else I haven't fixed.

Roland410:
Removed the wires from perma's windows, which means they are no longer electrified, but that's a small price for the sake of my sanity. This is ready for proper merging, if someone wants to make the windows electrified, please be my guest.

Roland410:
Fixed a missing closet, but it's good to go for real.

PR #486, opened 2020-03-02
Unblacklists head roles from antagonism.

Higgin:
:cl:
tweak: Naughty heads rejoice! The CO, XO, and CoS are no longer blacklisted from antag roles.
/:cl:

Adds more dynamism and potential to different antagonists and more impactful roles - if they choose to enable the preference. Nobody should be above suspicion.

### DIFF
--- a/maps/torch/torch_antagonism.dm
+++ b/maps/torch/torch_antagonism.dm
@@ -1,12 +1,12 @@
 //Makes sure we don't get any merchant antags as a balance concern. Can also be used for future Torch specific antag restrictions.
 /datum/antagonist/changeling
-	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/merchant, /datum/job/captain, /datum/job/hop, /datum/job/submap)
+	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/merchant, /datum/job/submap)
 
 /datum/antagonist/godcultist
-	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/chaplain, /datum/job/merchant, /datum/job/captain, /datum/job/hop, /datum/job/hos, /datum/job/submap)
+	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/chaplain, /datum/job/merchant, /datum/job/submap)
 
 /datum/antagonist/cultist
-	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/chaplain, /datum/job/psychiatrist, /datum/job/merchant, /datum/job/captain, /datum/job/hop, /datum/job/hos, /datum/job/submap)
+	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/chaplain, /datum/job/merchant, /datum/job/submap)
 
 /datum/antagonist/loyalists
 	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/merchant, /datum/job/submap)
@@ -15,7 +15,7 @@
 	blacklisted_jobs = list(/datum/job/ai, /datum/job/cyborg, /datum/job/merchant, /datum/job/submap)
 
 /datum/antagonist/traitor
-	blacklisted_jobs = list(/datum/job/merchant, /datum/job/captain, /datum/job/hop, /datum/job/ai, /datum/job/submap, /datum/job/hos)
+	blacklisted_jobs = list(/datum/job/merchant, /datum/job/ai, /datum/job/submap)
 
 /datum/antagonist/ert
 	var/sic //Second-In-Command


Roland410:
You've accidentally PR'd your .dme.

That aside, I do agree and disagree with you. The problem is that someone like a traitor/cultist CO/XO/CoS has access to a LOT of power, either be it their actual rank and ability to give orders to their underlings or literal one such as armory/all access. If an XO becomes a cultist, the entire cult immediately has access to a ton of weapons, armor, and all-access which will allow them to just bumrush everyone especially if they can manage to convert say the CoS as well.

Other than that, the psychiatrist could probably be a cultist now since it was split to chaplain, and not an "anti-cult" guy.

Andromeda-K22:
.dme changes need to be removed. This will need discussion, as head roles are antag protected for a reason.

Chaznoodles:
Those roles are blocked from antag roles because of game balance. They're also blacklisted on LRP servers for the same reason.

Higgin:
DME fixed, removed restrictions on psychiatrist as well, squashed commits.

The expanded case for allowing all crew on the ship to be antagonists:
- It extends tension and suspense. Will the captain calling me up to the bridge ask me to do something bad or suffer some punishment? Why is security with the CoS asking to come in here without a warrant - or if they have a warrant, who approved it? 
- To this same point: the bridge has been relatively free of this sort of tension due to the Bay-inherited blacklisting of these roles. Now, instead of just having to wonder about the good intentions of lower officers/heads, the relationship goes both ways. 
- Any in-game issue with COs/XOs/CoSs before this could just be attributed to incompetence or suspicion of bad behavior - this change forces players in-round to suspend judgement.
- The power and authority of these roles gives them the ability to vastly broaden the scope of what is possible in a traitor round; because of that, players won't so easily know what to look for, and player behavior won't fall into so many predictable patterns. 
- The greater ability to impact the round may draw more people to these roles which, with exception of CoS, are very irregularly-filled.

The concerns:
- Great access allows antag heads to get guns and equipment allowing them to snowball - especially in group antag roles.
- (The Bay argument) Players should trust these heads; not only is it plausible for people in their positions to want the mission to succeed - it also ensures the rest of the crew can turn to them during crises without fear and without their focus being on anything other than leading/coordinating/getting things done.
- Nobody wants to deal with the fallout of traitor heads; they're hard to get out of their positions and hard to bring evidence against.

The concessions:
- Heads have great access and power already - this may lead to them having a much easier time influencing the round, and shoeing them in as group antagonists may give those group antagonists much better access to secure areas/gear. 
- This may turn on meta-behavior, but crew may be slower to trust/work with heads. 
- Antag heads pose significant challenges to security and the crew for addressing their antagonism - especially without other leadership. 

The ultimate argument: 
Sometimes you're the dog and sometimes you're the fire hydrant. Even if players might rankle at being put on the back foot against antagonists because the most powerful people on the ship can now take the roles, we shouldn't presume antagonists are there to lose - they're in-game to create tension, conflict, and redefine the roles players take. This change allows tension and conflict to extend much more broadly; and it extends the possible outcomes and roles players will take in even the same gamemodes repeated.
As for gear: antagonists can usually access it anyway. The only thing heads have is card access and legitimacy in getting the gear before they do something with it. For group antagonists, all this does is allow heads to start off as antagonists - they could be cultists/gcultists before with conversion, and they still can if they don't roll for those roles, which this change does not force them to do.
People not necessarily trusting heads is alright - SS13 is Mafia in space, and part of the appeal/tension is the question, "who can I trust?" You might trust the CO/XO 9/10 times, but this adds just that sliver of doubt. 
Although antag heads pose a significant challenge to the crew, I think Bastion's players can handle it and enjoy it - both as antag heads and non-antag crew.

Sparrowbird611:
The bay argument is insanely stupid. However, I admit, it is a little concerning to have some of these guys as antagonists. Still, I think it would be an interesting to see how this goes, and I think it would be a good change.

Jeff-Dome:
I support the change but It's begrudgingly, as it just promotes antag only conflict than any char conflicting or and general non-antag conflict but there is dearly needed more conflict sometimes within antag rounds. 

CO and XO are both within a set position of a power which gives them a large amount of lee way with gimmicks but then again this slightly brings into question when it's sextended because everyone will be second guessing their orders but giving my latter comment about crew conflict this could be good but it could also lead to retards being retards, whoever the CoS would be a greater position of power with general antagging but due the large amount of power and stunning equipment granted to them and the general unquestioning of their orders could be fun but I could see this being slightly abused. 

I could see the XO being an traitor but why the CO and CoS? their power and hold over the crew is stupidly high, any sane man wouldn't unblack-list a CO much, much less the CoS. We could ploy on the how it would improve the conflict but I don't want the server to complete devolve into a false-hood of "RP" by hiding it with conflict with antags. Antags should be the main driving force behind rounds but there should be crew leeway, general head incom.

But I'm going to say, this was tried on bay and failed but I'll have to see how it plays out, given the larger population, granted this might make more work for the admins I'm also wondering what could occur from it.

Roland410:
Well, I'll bet 30 bucks that it'll be a clusterfuck where a bad choice by command will result in them being valided by security and other catastrophes like that such as CO/XO spearheading a cult etc. to victory because of their access and authority.

Higgin:
I'm not too concerned about balance in antag roles because they're hardly 'balanced' to begin with - sometimes the entire security department rolls merc and nobody's home; sometimes you're a solo cultist with a dream (after your two compadres go straight to cryo) up against a fully-staffed security and command department looking for something to do. 

Everyone plays the cards they're dealt. While mechanical and item balance improving _play itself_ is one thing, I think the configuration of who takes (and can take) what roles is best left open to better reflect player choice. Short of some sort of complex round-event-antag weighting system (/tg/ dynamic, anyone?) it's better to have a wider range of setups for the game allowing as much choice and as many varied, meaningfully-different outcomes as possible.

I wholly expect that some of those outcomes will be clusterfucks, more conflict with the backdrop of possible antagonism in mind (as opposed to character-driven conflict - which I suspect people who engage in will engage in anyway,) and put people used to being on top on the back foot. 

I don't think we should expect those outcomes with every round though - these are three slots as opposed to the 10 (!) available for security antagonists (which have had their own time in the corner for similar concerns,) and this change having an impact turns on people in these command roles turning on antag prefs in the first place - which most folks don't. 

On a closing note - the occasional clusterfucks have been some of my favorite and most memorable rounds both for the genuine height of in-game tension and the character moments they allowed to come forward. A large part of my motivation here comes from the hope others can have similar experiences.

bearrrrrrrr:
Genuinely unsure why this was merged. The only thing I have to say to all of those arguments is that the first round this was merged, the Captain rolled traitor, the CE rolled traitor, the borg rolled renegade, and the AI started a plasma flood. I'm so genuinely confused as to why this was merged.

> The greater ability to impact the round may draw more people to these roles which, with exception of CoS, are very irregularly-filled.

Do you know what we call it when someone disregards RP'ing to specifically fill in a jobslot that you can do more things on? It's, at best, jobhopping. At worst, it's also powergaming.

Yes, sometimes the antags have terrible, shitty rounds, but that's being handled by giving them better tools to work with. Turning the Dagon into a pseudo anarcho-monarchy where the captain and all of command have rolled traitor is NOT the solutions.

bearrrrrrrr:
Adding more points as they come to me.

On a mil-RP server, characters being disciplined is very important. Obviously, NTEF isn't strictly mil-RP at all. You don't have to be a huge tool that sucks command off and jumps to salute every time the CO walks into a room, but it's important you don't turn yourself into an absolute clown. Having a CO traitor is extremely overpowered for multiple reasons. Not only do they have access to, well, all the tator-tot things, but they also have access to something even more important: IC and OOC protection. Throwing an accusation of being a tator-tot to a captain is basically social suicide. Actually trying to get them DEMOTED is even more difficult. 

And for a CoS traitor, take that, but also add the fact that they are three times better armed than any other command member.

Roland410:
Worst thing is that these were literally said would happen and talked about before the merging of this PR.

PR #493, opened 2020-03-02
Mixed Mode Soup

Higgin:
:cl:
rscadd: Traitor-Nin! Who do they serve? An honor-bound Ninja warrior with ancient technology arrives to the ship just as the Traitors who have lain in wait begin to carry out their wicked deeds. 
rscadd: Wizard Crime! Magic, mischief, and mayhem abound - and SOMEBODY on this ship isn't having it today!
tweak: Traitorling minimum player threshold reduced.
/:cl:

Mixed modes spice things up and overcome the inherent weaknesses of several single-antag modes - that they can only do so much with so many people, and that they've only got the one shot to do it. 

This doesn't change or add anything into secret rotation, but this will make the modes addable and votable options for future use.


Roland410:
"Schism" already exists as a gamemode.

Andromeda-K22:
Schism already exists.

Higgin:
> 
> 
> Schism already exists.

Where is it hiding? I don't find it under either cult or deity; and I don't find it as a votable option when running a clean test server either.

Roland410:
Seems like it's actually been nuked and without a PR even on Bay. It was called Faithless when it existed.
https://github.com/Baystation12/Baystation12/pull/21437

Higgin:
You know, I might bring it back later, but schism's out for now. Thank you for looking into it!

Roland410:
Just keep it gone. It always devolved into cultists becoming godcultists as well anyways and then using double antag power.

ProbablyCarl:
Wizard is disabled due to being an unfun mess, and the traitor/ninja game-mode is unnecessary, as current in round options already exist to allow for such to coincide with admin intervention.

If you're wanting, make an event to mid-round spawn Ninjas, as they're 'intended' to be.
We won't be having them as a mainline gamemode for the foreseeable future, due to overwhelmingly negative player feedback in regards to it.

Higgin:
Will see about separate PR for eventja.

PR #494, opened 2020-03-03
First Step of SFL Implementation

Nakhi1:
More of a proof of concept, but one that works. Xenoic patch was commented out, replaced with the SFL patch. Main marine patch opened to all humans, vulps, and genemods. SFL patch for Unathi and Tajara. SFL patch will be automatically available for any additional alien races allowed into trooper ranks.

- Adds SFL Patch icon and onmob sprite.
- Adds SFL Patch to loadout for all marines.
- Opens SMC patch to vulps and genemods in loadout.


maps/torch/loadout/loadout_accessories_boh.dm
```diff
@@ -20,14 +20,20 @@
...
-/datum/gear/accessory/smc_patch_xenos
-	display_name = "Marine Corps patch (xenoic division)"
-	path = /obj/item/clothing/accessory/solgov/smc_patch/xeno
+/datum/gear/accessory/sfl_patch
+	display_name = "Marine Corps Foreign Legion patch"
+	path = /obj/item/clothing/accessory/solgov/smc_patch/sfl
 	cost = 0
 	allowed_branches = list(/datum/mil_branch/marine_corps)
 
+///datum/gear/accessory/smc_patch_xenos
```
MarinaGryphon: Don't comment out old code, remove it entirely. It's still in the git history if it needs to be readded later.

PR #497, opened 2020-03-03
BO Science Skill

Karl-Franz-I:
Bridge Officers need science skill to properly scan using the sensor console. This allows them to choose whatever rank they want, like the shuttle pilot.

### DIFF
--- a/maps/torch/job/command_jobs.dm
+++ b/maps/torch/job/command_jobs.dm
@@ -379,7 +379,8 @@
 	min_skill = list(   SKILL_BUREAUCRACY = SKILL_BASIC,
 	                    SKILL_PILOT       = SKILL_ADEPT)
 
-	max_skill = list(   SKILL_PILOT       = SKILL_MAX)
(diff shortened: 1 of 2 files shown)


maps/torch/job/command_jobs.dm
```diff
@@ -379,7 +379,8 @@
 	min_skill = list(   SKILL_BUREAUCRACY = SKILL_BASIC,
 	                    SKILL_PILOT       = SKILL_ADEPT)
 
-	max_skill = list(   SKILL_PILOT       = SKILL_MAX)
+	max_skill = list(   SKILL_PILOT       = SKILL_MAX,
+						SKILL_SCIENCE     = SKILL_MAX)
```
Roland410: Idk if because of mobile or not, but seems like an indentation issue here.

Karl-Franz-I: Double checked, it should be fine

PR #502, opened 2020-03-04
Tweaks away site and overmap spawns.

Roland410:
🆑Roland410
:tweak:Raises exoplanet and overmap feature budget which means more away sites and ruins (not necessarily joinable ones) CAN spawn. Nothing is guaranteed.
:tweak:Except now two planets will spawn, go ham and explore those.
:tweak:Inhabited colony can now spawn on shrouded and volcanic planets. You asked for this, but suffer the consequences.
/🆑

For the love of god, please testmerge this as this will surely mean a longer startup time and can cause performance issues with the two planets in the long run.

### DIFF
--- a/code/modules/overmap/exoplanets/exoplanet.dm
+++ b/code/modules/overmap/exoplanets/exoplanet.dm
@@ -41,7 +41,7 @@
 	//Flags deciding what features to pick
 	var/ruin_tags_whitelist
 	var/ruin_tags_blacklist
-	var/features_budget = 4
+	var/features_budget = 5
 	var/list/possible_features = list()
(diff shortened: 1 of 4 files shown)


ProbablyCarl:
Primary issues from planets rise from fluids. Hence why we disabled the chlorine exoplanet.

Roland410:
Not sure what's actually classified as a fluid by code, but the only one I can think of is the lava in the volcanic planet?
Other than that and the VERY small chance that a second sun aka. a planet with a burning atmosphere spawns, this should be quite stable.

My only concern is general resource use, that's why I recommend test merging.

PR #505, opened 2020-03-05
EMERGENCY FIXES FOR EXPLOIT

MarinaGryphon:
Ports Fail2Topic from Aurora and key cache checks for IsBanned from TG.

### DIFF
--- a/code/modules/admin/IsBanned.dm
+++ b/code/modules/admin/IsBanned.dm
@@ -1,8 +1,11 @@
 //Blocks an attempt to connect before even creating our client datum thing.
 world/IsBanned(key,address,computer_id)
+	var/global/key_cache[]
 	if(ckey(key) in admin_datums)
 		return ..()
-
+	if((key in key_cache))
+		return list("reason"="concurrent connection attempts", "desc"="You are attempting to connect too fast. Try again.")
+	key_cache[key] = 1
 	//Guest Checking
 	if(!config.guests_allowed && IsGuestKey(key))
 		log_access("Failed Login: [key] - Guests not allowed")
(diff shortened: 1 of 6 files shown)


AffectedArc07:
Notice, this only works on windows. If you host on linux this wont work

MarinaGryphon:
> Notice, this only works on windows. If you host on linux this wont work

Aware. We host on Windows, thankfully, otherwise we'd have to find an alternative to fail2topic.

deathride58:
This PR lacks proper detection for already-connected clients, meaning if you're hosting on 512, topic spam and connection spam will both cause all connected clients to disconnect with every packet.
A much more up-to-date baycode PR that handles this proper can be found here: NebulaSS13/Nebula#53 , which also includes a failsafe in the event that something in isbanned() runtimes, which will prevent cases where a client gets stuck in "connecting too fast" limbo

code/modules/admin/IsBanned.dm
```diff
@@ -1,32 +1,49 @@
...
+	if(key_cache[key] >= REALTIMEOFDAY)
+		return list("reason"="concurrent connection attempts", "desc"="You are attempting to connect too fast. Try again.")
+	key_cache[key] = REALTIMEOFDAY + 10 //This proc shouldn't be runtiming. But if it does, then the expiry time will cover it to ensure genuine connection attempts don't get trapped in limbo.
+
+	var/ckeytext = ckey(key)
+
+	if(admin_datums[ckeytext])
+		key_cache[key] = 0
+		return ..()
+	key_cache[key] = 1
```
deathride58: Setting `key_cache[key]` to 1 here with the expiry time in place means that the expiry time is always set to a world tick of `1`, which results in this passing through the expiry time check as seen above, which reopens the connection spam exploit.

MarinaGryphon: Oh nooo

PR #507, opened 2020-03-06
Deletes Sergal.

Sparrowbird611:
In the name of god, I, lord sparrow the fourth, cast down our unholy enemy, the foul sergals of the east, while impaling the southern tyrant upon my sword. From this day forth, we will crucify their kind, slaughter them at our feet, and bring about a new era of light!

What this actually does is removes eastern and southern sergals. Base sergal race is untouched, though I may screw with them in the future. If some one wants to play as an east or south sergal ~~I'd be surprised if any one knew serg lore~~ they have all the options in the base sergal race. This just deletes two redundant, barely different subspecies.

🆑
Removal: Northern and Eastern Sergals have been removed from the race list. Base sergal is untouched.
/🆑


Jeff-Dome:
Based change

Sparrowbird611:
Indeed it is. It's time to wipe our these foul curs! Join me, brethren, as we go on the LAST CRUSADE! ~~Or last until we find something else to crusade against~~

Sparrowbird611:
They had some slight differences, but they were incredibly minor. I believe eastern sergs were almost exactly like humans with claws, and southern sergs were less fragile than regular sergs, but still more fragile than humans, along with moving slightly slower than a serg. As it is, sergals aren't very different, but the other two variants didn't add anything.

PR #511, opened 2020-03-07
Promethean fixes and tweaks

Roland410:
🆑Roland410 and Lonefly
:tweak:Prometheans have received a few tweaks and bugfixes, read more below.
:tweak:Prometheans are more resistant to burn damage, less resistant to brute damage, have darkvision, and are a bit slower, additionally, they have got toxins resistance now, so they can't be one-tapped with an extinguisher.
:tweak:Prometheans now regen a lot slower, and can be toggled.
:bugfix:Prometheans can now select culture, homeworld, faction and religion.
:bugfix:Prometheans can now select hair and facial hair styles, and color them.
:bugfix:Prometheans now regenerate their limbs properly.
:bugfix:Prometheans have their brain placed in their chest now properly.
:bugfix:Prometheans now also gain hydration from eating from the floor.
:rscdel:Prometheans can no longer be put into FBPs or start as one (and it honestly is a terrible idea anyways) and can't have a lace, nor any assisted/robotic organs.
:tweak:Prometheans have been uncucked and can now play as every branch again, with only being able to be a trooper in the SMC because hurdur xenos.
/🆑

About 99% of the code is by @Loneguyfly I just made adjustments and tested it to work, sadly had to remove the heal timer because it was causing issues, but nerfed the health regen because of that.
I've been working on/bugtesting this for three days or so, but if anyone sees a bug, do let me know and I'll try to fix it ASAP.


Sparrowbird611:
Very nice. Might finally see a few more proms running loose.

bearrrrrrrr:
Yikes.

I'm really not a big fan of giving anyone a huge multiplier of 50% brute damage. I don't think that people *quite* know how big that is? ESPECIALLY given that Prommies (iirc) can't actually take organ damage like humans can, because, you know, no organs. (If this was changed, I didn't see anything in the files that did so.)

Part of the reason why Synths were nerfed was because they were 1man armies that could soak the fuck out of damage, had a measly 20% brute resist, and felt no pain. You are proposing a race that would be without the disadvantages associated with having fully robotic limbs, that would take even less damage from brute, and have even less damage from burn than IPC's/FBP's did. There's the response of 'muh water', which...voidsuits exist.

Like yes, Prometheans need a buff, but this is an attrocious balance change. I can maybe see this being okay if you let them take pain and heavily nerf resistances, but otherwise you're literally just making an IPC/Synth 2.0.

The bug fixes and everything else is super nice but...i really don't know about the balance.

Roland410:
Alright, they have received some tweaks after testing, and now they have an toggle-able aura, which is even slower than before. They can still regen limbs fast, as that's one of their main gimmicks.
Tweaked their brute mod to 0.8, burn to 1.2, brain health to 500.
I also managed to commit my DME, which is strange, because the file should exist on dev still.

Roland410:
Nevermind, @Sparrowbird611 forgot to change the DME in the Sergal removal, so going to keep that if noone makes a separate PR to remove it.

MarinaGryphon:
Haha what. There is literally no pretense of balance here. 20% brute resist? 500 chest HP? Reducing their 200% burn vulnerability to just 120% burn vulnerability? I, uh, really don't think this is even fit for a test-merge as-is.

Roland410:
I mean, I DID ask for a review, and the chest HP had to be raised to the brain HP's, because if their chest breaks, they can't regen that unlike other limbs. So if they receive ~100 damage iirc then they will be permanently fucked.

Roland410:
And the 2x burn mod was literally causing them to instantly die from even one hit laser in armor.

Roland410:
Removed the no pain flag from them and nerfed the regen a bit more along with brain and chest health.

Roland410:
@ProbablyCarl Could you review this please, considering noone else is going to at this rate?

Roland410:
That seems to have fixed the runtimes, but not sure what other effects it will have, probably nothing.

code/modules/species/species_shapeshift.dm
```diff
@@ -72,6 +72,8 @@ var/list/wrapped_species_by_ref = list()
 	wrapped_species_by_ref["\ref[H]"] = default_form
 
 /datum/species/shapeshifter/handle_post_spawn(var/mob/living/carbon/human/H)
+
+	wrapped_species_by_ref["\ref[H]"] = default_form
```
MistakeNot4892: As discussed on Discord, this needs to happen prior to spawning as the spawn logic relies a lot on retrieving appearance values and such from the wrapped species.

code/modules/species/species_shapeshift.dm
```diff
@@ -101,12 +103,34 @@ var/list/wrapped_species_by_ref = list()
 
 	last_special = world.time + 10
 
+	var/list/valid_hairstyles = list()
```
MistakeNot4892: Would probably be cleaner to use `generate_valid_hairstyles()` and override `get_hair_styles()` to take a mob as an arg so it can use the shapeshifter `get_bodytype()` override.

code/modules/species/species_shapeshift.dm
MistakeNot4892: Since you've returned it to pre_spawn you should probably nix it from post_spawn just to minimize any potential confusion.

Roland410:
Readded the code that changed their blood color and such, otherwise no changes. Waiting for merging/testmerging again.

Andromeda-K22:
Testmerging.

N8-Toe:
I still think 400 HP is to high, having a regen and MASSIVE health is still unbalanced imo

Roland410:
And I still think you should test it yourself then, as I have repeated the reasons about twenty fucking times by now.

code/modules/species/station/prometheans.dm
```diff
@@ -34,16 +34,19 @@ var/datum/species/shapeshifter/promethean/prometheans
 	blood_volume =        600
 	min_age =             18
 	max_age =             125
-	brute_mod =           0.5
-	burn_mod =            2
+	brute_mod =           0.8
+	burn_mod =            1.2
 	oxy_mod =             0
-	total_health =        640
+	total_health =        400
```
BurgerLUA: Me and many other staff feel that this number should be significantly reduced.

Roland410: Alright, so I'll explain one last time. The 400 health they get is their brain health. All damage they take to their chest has a 100% chance to hit their slime core aka their brain. The only thing that changes this is the brute and burn modifier + armor. They do not have the luxury of falling into crit like humans do. They do not have the chance to get cloned, ever. If they die, their core instantly gets deleted. Whoever thinks this is unreasonable, please go and boot up a server, or go into thunderdome or sth, and test it. This is basically necessary for them to be balanced, unless I implement a huge miss chance on their brain organ, in which case even with lowered brain health, they WILL be unbalanced.

Jaraci:
Prometheans are still super janky and the health modifier is totally irrelevant regarding their current state in this test merge. Seriously the number could be 2000 and it would be totally irrelevant:

- One single extinguisher blast transfers 120u of water into a Promethean, inflicting 1200 (!) toxin damage which goes directly to the brain, ensuring death even through the regen aura
- A neural-laced Promethean's lace goes into the head, not the chest, killing them on decap even though their brain/core can be totally untouched in the process
- No indicator other than code-diving that the way to regenerate nutrition is to walk on blood/oil/whatever spatters on the floor

Given this and a ton of other stuff that I can't quite marshal together into a coherent comment at 5am as of the time of writing (PM on Discord for details, Omi#9999), the 600 health really is not a problem considering how horrendously weak a one-shot and utterly guaranteed _permadeath_ can be dealt to them by 40u or more of water.

code/modules/species/station/prometheans.dm
Jaraci: Even if you could miss their brain with chest hits, they would still be overkilled three times over by a single extinguisher blast. There is no fear of them ever being overpowered while their absolutely obscene weakness to water is still in the code and unmitigated.

Roland410:
What Jaraci/Omi said above is fairly accurate (although didn't test injecting water too much) except the fact that water actually knocks them down and stuns them, on top of giving them a HUGE stack of confuse (you can easily get 5 minutes of confusion with barely any water), so if you spray a prommie with water then chuck them under a shower, they can't get out without outside help and will die in about a minute even if they have max nutrition amd regen on.

Roland410:
By the way, thanks Omi for actually testing it yourself and forming an objective response.

Jaraci:
In their current state, even as a stopgap measure, I would strongly suggest a `toxins_mod` of at most 0.33 to prevent permakills from a single click. They will still get dropped and suffer severe brain damage + confusion from being hosed down with an extinguisher, but at least they won't be guaranteed permanent death. Given that it's impossible to clone them right now and that their weakness to water is piggybacking off code that only seems to have ever been intended for simplemobs, this is probably the quickest hackiest fix that you can include in this PR without more extensive code work to make their water weakness severe/crippling but not 100% round-ending.

code/modules/species/station/prometheans.dm
```diff
@@ -82,57 +167,19 @@ var/datum/species/shapeshifter/promethean/prometheans
 					"<span class='notice'>You glomps [target] to make [G.him] feel better!</span>")
```
Jaraci: Should be "glomp", not "glomps"

Jaraci: also I'm not good enough with Git to figure out how the fuck I can quote the lines in this fancy review inline format thing when you didn't specifically change them, but this section at the bottom needs a space at the start of every string:

```switch(stored_shock_by_ref["\ref[H]"])
		if(1 to 10)
			return "[G.He] [G.is] flickering gently with a little electrical activity."
		if(11 to 20)
			return "[G.He] [G.is] glowing gently with moderate levels of electrical activity.\n"
		if(21 to 35)
			return "<span class='warning'>[G.He] [G.is] glowing brightly with high levels of electrical activity.</span>"
		if(35 to INFINITY)
			return "<span class='danger'>[G.He] [G.is] radiating massive levels of electrical activity!</span>"
```
since otherwise it looks very bad with no space.
image

Roland410:
```suggestion
					"<span class='notice'>You glomp [target] to make [G.him] feel better!</span>")
```

Roland410:
Updated this so Prometheans can no longer choose robotic internal organs and they can't use a lace anymore either as it was being really buggy.
Additionally had to give them a 0.1 toxins mod, as a direct hit from a fire extinguisher was a guaranteed kill, causing 1200 toxins damage which is brain damage for them as well. With these numbers, you can still kill them with a fire extinguisher, but it'll take a few more "shots", approximately 4.
Going to attach this picture as well:
image

Roland410:
This is still ready to merge by the way.

BurgerLUA:
Last call for anyone to complain about this PR. I am about to merge this if there are 0 objections.

PR #512, opened 2020-03-07
Veteris Unathi Improvements, or Why the fuck did I even add Veteris Unathi

BurgerLUA:
Ok so now the Unathi can wear most suits, however since NanoTrasen doesn't really give a fuck about your species, you have to apply a unathi helmet and then a human body for you to wear it.

Also Veteris Unathi can no longer wear human gloves, and must use modified gloves.

This PR has drained me.


Sparrowbird611:
.....…...………..I've really gotta ask.

Is there a reason why you didn't just make them use the unathi suit options, instead of both a human suit and an unathi helmet? This seems needlessly tedious, even by suit cycling standards. Also, I kind of assumed the cycling was more for the tails, not the legs.

BurgerLUA:
> .....…...………..I've really gotta ask.
> 
> Is there a reason why you didn't just make them use the unathi suit options, instead of both a human suit and an unathi helmet? This seems needlessly tedious, even by suit cycling standards. Also, I kind of assumed the cycling was more for the tails, not the legs.

Because Veteris Unathi have a closer bodytype to humans than Unathi.

Sparrowbird611:
Kind of seems needlessly painful to need to only swap the suit and not the helmet, as well as rather awkward, since most people would assume you'd need to have both an unathi helmet and an unathi suit for this to work, and not some confusion where you need a human body and an unathi helmet. I think it'd be better off keeping it consistent with either using a full unathi suit or a full human suit, but this is your PR so.

MarinaGryphon:
Personally, I think the mismatched suit/helmet stuff is actually really cool and unique. Code-wise this PR seems solid, I think.

PR #513, opened 2020-03-07
turns the panic bunker off

Andromeda-K22:
title.


MarinaGryphon:
Nope. Nope nope nope nope nope. Please, PLEASE just make it an actual config option so it can be changed in the future without additional PRs.

Andromeda-K22:
Yeah i'l,add that soonish

PR #514, opened 2020-03-07
Fixes Wheelchairs and other vehicles transcending space and time.

BurgerLUA:
Basically ports https://github.com/Baystation12/Baystation12/pull/28003

If you try to move up or down in your vehicle, you'll just get off.

### DIFF
--- a/code/modules/vehicles/bike.dm
+++ b/code/modules/vehicles/bike.dm
@@ -133,6 +133,11 @@
 		to_chat(user, "You unbuckle yourself from \the [src]")
 
 /obj/vehicle/bike/relaymove(mob/user, direction)
+
+	if(direction & (UP|DOWN))
+		unload(user)
+		return
+
 	if(user != load || !on)
 		return
 	if(user.incapacitated())
--- a/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
+++ b/code/game/objects/structures/stool_bed_chair_nest/wheelchair.dm
@@ -29,6 +29,9 @@
 	..()
 
 /obj/structure/bed/chair/wheelchair/relaymove(mob/user, direction)
+	if(direction & (UP|DOWN))
+		unbuckle_mob()
+		return
 	// Redundant check?
 	if(user.stat || user.stunned || user.weakened || user.paralysis || user.lying || user.restrained())
 		if(user==pulling)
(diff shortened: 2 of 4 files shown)


Sparrowbird611:
You can't tell me that interdimensional wheel chairs aren't a feature.

Jokes aside, this is good, considering the fact that you could literally phase through walls and people with these things.

Roland410:
Won't this mean you also get unbuckled every time you move up or down with the elevator or stairs? Also, does the spacebike retain its proper functionality, considering you use those to fly in space and between decks?

MarinaGryphon:
> Won't this mean you also get unbuckled every time you move up or down with the elevator or stairs?

No, that uses `Move()`, not `relaymove()`.



> Also, does the spacebike retain its proper functionality, considering you use those to fly in space and between decks?

The spacebike will probably be broken by this fix as-is, so that needs to be addressed, yes.

code/modules/vehicles/bike.dm
```diff
@@ -133,6 +133,11 @@
 		to_chat(user, "You unbuckle yourself from \the [src]")
 
 /obj/vehicle/bike/relaymove(mob/user, direction)
+
+	if(direction & (UP|DOWN))
```
MarinaGryphon: Space bikes are supposed to be able to move up and down in space. Instead, I believe there are some other checks that would be applicable here.

MarinaGryphon:
```suggestion
	if(direction & (UP|DOWN) && !isopenspace(get_step(user, direction)))
```
Apparently we don't have an ACTUAL multiz adjacency/reachability check. Huh.

MarinaGryphon: Nvm. Am big dumb.
```suggestion
	if(direction & (UP|DOWN) && !CanMoveOnto(src, get_step(src, direction))
```

MarinaGryphon: Nvm. Am bigger dumb. Don't use that one.

MarinaGryphon: May also cause issues with nonconnected Z-levels suddenly becoming connected if there's an openspace on the bottom or top level. That would be bad. This PR is fine as a stopgap fix.

BurgerLUA: ok so which one do you want me to use

MarinaGryphon: Neither, use the one you have.

MemedHams:
duplicate the current wheelchair before fixing it and rename the copy "Bluespace Wheelchair", this is too excellent a capability to not be preseved

PR #515, opened 2020-03-09
Removes E1s, trainees are now E2s/E3s.

GrayRachnid:
E1 signifies that they are still a recruit in bootcamp that hasn't even finished basic training. No E1 would be deployed on a ship so far away from base. Also makes Forensic Techs E4 because no Forensic tech would be a Forensic tech without finishing their apprenticeship.


Jeff-Dome:
Then this begs the question, of why bother having E1 ranks in the first place

GrayRachnid:
> 
> 
> Then this begs the question, of why bother having E1 ranks in the first place

All the ranks are in the code. The ones that don't apply are disabled.

PR #518, opened 2020-03-09
Ports Burgerstation Time Formatting

BurgerLUA:
Time formatting is limited for byond and causes issues with daylight savings time with how everything is setup. This fixes that without having to write snowflake daylight savings time checking code.


BurgerLUA:
Why is this high risk?

MarinaGryphon:
Testmerge successful. Questionable stylistic choices (`0:20` instead of `00:20`, really??) but a sound PR that fixes the shitty timekeeping we've been having. Good work.

PR #523, opened 2020-03-10
adds pill packs

Andromeda-K22:
🆑 Adds packs of pills, to be added to first aid kits / pouches, maybe? Use in hand to pop a pill out, examine shows the pills they contain. 

To be expanded to chemistry, for now, DNM.

🆑


baystation12.dme
```diff
@@ -1106,7 +1106,7 @@
 #include "code\game\objects\items\weapons\storage\storage_boh.dm"
 #include "code\game\objects\items\weapons\storage\toolbox.dm"
 #include "code\game\objects\items\weapons\storage\uplink_kits.dm"
-#include "code\game\objects\items\weapons\storage\wall_mirror.dm"
```
Andromeda-K22: `#include "code\game\objects\items\weapons\storage\wall_`

Roland410:
Soooo, pill bottles you mean? Seems like a redundant item that does the same thing, but not reusable?

Sparrowbird611:
What roland said. What's the point of this, exactly? Chemists already have like two boxes of pill bottles, and pill requests are rare in and of themselves. Seems like needless bloat.

Andromeda-K22:
These are slightly easier to use than pill bottles, also, this was requested by RenGusta and is a precursor to another item which is not intended to contain pill bottles.

This is to be used in situations where giving an entire bottle of bills (often +200 units of a medicine!) on prespawn would be overpowered.

Roland410:
Or you could just subtype the pill bottle/preassign the amount of pills it has. This is basically code bloat that can be solved with existing items way easier.

Andromeda-K22:
I'm unwilling to consider 87 lines of code 'code bloat'.

MistakeNot4892:
Possible alternative implementation in the PR linked above.

I quite like these, the little pop is a lot more satisfying than just getting some loose pills in a pouch.

BurgerLUA:
I'm waiting for the label to be removed before I can start reviewing.

PR #525, opened 2020-03-10
Genemod trait tweaks and additions.

Sparrowbird611:
Does some rebalancing, renaming, and additions to genemod traits. I'll list these all here, and the reasoning for these changes.

**Tweaks**

Incredibly frail: Brute mod reduced to 2.5.
Reasoning: Who in the bloody hell made this 5.25. 2x damage is already a crazy mod, 5.25 is literal punch gibbing. Reduced so people don't die on impact with a plastic bottle.

Recovering addict: Renamed to "Weak Immune System," and given an updated description.
Reasoning: Genemods are supposed to be physical characteristics, not character traits and flaws.

Telepathy: Reduced to 2 point cost.
Reasoning: Telepathy has incredibly niche practical applications compared to other 4 pointers, and as such, has been reduced.

Muscular Atrophy: Boosted to 2 points.
Reasoning: I'm not gonna lie,  I don't entirely know how this one works. I'm assuming this basically means you're screwed if any one grabs you, and trying to pull stuff is going to tire you insanely fast, so I thought it'd be worth a point increase.

Thin Skin and Inflammable Skin: Both increased to 2 points.
Reasoning: 1.5 brute and burn mods are nothing to laugh at.

**Additions**

Two new neutral traits, herbivore and carnivore.

Carnivore: "For one reason or another, you're only capable of eating meat. Vegetables won't kill you, but they won't help you either." 

The Unathi's carnivorous trait, given to jean mods.

Herbivore: "You're only able to eat plants. Eating meat and other animal protein will poison you." 

Exactly the same as the way skrell work. Bean mods rejoice.

I'd like to add more traits in the future, but for now, I'm sticking with this. If you have any feedback, hand it over ~~and your wallets~~.


Jeff-Dome:
Muscular Atrophy, increases your strength to STR_HIGH flag, which compared to every human upon the ship which only has the STR_MED flag makes grabs MUCH harder to get out of but not impossible, unless your a walking plant aka the Diona, which gives you STR_VHIGH which coupled with CQB means your done for which is a cringe lel moment but it is held only that race, but given gene modders can have high strength flag and move at the same speed as humans is abit broken and wonky

Sparrowbird611:
That's actually muscular hypertrophy, not muscular atrophy. Muscular atrophy gives you STR_LOW.

I don't recall diona using the STR_VHIGH flag, as I thought it was STR_HIGH. I'll have to look at that some time.

Jeff-Dome:
I got the names mixed up but my point remains the same to be fair.

Sparrowbird611:
Eh, considering the grav adapted human can do the same with fewer downsides, I'm not worried about balancing hypertrophy.

PR #530, opened 2020-03-12
ZAS, GC optimisations, QDel hint fixes

MarinaGryphon:
Oh man this is a doozy.

Ports some ZAS optimisations from Aurora.
~~Also ports ALL OF SSGARBAGE_COLLECTOR FROM AURORA.~~


code/modules/atmospherics/datum_pipeline.dm
```diff
@@ -15,7 +15,8 @@
 
 /datum/pipeline/Destroy()
 	STOP_PROCESSING(SSprocessing, src)
-	QDEL_NULL(network)
+	if(network)
+		QDEL_NULL(network)
```
MistakeNot4892: Going through this PR to steal lines and noted this - `QDEL_NULL` already falsey-checks the arg.

MistakeNot4892:
`. = ..()` isn't the same as `..()`. `Destroy()` needs to return a valid qdel hint (which is what `. = ..()` is doing) for garbage collection to work properly.

BurgerLUA:
Interesting

BurgerLUA:
@MarinaGryphon Are you okay with this PR?

code/modules/atmospherics/datum_pipeline.dm
```diff
@@ -15,7 +15,8 @@
 
 /datum/pipeline/Destroy()
 	STOP_PROCESSING(SSprocessing, src)
-	QDEL_NULL(network)
+	if(network)
```
MarinaGryphon:
```suggestion
```

code/modules/atmospherics/datum_pipeline.dm
MarinaGryphon:
```suggestion
	QDEL_NULL(network)
```

PR #532, opened 2020-03-12
The Society PR.

Sparrowbird611:
Due to low crew moral, Sol Gov and NT have agreed to allow the crew to acquire clown and mime gear.

🆑
Addition: Adds clown and mime gear to xenowear. They aren't human, after all.
/🆑

### DIFF
--- a/code/game/objects/items/weapons/storage/boxes.dm
+++ b/code/game/objects/items/weapons/storage/boxes.dm
@@ -619,3 +619,25 @@
 
 /obj/item/weapon/storage/box/canned/tomato
 	startswith = list(/obj/item/weapon/reagent_containers/food/snacks/canned/tomato = 6)
+
+
+//Box of clown gear
+/obj/item/weapon/storage/box/large/clown
+	name = "Clown Kit"
+	desc = "Just looking at this box gives you the urge to honk."
+	startswith = list(/obj/item/clothing/under/rank/clown = 1,
+					/obj/item/clothing/mask/gas/clown_hat = 1,
+					/obj/item/clothing/shoes/clown_shoes = 1,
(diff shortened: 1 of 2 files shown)


Roland410:
Please no. This will be just another step toward LRP bullshit because "muh funneh clewn and mime".

Jeff-Dome:
Society

Sparrowbird611:
Ok, it's a little bit of a meme, but frankly, it seems like a little bit odd that the entertainer doesn't have access to this stuff. More options are always nice.

As for the LRP stuff, gives security another person to target. We all know that security watches the clown like a hawk watches it's prey. Frankly, I can't imagine it being worse than vox pariah, and I don't think this would be abused as often.

chaoko99:
be ready to revert it when it does though. Security doesn't have an actual reason to hawk the clown, as that is meta. LRP roles like these are bad, and cause issues.

Roland410:
It shouldn't be merged in the first place, this is an awful PR that even if it'd have legit uses, will just create more work for admins and annoy people.

Sparrowbird611:
Eh, I thought it'd be nice for entertainers to have more options, and let a rarely seen outfit get used a little more.

~~Also, can we just take a moment to reflect on the fact that we have both a sexy mime and a sexy clown outfit? Who allowed this to happen?~~

Sparrowbird611:
On the bright side though, none of you can accuse me of arming the clown with a banana. Just a clown horn.

Jaraci:
Meme PRs shouldn't be allowed.

Andromeda-K22:
No.

Sparrowbird611:
Can't say it's a meme PR if it did stuff, but this is fair.

PR #533, opened 2020-03-14
Adds instruments to the loadout.

Sparrowbird611:
Just lets you pick instruments in the loadout. Honestly surprised this wasn't a thing before hand. 

~~Yall can't say this is a meme PR, these actually make sense onboard the ship, unlike a clown and mime~~

🆑
Addition: Adds several instruments to the loadout, under utility. Talent not included.
/🆑

### DIFF
--- a/code/modules/client/preference_setup/loadout/lists/utility.dm
+++ b/code/modules/client/preference_setup/loadout/lists/utility.dm
@@ -98,13 +98,13 @@ modular computers
 	path = /obj/item/modular_computer/laptop/preset/custom_loadout/advanced
 	cost = 6
 
-/datum/gear/accessory/solawardminor
+/datum/gear/utility/instrument
 	display_name = "Instrument Selection"
 	description = "An instrument to play beautiful music. Talent not included."
 	path = /obj/item/device/synthesized_instrument
 	cost = 2
 
-/datum/gear/accessory/instrument/New()
+/datum/gear/utility/instrument/New()
 	..()
 	var/instruments = list()
 	instruments["Guitar"] = /obj/item/device/synthesized_instrument/guitar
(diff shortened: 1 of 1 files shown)


Andromeda-K22:
This seems like a bad idea.

Sparrowbird611:
Fun is a mistake. Delete fun.

code/modules/client/preference_setup/loadout/lists/utility.dm
```diff
@@ -97,3 +97,19 @@ modular computers
 	display_name = "laptop computer, advanced"
 	path = /obj/item/modular_computer/laptop/preset/custom_loadout/advanced
 	cost = 6
+
+/datum/gear/utility/instrument
+	display_name = "Instrument Selection"
+	description = "An instrument to play beautiful music. Talent not included."
+	path = /obj/item/device/synthesized_instrument
+	cost = 2
```
BurgerLUA: You should 100% increase the cost of this from 2 to something like 8.

Sparrowbird611: Why would I increase it to eight? It's supposed to be a fun little thing. Compared to something that actually gives an advantage, like a knife, I'm not seeing the reason.

BurgerLUA: > Why would I increase it to eight? It's supposed to be a fun little thing. Compared to something that actually gives an advantage, like a knife, I'm not seeing the reason. Hell, the toolset implant isn't even eight points, it's four. I see no good reason to increase the price.

Because 2 is extremely cheap for a device that makes very loud sounds on a server that regularly has 50+ players.

Sparrowbird611: Meh, I'll raise it to four, but I'm not going any higher than that. Eight is a high price for an item with little in the way of advantages.

BurgerLUA: Fine. If it ends up being abused it will have to be raised.

PR #538, opened 2020-03-14
Fixes almost everything wrong with the Job Whitelist PR aka The PR that almost made me resign as dev

BurgerLUA:
The PR had a fucking mess of issues that caused it to not work properly, even when it was disabled. This PR fixes it and puts it back up to fucking standard. I will need to make a follow up PR that fixes everything wrong with it when it's enabled, right now I don't have the mental energy to deal with that, so here is a fix that works in the meantime.


BurgerLUA:
This PR has been tested to the best of my ability. You can save your preferences, join roundstart, and it prevents unlimited slots from being open up. It should have 0 issues regarding everything else

PR #540, opened 2020-03-15
Disables fluids.

MistakeNot4892:
Requested by @Andromeda-K22 due to apparent lag issues. I have left the code commented in despite being in source control as this should be temporary and it would be a shame for it to be forgotten when the underlying source of the lag is found.


MistakeNot4892:
I have compiled this but I haven't tested it, I make no promises.

MistakeNot4892:
I don't think deleting the code is wise in this case because it will lead to the feature being left disabled permanently by pure inertia.

MistakeNot4892:
Tested, it's fine locally.

BurgerLUA:
Conflicts need to be resolved and I will merge.

PR #542, opened 2020-03-16
Limits varedit selection to prevent lag

BurgerLUA:
Limits varedit object selection to the client's view to prevent the lag of it trying to generate a list of every single fucking object in the world.


baystation12.dme
```diff
@@ -1458,6 +1457,7 @@
 #include "code\modules\boh\infantry\outfits.dm"
 #include "code\modules\boh\infantry\storage.dm"
 #include "code\modules\boh\infantry\vending.dm"
+#include "code\modules\boh\job_whitelist\jobwl.dm"
```
MarinaGryphon: idk why it was at the wrong place lmao

MarinaGryphon:
kind of :thinking: about using `world.view` instead of the mob's viewsize, cause you can modify it with admin verbs, but otherwise good

BurgerLUA:
Selfmerging.

PR #544, opened 2020-03-16
Readds Legalese

Roland410:
🆑Roland410
:rscadd:Readds Legalese, it ain't dying so easily.
/🆑


Sparrowbird611:
Absolutely based.

peeesh:
ok but why. why is legalese useful or needed

Roland410:
Legalese is mostly a gimmick-ish language for people who want to play an educated lawyer/bureaucrat type of character basically. It's just about as "useful" as any language other than ZAC, but this makes taking higher bureaucracy skill (experienced) have some meaning as well, unlike other selectable languages.

BurgerLUA:
The early polls for this PR looks good. Will be merging this tomorrow if it still has approval from the community.

PR #554, opened 2020-03-18
Nerfs IPCs. Again.

Torque4607:
Holy biscuits IPCs are more than slightly OP with how many shots they can take.

They take more brute and burn, we're shooting at a robot here. They ain't that tough.

### DIFF
--- a/code/modules/species/station/machine.dm
+++ b/code/modules/species/station/machine.dm
@@ -83,8 +83,8 @@
 		TAG_FACTION = FACTION_POSITRONICS
 	)
 
-	brute_mod =      1.1
-	burn_mod =       1.5
+	brute_mod =      1.6
+	burn_mod =       2.0 //Yep, increased again. Machines are easier to damage by shooting than squishy, also IPCS are OP.
 
 /datum/species/machine/handle_death(var/mob/living/carbon/human/H)
 	..()


Roland410:
Going to disagree with this for multiple reasons. Modifiers above 1.5x are fucking crazy high (prommies have/had 2x burn mod and one laser instakills them) and because their cell gets damaged easily by chest damage, and they aren't half as OP as you claim them to be.

Sparrowbird611:
>IPC
>OP

I'm gonna stop you there friend.

I've never had an issue dealing with an IPC, as a security officer or a traitor. Traitors run around with a grenade that's cheaper than a sleepy pen that can blast every IPC in a 20 tile radius, and security has access to tools such as shock taser mode, flash bulbs, and flash bangs.

A damage modifier like this? Really? That's absurd. IPCs are in a fine place.

Frankly, if there is one thing I would do, it's give the ballistic kit access to a shock magazine, (Is that a thing? If not it needs to be, and I'd gladly code that since I plan on touching those kits any way.) so people who took the ballistic kit can actually say they're doing something ~~Even though ballistics is just a weaker laser~~.

code/modules/species/station/machine.dm
```diff
@@ -83,8 +83,8 @@
 		TAG_FACTION = FACTION_POSITRONICS
 	)
 
-	brute_mod =      1.1
-	burn_mod =       1.5
+	brute_mod =      1.6
+	burn_mod =       2.0 //Yep, increased again. Machines are easier to damage by shooting than squishy, also IPCS are OP.
```
BurgerLUA: I would change this to 1.5 brute mod and 1.5 burn mod for now. The secondary issue of IPC powergaming is that they're a little tanky, while the primary reason is that no one knows how to shut down an IPC.

Torque4607:
Aero and Burger were saying things in MSay a few rounds ago, but now that Burger pinged people in announcements about how to dunk powergaming IPCs, this probably isn't needed. Probably. Mistake on my part opening this up.

PR #556, opened 2020-03-18
Adds a lazy check that prevents ERT from spawning where their ghost is also makes the ERT base bigger just in case too.

BurgerLUA:
See title.


BurgerLUA:
Selfmerging.

PR #562, opened 2020-03-19
changes fuckin uhhh diety

bearrrrrrrr:
cost on a lot of diety's stuff
so they don't have to spend 3000 years gathering power

cut down a lot of stuff by 20% or 30% or even 50% in some cases. a lot of diety happens in the last HOUR of the round which is kind of iffy so this is aimed at maybe getting the ball rolling faster

### DIFF
--- a/code/modules/mob/living/deity/phenomena/starlight.dm
+++ b/code/modules/mob/living/deity/phenomena/starlight.dm
@@ -1,7 +1,7 @@
 /datum/phenomena/herald
 	name = "Bestow Heraldry"
 	desc = "Turn one of your followers into a herald of your coming."
-	cost = 100
+	cost = 70
 	cooldown = 60 SECONDS
 	flags = PHENOMENA_FOLLOWER
 	expected_type = /mob/living/carbon/human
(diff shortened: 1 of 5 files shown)


code/modules/mob/living/deity/phenomena/starlight.dm
```diff
@@ -165,7 +165,7 @@
 /datum/phenomena/divine_right
 	name = "Divine Right"
 	desc = "Trigger your rebirth into the body of someone wearing a herald's uniform. This takes time, requires 3 open gateways, and if the body is destroyed during the ritual... so are you. But once complete, you become an unstoppable demigod of unnatural power."
-	cost = 300
+	cost = 200
```
Roland410: Probably should change the ritual time from 10 minutes to something more sane like 5 at most, because with time dilation it can easily become 15 minutes as it did last round. It's only a few lines below this on line 201.

Roland410:
Good change overall. Most of the issues of deity is that you need to make altars everywhere to get enough power for the deity so you can actually do cool stuff.

code/modules/mob/living/deity/phenomena/starlight.dm
bearrrrrrrr: OOF, good point tbh. 

i'll add that when i merge completely unrelated fixes in a few days, dw

Sparrowbird611:
Do you really wanna know?

PR #564, opened 2020-03-19
Makes Prosthetics Render Properly On Custom Species/Genemodders

Michiyamenotehifunana:
:cl: Toriate
fix: Genemodders can now use prosthetic limbs properly
/:cl:

I had not realized that the game rejuvenates all characters on generation, and custom species' rejuvenation code forces any limbs to have the same iconstates as their iconbases. The rejuvenation code (for custom species only) now checks if a limb is robotic before force changing their iconstates.

BhF0eQUYno

### DIFF
--- a/code/modules/organs/external/_external_icons.dm
+++ b/code/modules/organs/external/_external_icons.dm
@@ -80,7 +80,7 @@ var/list/limb_icon_cache = list()
 
 	//BastionStation edit - START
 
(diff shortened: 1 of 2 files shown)


SilvernKell:
Oh wow. This would fix customspecies machine sprites working too. Good job. Could do with limitations to stop people taking negative traits that won't affect them though

PR #567, opened 2020-03-20
Up hint + Look Up/Down refactor

ghostsheet:
Ported and modified from aurora.
Comes in 6 Ui style.
Currently only avaliable for carbon/human and robot.
Works inside of mechsuits.

Refactored Look up and added look down verb
They can be use by all mob/living.
They're less buggy and checks if you can actually see through the floor.

klix4Ka62K


BurgerLUA:
Seems to change a lot of other things as well.

BurgerLUA:
Checks are failing and I cant' restart them. Make a dummy commit to get the new checks.

BurgerLUA:
Changelog is missing, I think.

PR #571, opened 2020-03-20
Removes the low Yield EMP grenade from the equipment kits.

Sparrowbird611:
Does what it says on the tin. Removes those grenades from both energy kits.

Why it's good for the game:

First off, there's far too much friendly fire potential on these things. There's IPC crew men, borgs, random machines, and people with prosthetic organs walking around, and you're telling me it's a good idea to throw robo killer nades? Yeah, no. Secondly, makes dealing with IPC antags a little harder than just "Lemme throw my memenades out."


RenGusta:
Replace with one box of 7 in soft armory and i think this is a good change. It's not handed to you instantly, and you'd only grab it if you see plenty of IPCs, or hear of one starting trouble, and also don't have to go straight to Ions.

Sparrowbird611:
Ehhh…. I disagree. IPCs can be taken down quite easily just by leg shots. Even the rubbershot rounds are capable of disable a leg in just a few shots. The EMP grenades are unnecessary. Security doesn't need a dedicated anti IPC grenade in the soft armory. The Ion gun in the hard armory is enough.

I will be testing shock in the future to see if it needs buffs or nerfs, however, in my experience, leg shots on shock are incredibly effective. The things you're suggesting I disagree on entirely.

bearrrrrrrr:
yeah agreed tbh

PR #572, opened 2020-03-20
Replaces the brig chief's kit with the regular kit.

Sparrowbird611:
It was pointed out to me while I had the code open how the BC kit is literally just rarely used stuff from the armory. So I thought, "Eh, might as well fix that too."

Deletes the BC equipment box, and replaces it with a regular one. To call this a nerf would be an over exaggeration at minimum, as the guns in these kits are seldom used, and the armory is filled with them.

This also gives the SEA the regular kit, because for some reason he used the BC kit.


GrayRachnid:
The energy variant has an electrolaser carbine with double the shot capacity. These boxes are also used by the SEAs

Sparrowbird611:
The SEA uses this kit? Huh. I suppose the question is then, does the SEA really need anything more than a pistol?

GrayRachnid:
> 
> 
> The SEA uses this kit? Huh. I suppose the question is then, does the SEA really need anything more than a pistol?
> 
> Also I really hope it's not double the shots. That's 48 shots.

The normal capacity is 12. The BC and SEAs get 24.

Sparrowbird611:
I'm 90% sure you're thinking of the stun rifle, which has ten shots. The stun carbine is a different, and actually inferior gun.

GrayRachnid:
> 
> 
> I'm 90% sure you're thinking of the stun rifle, which has ten shots. The stun carbine is a different, and actually inferior gun. Not to mention there's two of those inside the brig with the normal 24 shots.

The two in the brig have 12 shots. Go check them.

Sparrowbird611:
Alright, it turns out, you were right. I have no idea why they have two, almost exactly identical guns.

Any who, just going to edit the title and get feedback on giving the SEA a regular kit. Frankly, the man doesn't need a shotgun or carbine in my opinion, but we'll see what others say.

Sparrowbird611:
Thank you for pointing that out to me gray.

PR #577, opened 2020-03-21
Revert "Unblacklists head roles from antagonism."

Andromeda-K22:
Reverts BoHBranch/BoH-Bay#486


MarinaGryphon:
Commit has no content, somehow

MarinaGryphon:
oh I think it was already reverted lmao

PR #582, opened 2020-03-22
Adds some workflows.

BurgerLUA:
Adds some workflows that should help with automatic labeling.


MarinaGryphon:
Label still doesn't work on PRs, it's a known issue, just has to be removed.

BurgerLUA:
> Label still doesn't work on PRs, it's a known issue, just has to be removed.

Works on burgerstation. It's failing because I don't have access. When it's merged, it shouldn't have issues.

PR #585, opened 2020-03-22
Dylovene counters sleeping chems

Torque4607:
Now Dylovene will remove soporific and chloral hydrate from the bloodstream, same as it does with zombie powder. Counters sleeping meds, makes them less of a pain once you're in a doctor's hands.

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -93,6 +93,8 @@
 	var/remove_generic = 1
 	var/list/remove_toxins = list(
 		/datum/reagent/toxin/zombiepowder
+		/datum/reagent/soporific,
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
As some one who uses sleepy chems, and has no option other than to sit there and wait for them to wake up, this is great.

Torque4607:
Also makes it easier to recover when Urist McGreyshit makes ten soporific syringes and goes on a stabbing spree.

chaoko99:
Interestingly, it was already supposed to do this.

code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
```diff
@@ -93,6 +93,8 @@
 	var/remove_generic = 1
 	var/list/remove_toxins = list(
 		/datum/reagent/toxin/zombiepowder
```
Andromeda-K22: wait a sec you missed a comma here

PR #587, opened 2020-03-23
Halves handcuff and buckling escape times.

Higgin:
What it says on the tin - this has been getting obnoxious for how many years now? Tighten up the window within which your hostage-taker will have to babysit you for more opportunities to escape.

### DIFF
--- a/code/game/objects/items/weapons/handcuffs.dm
+++ b/code/game/objects/items/weapons/handcuffs.dm
@@ -15,7 +15,7 @@
 	matter = list(MATERIAL_STEEL = 500)
 	var/elastic
 	var/dispenser = 0
-	var/breakouttime = 1200 //Deciseconds = 120s = 2 minutes
+	var/breakouttime = 600 //Deciseconds = 60 seconds = 1 minute
 	var/cuff_sound = 'sound/weapons/handcuffs.ogg'
 	var/cuff_type = "handcuffs"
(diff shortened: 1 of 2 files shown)


Sparrowbird611:
Ok so, I'm gonna have to disagree on the unbuckling, but everything else I agree on.

code/game/objects/items/weapons/handcuffs.dm
```diff
@@ -137,7 +137,7 @@ var/last_chew = 0
 	name = "cable restraints"
 	desc = "Looks like some cables tied together. Could be used to tie something up."
 	icon_state = "cuff_white"
-	breakouttime = 300 //Deciseconds = 30s
+	breakouttime = 400 //Deciseconds = 30s
```
Sparrowbird611: Also, you ought to change this to say 40seconds now.

Roland410:
Getting bucklecuffed means you need to spend 4 minutes to get out. Moving, or being moved resets it. Having to spend 2 minutes in one place (or at least 1 minute for the buckle) while you can't even open public doors etc. has always been an awful design.

Sparrowbird611:
Eh.... I still don't think it ought to go below a minute and thirty seconds. I really would prefer not having to shoot those captives who just spam struggle in their chairs while I'm gloating at them.

PR #589, opened 2020-03-23
Brain thresholds tweak, headshot meta nerf.

Higgin:
Lowered threshold of brain health left at which certain levels of oxygenation (i.e. 30% - 60%) will stop doing damage **edit: and made the last two levels progress a bit faster vis a vis damage.**

Under the previous values you could practically hover above the level of oxygenation (30%) at which brain damage would finally kill you because it could only lower your brain health to 40%. 

This change effectively 'stretches out' the different progressions of low oxygenation effects and pushes that value down to 20% - below the 25% threshold at which your heart can stop and you can start the final descent. 

it wouldn't be me if i didn't put a pr in fucking with the combat/health/brains, but i think this might be the ticket to finally making serious injuries ultimately lethal without actually speeding up how long it takes to get there if you're really tweaked or making people made out of tissue paper. 

edit because i can't help myself:
nerfed the headshot meta. Now helmets matter; and you're only going to get confused at worst unless you're being executed or shot with something big.

### DIFF
--- a/code/modules/organs/internal/brain.dm
+++ b/code/modules/organs/internal/brain.dm
@@ -176,12 +176,12 @@
 					if(prob(1))
 						to_chat(owner, "<span class='warning'>You feel [pick("dizzy","woozy","faint")]...</span>")
 					damprob = owner.chem_effects[CE_STABLE] ? 30 : 60
-					if(!past_damage_threshold(2) && prob(damprob))
+					if(!past_damage_threshold(4) && prob(damprob))
 						take_internal_damage(1)
 				if(BLOOD_VOLUME_BAD to BLOOD_VOLUME_OKAY)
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
BIG fan of this. putting dexa/inaprov in someone with fading brain activity and then doing CPR while working on their brain was just an ez win-game cheat code.

Sparrowbird611:
Well, I've already said I'm against confusion effect on the headshots, but aside from that, I like everything else this PR does.

BurgerLUA:
Made a poll on this. Will merge if the poll looks good.

BurgerLUA:
The poll has overwhelming support. Will merge tomorrow if no one has an A+ argument as to why this shouldn't be merged.

PR #595, opened 2020-03-25
Removes Wizard LRP Spell

EldritchSigma:
Apparently somebody was faster than me and theres no single instance of energy/staff left in the wizards obtainable spellbooks. If the Staff of Change is still obtainable by other means please let me know because that kind of thing is bottom tier LRP and just a grief tool.

However Polymorph was missed. This removes polymorph from the clerics arsenal. Reason: Transformation spells/staffs/whatever are very bottom LRP and are almost exclusively used as griff tools.

### DIFF
--- a/code/modules/spells/spellbook/cleric.dm
+++ b/code/modules/spells/spellbook/cleric.dm
@@ -16,7 +16,6 @@
 				/spell/targeted/heal_target/area = 					1,
 				/spell/targeted/heal_target/sacrifice = 			1,
 				/spell/targeted/genetic/blind = 					1,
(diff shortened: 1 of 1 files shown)


Roland410:
I'm fairly sure that spell only transforms the wizard though?

Roland410:
Nevermind, that's the other polymorph (smh so many polymorphs).

Sparrowbird611:
I honestly can't see how polymorph is an LRP spell, and only the wizard can use it, as long as the effect is temporary. If it's permanent, then there might be an issue.

PR #600, opened 2020-03-26
Imports zipties

nerocavalier:
:cl: 
rscadd: Imports zipties: Restraints for sec to use on certain hardsuit rigs without having to run to get duct tape or cables from tool storage. 30 seconds to resist out of.
rscadd: Added 8 zipties to SecTech so ERT also gets them as well.
maptweak: Added a box of zipties to Sec equipment.
/:cl:


Sparrowbird611:
Uh, bro. Why didn't you just give them a box of cable cuffs.

BurgerLUA:
Zipties are actually fine and welcome. I don't understand why this PR has two downvotes already.

Sparrowbird611:
Well, my downvote stems from the fact that they're literally just cable cuffs. Why not, instead of adding a new, mostly redundant item, just add a box of cable cuffs to the brig and other places that need them. We don't need a whole new item to fill this niche when one exists.

BurgerLUA:
> Well, my downvote stems from the fact that they're literally just cable cuffs. Why not, instead of adding a new, mostly redundant item, just add a box of cable cuffs to the brig and other places that need them. We don't need a whole new item to fill this niche when one exists.

This is literally a non-issue. Makeshift cable cuffs do not make sense for police to have as it is a waste of materials. Zipties, being cheaper and have the same purpose, do.

Literally every single server has zip ties.

nerocavalier:
The original reason I wanted to add them was for flavor because I found the idea of Security having to use makeshift equipment for it to be silly.

Sparrowbird611:
Meh.... Honestly, I'd just rename regular cable cuffs zip ties, and give em those. I honestly just don't see a reason to add a new type of cuff, but at the end of the day, it doesn't really matter what security handcuffs ya with.

Jeff-Dome:
zip ties are faster to apply than handcuffs or aleast should be much faster to apply, given them aleast the use of a quick securing tool

Sparrowbird611:
Handcuffs already apply really fast, and a cuffed person has already lost the battle. Lets not make this stuff any faster.

PR #601, opened 2020-03-26
MAA access to guards vendor

nerocavalier:
:cl: 
tweak: Allows MAAs to access the arms/legs guard vendor inside of armory.
/:cl:
It allows them to color match their guards to their plate carrier instead of being given random colors.


Roland410:
This also allows them to get armor plates etc.

nerocavalier:
Yes, but they don't have access to vendor unless someone lets them in. By the time they're in, it's usually blue and they're allowed to have tactical plates.

There won't be the problem of people walking in at round start to take better plates.

Roland410:
I mean, fair point.

PR #603, opened 2020-03-27
Removes the meteor gamemode from being votable

Roland410:
🆑Roland410
:rscdel:Removes the meteor gamemode from being votable (meteors still exist otherwise).
/🆑
Updated the PR to only make this not be votable and as such can't roll in secret, as seemingly Carl and Mazian want to bus this round in or sth, idk. I just don't want this to be ever played, and if this is the compromise I must take, so be it.

### DIFF
--- a/code/game/gamemodes/meteor/meteor.dm
+++ b/code/game/gamemodes/meteor/meteor.dm
@@ -8,7 +8,7 @@
 	extended_round_description = "We are on an unavoidable collision course with an asteroid field. You have only a moment to prepare before you are barraged by dust and meteors. As if it was not enough, all kinds of negative events seem to happen more frequently. Good Luck."
 	config_tag = "meteor"
 	required_players = 15				// Definitely not good for low-pop
-	votable = 1
+	votable = 0
 	shuttle_delay = 2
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
This is an incredibly bad way to go about this.
You can disable the ability to vote round types. No reason to gut it entirely.

Roland410:
I know that you can make it not be votable and such, but I see no reason for this to exist. When it's voted/set it's basically extended because of the shields, or immediately fucks the round when a power outage happens.
I find zero fun in it, and having it even exist for admins to bus it in, is not something I agree with.

Roland410:
Also with the existence of the salvos, it makes it even less of a danger because that + shields means they mostly get destroyed before even being able to damage the shield, which means that outside of said power outage, they are unlikely that shields go down AND if it does, you can turn on one of the other three precharged ones.

Andromeda-K22:
Still a bad way to go about it .

I don't agree with this change at all.

Roland410:
Updated the PR to only make this not be votable anymore. Still not a fan of the idea of keeping this ever exist, but I just want to see it gone.

PR #607, opened 2020-03-28
renames thalers to something less (more) dumb

chaoko99:
🆑 chaoko99
rscadd: Resprited thalers.
rscadd: renamed thalers to Brouzouf
🆑 
"Brouzouf" is TEMPORARY. I don't have a good name yet.


chaoko99:
I know it's stupid but I consulted nobody for a name. It was GONNA simply be "dosh" at the inception of this.

MemedHams:
Ok- I understand that you went to the trouble of making this PR, but Thalers is a believable and satisfying name for money. Why change it?

chaoko99:
Thaler always felt unnatural to me in some way.   Someone in the discord put it well, calling it "hipster lore".

It's an ancient standard, https://en.wikipedia.org/wiki/Thaler, it's like making a new currency called "Drachmas"

chaoko99:
For that matter, they're not exactly weird colored dollars anymore, so they might as well get a name to match

Roland410:
Not a fan of the rename or the sprite (which looks like a crossbreed between an egg and a soda can).

From a code perspective, you have removed the 1000 sprite, but modified the 500, while removing both from code.

chaoko99:
I ran out of good color combos, so I reduced the maximum amount from the list. The 500... I didn't like, but at least felt obligated to put in the code.

MemedHams:
Just some advice: The name of a type of money should be short, two syllables at best, and should not have any significant emphasis that will make it difficult to say. Dollar(Do-lar) is probably the golden standard for this. It's extremely simple and literal to pronounce, being even shorter then it's written form, and has no complex noises that would make it difficult to translate over barriers of linguistic type, knowlege, and age. Brouzouf(Brouh-Zhoof) is not... well, great, for two reasons: It's pronunciation is long and emphasized, and it contains at least one complex noise(The guttural "Brouh"). It seems like a simple thing but the name and attributes of a monetary system have a pretty significant, if subtle effect on worldbuilding, and it can be difficult to find one that has good roll-off-the-tongue attributes.

chaoko99:
Reminder: The current name is BIG TEMPORARY until someone comes in with something better that isn't just the old timey name for dollar.

But not credit. Credit is boring.

MemedHams:
I didn't realize how interested I was in this sort of thing until you posted this. Scuse me while I go look up the names of money systems in history and popular fantasy.

chaoko99:
honestly actually I'd be totally down with calling them Drachmas

Sparrowbird611:
I preferred the other sprite tbh, but it's not terrible.

But what in GODS NAME is that name. Really man, leave it as thaler until you get a good one, don't literally name it "broseph."

PR #609, opened 2020-03-28
GCC/UCG lore compliance

N8-Toe:
Done on behalf of the GCC/UCG Dev. edits them to use new system factions name, as well as removing mentions of such

All mentions of Gaia, Terra, GCC, replaced.


N8-Toe:
forgot a changelog. will fix

N8-Toe:
right, was my fault for not testing enough. got it worked and Changelog to right file type.

N8-Toe:
right! fixed that, sorry

N8-Toe:
This was done at the request of the UCG/GCC lore dev. so I think that should count to remove the "lore change" tag. what further approval or check is needed for a merge

PR #611, opened 2020-03-29
Makes the abandoned, unused Ammo Dump into something more.

BurgerLUA:
Basically makes it a construction area for engineers to do projects on. Also solves an issue with security locking it down with airlock braces every single fucking round by adding more than 2 doors to airlock brace if they're going to that shit.


BurgerLUA:
Merge Conflict

PR #613, opened 2020-03-29
Nerfs Mindslave/Imprint

EldritchSigma:
They are detectable by medical scanners. 10 TC for basicly a permanent ally with no way out of it other than rolling RNG Biodegrade or getting (un)lucky and being EMPed. This should raise a couple eyebrows should a mindslave be scanned by medical.

### DIFF
--- a/code/game/objects/items/weapons/implants/implants/imprinting.dm
+++ b/code/game/objects/items/weapons/implants/implants/imprinting.dm
@@ -2,7 +2,7 @@
 	name = "imprinting implant"
 	desc = "Latest word in training your peons."
 	origin_tech = list(TECH_MATERIAL = 1, TECH_BIO = 2, TECH_DATA = 3)
-	hidden = 1
+	hidden = 0
 	var/list/instructions = list("Do your job.", "Respect your superiors.", "Wash you hands after using the toilet.")
 	var/brainwashing = 0
 


Roland410:
In my opinion, raise the price instead. The whole reason the hidden implant system was added was because of this. Once someone acted strange, they were sent to medical, opened up, and the implant got removed instantly.

EldritchSigma:
The vote from a few weeks back had the majority of players vote for "Remove Mindslave implants" followed by "dont remove them" (61 to 40). This is a compromise between removing them entirely and keeping them in.

Reason for this change is basicly that it is, depending on what the implanter does to you, its ABSOLUTELY ZERO fun for the implantee. You basicly get thrown into perma and considering secs interaction so far with people and consideration on their character records etc you are basicly tossed in there and forgotten. Not fun. This atleast makes implanters choose their words and actions carefully to not loose their "investment".

Roland410:
Screenshot_20200330-103231
According to this, first was remove, second is fine as-is, third is tied to raise price.
If you've read my comment above, you see that my suggestion is option 2, raise the price, although if we were to respect the poll, it should be removed.
As I said, what will happen when someone uses one, and the person starts acting strange, is that they will be sent to medical, and get the implant removed from them. Instead of that, the raised price will mean that you would be limited to two at most unless it's discounted, and you would still get the price of your investment back.

Mindslave implants are pretty meh, and I agree with their removal, but having them not be hidden is basically the same as removing them in my opinion, as noone will want to waste TC for it, since people will basically "legally" meta it every time.

Sparrowbird611:
I'd honestly just raise it to like, 25-35 TC. Mindslaves aren't a bad pick, but they're not particularly fantastic, since most things that another person can do can be done yourself with a cheaper pick, without having to go through the trouble of isolating them, spending another 14 TC to get a para pen or some other weapon to threaten them, and then implant them.

BurgerLUA:
making a poll about this tomorrow

BurgerLUA:
Poll is up. I forgot to set one.

Roland410:
Gonna post an alternative to this as well hopefully in the next few hours.

PR #635, opened 2020-04-02
Readds removed officer gunbox, returns said gunbox to brig chief

Jaraci:
Readds removed (not commented out, flat removed) code pertaining to the standard size gunbox with modified electrolaser + smartgun or beanbag + milpistol w/ rubbers. Also gives it back to the brig chief.

Essentially a near-total revert of #572. I don't know why the code was straight-up removed instead of being simply commented out, or just having the gunboxes in question changed inside the lockers, but I dislike that it's no longer available even as a spawn option for special occasions.

I also disagree with it being 'unnecessary' for the brig chief in particular. They are senior enlisted and a senior member of security, and I don't see why it was pulled from them in the first place. The SEAs losing it I can understand, though.


Sparrowbird611:
Yeah, that was my fault. In hindsight, removing it entirely may not have been the best idea.

As for the BC part, the guy has access to the armory, he didn't need to really spawn with the guns. However, I really don't care that much, but his current carbine should be replaced with the regular one.

Jaraci:
The box carbine was never removed, it's still in the code. I don't see why it needs to be replaced, it shouldn't cause any issues.

I think I broke one of the Git checks by renaming the PR so soon because I forgot to give it a proper title. Oops.

Sparrowbird611:
Mainly because the modified one has 24 shots compared to the regular one, which is ridiculously high.

Jaraci:
You've mentioned as such, yeah. I don't agree. I think the brig chief having access to advanced non-lethals, given his potential role in dealing with rowdy prisoners -- particularly violent ones confined to perma or long brig sentences -- is fine. If it were a las-carbine or a big revolver or something deadly like that I could see where you're coming from.

Sparrowbird611:
I actually see even less reason to use a rifle of all things against prisoners. If the BC is legitimately worried about getting jumped by prisoners, then he should use the pepperspray or flashbulb, two items that he's immune to, and aren't any where near as high on the escalation scale as a literal stun rifle. As it is, the brig is retardedly immune to prison breaks, what with the floor mounted flashers and stuff everywhere, and the stun pistol is already super effective at putting people down.

You could say he needs the carbine to defend the brig from possible invaders, but if there's some one invading the brig, the BC is going to run straight for the armory to get a shotgun or some other more powerful gun, not use the stun carbine.

Roland410:
Edit the map instead of this to change the boxes to the small one. Noone needs to have a sidearm+rifle on spawn, especially no the brig chief who has access to the ENTIRE armoury, with less than lethals starting in the equipment part which is not even access locked (outside of sec obviously).

Roland410:
Also it was removed completely because it's unnecessary, and we have Git to put back any such code. It's almost always preferable outside of testing to completely remove the unused code. Less confusion and less issues in the long run.

Sparrowbird611:
Really, what legitimate need does the BC have to spawn with it in his office? If he's that desperate, he can run and get the default stun weapons. Your only real case for it has been dealing with prisoners, and if you deal with prisoners with a rifle, you deserve to be fired.

Also, I can't really imagine why you'd want to spawn this box. It's basically the small box with a rifle/shotgun in it, you could just spawn in the stun carbine on the person.

Jaraci:
Prisoners are frequently allowed to keep their uplink by virtue of meta protection. Not every prisoner you're going to have is going to be unarmed.

There are only four advanced non-lethal options as it is, and they are frequently distributed ahead of the lethals to regular security + cadets anyway.

I've heard your arguments and I've already told you I disagree with them, you don't need to keep rattling them off again and again.

Sparrowbird611:
Alright, I'll stop, though I can't say meta protected uplinks is a fair argument either.

Higgin:
the invisible gunbox issue is fixed in #583 - mind merge conflicts 
this seems a bit redundant to me with the warden having a room full of easily-grabbed nonlethals right across the hall

Jaraci:
Should now be compliant with #583 -- a PR that changed a lot more than it advertised, but whatever. I don't really know how to work Git and I'm just winging this, so if I broke something please tell me.

There are only four advanced non-lethal options available in the SEC-ARM and two of them have just been slapped in the dick. I still don't see the issue with the brig chief getting something chunkier given their role, rank, and position on the ship from both an IC and OOC perspective.

PR #640, opened 2020-04-03
Revert "Removes Black Tar from Uplink"

Roland410:
Reverts BoHBranch/BoH-Bay#586
Vote was against removal, and this is highly underused as it needs two traitors to cooperate, and even then you only get one use of it.
Screenshot_20200403-064628

### DIFF
--- a/code/datums/uplink/stealthy_and_inconspicuous_weapons.dm
+++ b/code/datums/uplink/stealthy_and_inconspicuous_weapons.dm
@@ -44,3 +44,10 @@
 	desc = "One of the most dangerous chemicals you could think of, easily vialed up. Be extra careful not to drink it!"
 	item_cost = 64
 	path = /obj/item/weapon/reagent_containers/glass/beaker/vial/hfc
+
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
you don't need 2 traitors, you can wait for a discount. it can cost as little as 21tc
I'm not up for its perma removal but i'm not for keeping it in the state its in. it essentially just encourages deathmatching against unarmed crew. 0/10

Roland410:
Fair point, forgot that this can go on discount as well. Really should implement a blacklist for discounts, would be good for future use as well to blacklist powerful items from discounts.

AdamElTablawy:
It should stay out until reworked. As is being a zombie is fucking lame. Even TG does it better.

Sparrowbird611:
I definitely want it to come back, but I would be fine waiting until it was reworked before it did come back, as long as it definitely does come back. Current black tar is a little bit retarded, with the whole lack of a cure, and zombies who are barely better than baseline humans.

Roland410:
Well, to be fair this only removes it from the uplink, so gonna close this.

PR #641, opened 2020-04-03
Adds Adrenalin Implant to traitor uplink

EldritchSigma:
Adds an adrenalin implant to the traitor uplink for 25 TC. 

Why only 25 TC? 

It does jackall to pain. Pain is the major source of being stunned. This implant is a counter against usual one shot cheese methods people tend to employ like going for the head or going for the legs.
Being in pain is still an easy way to get stunned rather fast.

### DIFF
--- a/code/datums/uplink/implants.dm
+++ b/code/datums/uplink/implants.dm
@@ -40,3 +40,11 @@
 	set of law-like instructions to follow. This kit contains an autoinjector with a dose of Mindbreaker Toxin."
 	item_cost = 10
 	path = /obj/item/weapon/storage/box/syndie_kit/imp_imprinting
+	
+/datum/uplink_item/item/implants/adrenalin
+	name = "Adrenalin Implant"
+	desc "An implant with precisely 3 charges. On use it removes any and all none pain related means of crowd control. \
+	namely knockdowns, stuns and weakening. Does nothing against pain related crowd control. Usefull in case you get shot \
+	in the leg."
+	item_cost = 25
+	path = /obj/item/weapon/implanter/adrenalin
--- a/code/datums/uplink/implants.dm
+++ b/code/datums/uplink/implants.dm
@@ -43,7 +43,7 @@
 	
 /datum/uplink_item/item/implants/adrenalin
 	name = "Adrenalin Implant"
-	desc "An implant with precisely 3 charges. On use it removes any and all none pain related means of crowd control. \
+	desc = "An implant with precisely 3 charges. On use it removes any and all none pain related means of crowd control. \
 	namely knockdowns, stuns and weakening. Does nothing against pain related crowd control. Usefull in case you get shot \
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
I honestly think you should just make them injector pain killers tbh. This is pathetically niche, and when you remember that the stimpack is only 3 TC, I'm not hearing too much reason not to give this one pain immunity as well.

EldritchSigma:
How do I add a proc to inject like 7 units of oxycodone per injection?

Sparrowbird611:
I don't entirely know, but you might be able to use what changelings use.

	var/mob/living/carbon/human/T = changeling_sting(35,/mob/proc/changeling_paralyzedsting)
	if(!T)	return 0
	to_chat(T, "<span class='danger'>You feel a small prick and your muscles start to go lax!</span>")
	if(T.reagents)	T.reagents.add_reagent(/datum/reagent/vecuronium_bromide, 35)
	SSstatistics.add_field_details("changeling_powers","PRAS")
	return 1


T.reagents.add_reagent(/datum/reagent/vecuronium_bromide, 35) This seems like it could work, if you can get it to target the user.

BurgerLUA:
Compile errors.
```code/datums/uplink/implants.dm:48:error: desc"An implant with precisely 3 charges. On use it removes any and all none pain related means of crowd control. namely knockdowns, stuns and weakening. Does nothing against pain related crowd control. Usefull in case you get shot in the leg.": value not allowed here
baystation12.mdme.dmb - 1 error, 0 warnings (4/3/20 3:00 pm)```

File: code/datums/uplink/implants.dm [hunk 1]
```diff
@@ -40,3 +40,11 @@
 	set of law-like instructions to follow. This kit contains an autoinjector with a dose of Mindbreaker Toxin."
 	item_cost = 10
 	path = /obj/item/weapon/storage/box/syndie_kit/imp_imprinting
+	
+/datum/uplink_item/item/implants/adrenalin
+	name = "Adrenalin Implant"
+	desc "An implant with precisely 3 charges. On use it removes any and all none pain related means of crowd control. \
```
BurgerLUA, line 46: should be ``desc = "An implant..."``

Sparrowbird611:
Y'know what, you could also try and look at the chemical implant, and see what that one uses if you still want to make it inject oxycodone.

SilvernKell:
You probably don't want to make it inject Oxycodone. That makes you confused and slurry, fall down and such.

I would add tram. It's strong enough tbh. The more tram you have in you, the stronger it is. OD is 30 units. That would probably make you fine until extreme burns or a leg falls off and you can't walk anyways.

EldritchSigma:
Would add the chems if I frankly would know how to make it add reagents to self, for now figuring out what caused the compile error, most likely that blank space at the bottom that shouldnt be there.

PR #642, opened 2020-04-03
Changes the Adherent job list.

Sparrowbird611:
Little PR that gives Adherent the same job restrictions as Diona. Because if we can have a Diona surgeon, or, god forbid, a Diona paramedic, we certainly can have an Adherent one.


Sparrowbird611:
Also considering allowing Adherent into the NTEF. Thoughts on this?

PR #651, opened 2020-04-06
Revert "Energy projectile nerfs, plate carrier QoL, new gun options in kits."

Andromeda-K22:
Reverts BoHBranch/BoH-Bay#583

Sorry, but this is wildly unbalanced and is too hard of a nerf.


Higgin:
how do you feel about the shot count changes because i can readd those with the flashlight qol/gunbox stuff in a later PR

Roland410:
I think these nerf were alright, and removing AP means armor actually is worth something.

AdamElTablawy:
bad revert, don't do it.

change selectively what you dislike but dont bring back oldcarbines

BurgerLUA:
A poll is needed on this, especially since that there was a poll on the previous PR.

Andromeda-K22:
image

Andromeda-K22:
@Higgin you nerfed AP too hard.

Roland410:
It just means security can't gun down everything as easily. It still has 45 damage.

Andromeda-K22:
45 damage that immediately drops to like 5 or 6. The maglight that security
gets literally does more damage than a laser carbine against armor.

Andromeda-K22:
[image: image.png]
left to right:
normal sec armor (medium plate), heavy merc armor (avaliable from uplink),
syndicate voidsuit (avaliable from uplink).

I emptied a full laser carbine into each target.
NONE of the syndis took internal organ damage, and took roughly 10 damage
per shot.

10 damage.

45 damage.

Down to 10.

Yeah, that's 'fair and balanced'.

Roland410:
Why is our goal to kill? From my perspective, all of those will put a human into paincrit at least. The goal of sec shouldn't be murdering antags, but capturing them.
Causing internal organ damage should probably be the last thing/option of anyone ever. Sure, medical will not have to perform surgeries as often, but I'd say that's an acceptable drawback of this.

Andromeda-K22:
Because antags don't have mercy.

The antag's goal is not to capture security. It is to kill anything that
gets in their way.

Why should security be massively hamstrung about defending themselves?

This literally just encourages antags to stomp all over security - and 90
burn, is not enough to put someone in pain crit.

Roland410:
If an antag's goal is to kill everyone in their way, I have serious doubts about their competence. I'd say a minor amount of AP is acceptable, but not a full revert of the PR.

MemedHams:
Yeah I gotta say, "An antag's goal is to kill everyone in their way" is not something I'd actually agree with in HRP. Ultimately, in HRP, your primary objective as an antag is to make things interesting, by being the crew's foil. Mindlessly slaughtering everyone in your way is not interesting, which is why we don't allow reason-less murderboning.

Jaraci:
The laser carbines are (correct me if I'm wrong here) meant to be second-tier weapons that you escalate to on code blue/red to deal with more serious threats.

They deal _ten_ damage per hit to an armoured antagonist. _Ten._ Less than ten, on average, actually.

You're never going to hit all ten shots on someone, either, given weapon accuracy and how firefights go, so saying that the ninety-odd burn damage will paincrit someone (it won't until at least a minute or two has elapsed, by the way, I've tested it) is missing the point.

With the exception of the smartgun vs the military pistol, ballistics already outperformed their energy equivalents in almost every aspect bar projectile travel time. For some reason, Higgin (and maybe a few others) keep using the fact that lasers "can be recharged" as a benefit rather than an issue -- this isn't a pro, this is a massive con. Rechargers are not easy to find and you are hard capped to a very low number of shots per gun -- compare and contrast to a shotgun or Tectonic, which allow you to happily bring along the entire ammo box for 40+ shots if you so desire.

Slug nerfs were also snuck into the PR that's being reverted, and ones that make no sense at all to boot. Why are slugs, of all things, losing their armour penetration and gaining damage falloff?

Sparrowbird611:
I agree that the slug nerf and the laser carbine/laser cannon parts should be reverted. While I am of the opinion slugs do need some kind of nerf in the future, this was the WORST possible way to do so. The laser carbine can also be nerfed in the future, but I fear Higgins doesn't realize just how important AP is on guns.

Everything else though, such as the smart gun nerf, the flashlight QoL part, and the small laser nerf were more than acceptable, and should not be taken out.

Jaraci:
Sparrow has the right of it. The smartgun is still more than fine with ten shots and is by far the strongest basic kit around due to its fire rate etc (the stun revolver is useless in comparison), flashlight stuff is cute and warm, and the small laser doesn't need to be strong anyway (assuming that's what the smartgun uses?)

Higgin:
@Andromeda-K22 @Jaraci @Sparrowbird611 

I'll have a PR up tonight putting these specific reversions into effect.
+laser carbine beams back up to where they were
+slugs back up to where they were

If we're fine with the rest, it'll save everyone involved some trouble.

AdamElTablawy:
Since when has the antag's goal been to kill anything that gets in their way, and why is the argument 'why should security be hamstrung about defending themselves'?

Sounds like validhunting to me.

The consistent issue is that guns and armor are too strong. If people are saying 'well, now shotguns and the tectonic etc are better than carbines because carbines do no damage to armor' maybe it's time to change armor values and the power of those other guns. 

Security here has more firepower than they do on bay for some inexplicable reason, and they have nine officers to boot. If ballistics are so good because they're easily reloadable, give them a drawback. If every officer's gun is capable of one-tapping someone into broken chest territory, and there are nine officers, that's a bit overkill.

All of those design decisions make sense if you're trying to create an environment where the antag just murders everyone without saying anything, and security doesn't need to care about interacting with them because they're just a lol antag so kill them and move on, but frankly, if you wanted that, just go to TG.

Andromeda-K22:
The Tectonic is strong because it's a semi-auto shotgun with good armor
penetration.

Security here is the exact same size that it is on Baystation; there aren't
nine officers here, what are you talking about? Cadets aren't officers. Not
even remotely close. We have the same amount of masters at arms as Bay
does. I don't know where you're getting that number, frankly.

'just go to TG' isn't a valid arguement and never has been. This is the
process of balance, and the reality of things. When an antag is hostile,
they will not hesitate to shoot security dead. That's the long and short of
it.

Andromeda-K22:
Closing due to #658

AdamElTablawy:
@Andromeda-K22 

it doesn't matter if they're officers or not, they still fulfill the same role. shoot antag.

PR #655, opened 2020-04-06
Revert "Readds removed officer gunbox, returns said gunbox to brig chief"

Roland410:
Reverts BoHBranch/BoH-Bay#635
The whole gunbox is stupid and unnecessary when the BC has access to the armory. Letting them spawn with a combat shotgun or a stun carbine as such is redundant, when they can take one from there.

### DIFF
--- a/code/modules/boh_misc/boxes.dm
+++ b/code/modules/boh_misc/boxes.dm
@@ -63,27 +63,6 @@
 		qdel(src)
 
 /////////
-// Advanced Lockbox
-/////////
-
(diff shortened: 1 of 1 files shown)


bearrrrrrrr:
me make it able print weapons

me think it cool if no box, make people ask brig chef for thing :)

Sparrowbird611:
This is technically only the brig chief's special gunbox that starts with the beanbag shotgun and the electro carbine, not the security one.

PR #656, opened 2020-04-06
More glows

chaoko99:
🆑 chaoko99
rscadd: Shield generator, fire alarms, SMES, and air canisters have a glow now. Also the TEG but I kinda halfassed it a bit.

/ 🆑


chaoko99:
image

PR #663, opened 2020-04-07
Removes imprinting/mindslave implant from uplinks.

Roland410:
🆑Roland410
:rscdel:Removes imprinting/mindslave implant from the traitor uplink.
/🆑
This is an alternative to #613 which removes the mindslave implant from the uplink, but keeps it as an admin-spawn or mappable item. Currently the mercenary base should have a box of them I think, so outside of that, it's not in play anymore. I've let them keep it because it gives them more options should they go with infiltration.
Why? Even if the cost is unchanged in the other PR, making them not be hidden anymore means the implantee will always be dragged to medical if they act out, scanned, and the implant removed. If anyone came from Bay and at least read the PRs there, this whole feature was implemented exactly because of this. 
Link to the original PR for reference: https://github.com/Baystation12/Baystation12/pull/26158


Higgin:
Hate to say it - I'm not a fan of either this or #613 

The mindslave implant gets used so seldom as it is - and anybody in a position to have it used on them is already in trouble enough that I don't see why not have it be a more accessible, subtle item. 

Having it show up on scanners and cost a bit more is preferable to removing it outright from 99% of use-cases.

Roland410:
I've only opened this PR because #613 in my opinion is objectively worse. I'd rather not have it be a thing at all outside of adminspawn than to have it, use it on someone and let sec/medical legally meta it out. It's what happened a LOT of times on Bay, and if it becomes visible again, that's what'll happen here as well, I can assure you.

BurgerLUA:
Needs a poll as the previous version had a poll. Will give it another day or two while people see the current changes in action.

Jaraci:
Remove this fucking implant from the game before I absolutely lose my shit. If I have to deal with another mindslave implant related ahelp I will start simply closing the tickets on reception, because without being allowed to reveal IC in OOC in ahelp responses I have no other recourse for explaining to people why shitsec or even the CO him/herself is breaking SOP and shitting around on their day.

These things are absolute, weapons-grade, stage four ass and I have zero interest at all in using them, having them used on me, or participating in a round at all where they are being used on or by third parties.

EldritchSigma:
I stand corrected.

Apparently nerfing them/making them visible on medical scanners was not enough. Even if a person is scanned the implant isnt removed even if said person has records that say they dont have any implants.

Apparently people dont read records unlike I do. That is very sad.

DriedCrust:
Can confirm, regardless of wether the community likes this or not. This is an absolute nightmare to deal with from the staff side of things. The implants themselves has contibuted with alot of issues that we simply cannot adress due to the way their work. No, removing all protection won't work because it's extremely easy to hide it and not get scanned,

Roland410:
I mean, if staff agrees for a full removal, I'd gladly oblige. I hate them personally as well.

BurgerLUA:
The vote is extremely close, with  a 45 to support and 55 for against. Leaving the poll for another few days.

BurgerLUA:
Right. The community appears to be very split on the issue. My personal opinion is that removal of features, unless they're absolutely 100% confirmed shit, is bad. Right now mindslaving is 60% confirmed shit.

Unless you can convince 10 additional players that it's bad, I don't think this will be merged, but I will be bringing up policies that we can implement that would make it crystal clear what expectations are when it comes to mind slaving and antagging.

BurgerLUA:
Sorry. This is the first PR I'm going to close. The vote was close but it was in favor of not removing it.

PR #664, opened 2020-04-07
Antag Chem Buffs + Bone-Fixing Juice

Higgin:
Adds Hypeross-7 - bone-fixing juice with terrible consequences if overused.
Adds Hypeross-7, alkysine, and peridaxon autoinjectors to the antag uplink.
Gives stimpacks synaptizine and hyperzine - raises their cost.
Buffs lexorin and potassium chloride.
Buffs bicaridine OD's chance to heal arterial bleeding a lil bit.


EldritchSigma:
Can we not add hyperzine/synaptizine to stimpacks? 9/10 times this crap backfires and it already gives you pain immunity, a speed buff and fucking increased dodge. I rather leave it as it is than having a fucking delayed death in a bloody stimpack cause bay combat is still stun and grab focussed.

Higgin:
@EldritchSigma you know what, fair enough - more options are fun, I'll split it up a bit.

EldritchSigma:
Alright so codedived further. If ya want to make a super combat stimpack thingy (dunno on the name, Still think experimental stimpack is a good name), give it like.. .5 units of synapse (it lasts a LONG time) and I think it was 5 units of hyperzine. Add some tricord (5 units) to heal off minor physical wounds. That way you have the desired effect of a souped up stimpack. Have some downsides since it can literally stop your heart. But you wont suffer from acute organ failure cause synaptizine memes.

The bone chem looks nice, though I'd argue that 15 TC is a bit much considering the side effects on it. Also it uh.. lets just say every chem can be used malliciously. Its in a hypo. Hypo is an instant injector. What stops me from just using that on a sec officer that tries to arrest me while I am in a bloodred voidsuit (stun resistant). I would say make it an oral med. And a pill.

Alkysine should be very cheap, you can just steal it from medical. Peridaxon needs a buff in general tbf. Otherwise should be good from what I can tell.

Rayfod:
20u feels like a strange number to keep in the hypospray, I would recommend to make it 30u
As that would be the max for the hypospray.
Also yeah, it feels kind of like a thing that should be cheap-ish, it feels a bit steep with the price of the items.

EldritchSigma:
/datum/uplink_item/item/medical/zerk
	name = "Zerk"
	desc = "An autoinjector containing a powerful blend of toxic, combat-enhancing drugs. Comes with three doses - or one if you don't plan to come back."
	item_cost = 5
	path = /obj/item/weapon/reagent_containers/hypospray/autoinjector/zerk


God I love you. This is amazing.

3 Units of Synapse, 12 units of hyperzine.

This basicly turns the underwhelming "Only situationally useful" esword into a valid choice, you can close the gap quickly with those chems, are hard to stun and keep down. Combine with an eshield from cargo.. Yeah Khornite Joe the robber of blood comes soon.

Sparrowbird611:
Might be a good idea to add some dyovline into that injector. I don't know if hyperzine requires an OD or not to start poisoning you, but I don't quite like the idea of liver failure.

Higgin:
Synaptizine doesn't cause enough toxloss to give you liver failure even if you use the whole auto at once since it removes so slowly. 

It'll only mess you up if you're for some reason more susceptible to toxins i.e. already having organ damage, a higher tox mod, or faster metabolism.

What the tox damage it does do accomplishes is ensure you can't double it up with peridaxon without suffering confusion and drowsiness as the peri tries to treat the small amount of naturally-healing brain damage - balancing out its use for going hard all at once with its use for limited engagements after which you dip out to recover.

BurgerLUA:
Going to let feedback sit for another 3 or so days.

PR #665, opened 2020-04-07
Stun/Pain Punishment Reduction

Higgin:
Makes limb stuns proportional to the amount of pain inflicted over/under max damage. Reduces probability to make them no longer a guaranteed down on the first shot.
Makes slowdown from pain/damage a little bit less hardcore and more directly proportional to pain/damage.


code/modules/organs/external/_external_damage.dm
```diff
@@ -282,30 +282,31 @@ obj/item/organ/external/take_general_damage(var/amount, var/silent = FALSE)
...
 
 		else if((limb_flags & ORGAN_FLAG_CAN_STAND))
-			if(prob((agony_amount/max_damage)*100))
+			if(prob(drop_prob * (agony_amount/max_damage)))
 				owner.stance_damage_prone(src)
 				return 1
 
 		else if(agony_amount > 0.5 * max_damage)
 			owner.visible_message("<span class='warning'>[owner] reels in pain!</span>")
 			if(has_genitals() || agony_amount > max_damage)
```
Roland410, line 300: Why does this check even exist..

Higgin: When agony = max_damage, the probability of a stun = drop_prob. Over that, it multiplies up - under that, it multiplies down. 

Agony can go over max_damage - this makes the chance of a leg/foot-stun less under max_damage for the limb.

Roland410: if(has_genitals() is what I meant.

Higgin: more genitals = more pain

code/modules/organs/external/_external_damage.dm
```diff
@@ -282,30 +282,31 @@ obj/item/organ/external/take_general_damage(var/amount, var/silent = FALSE)
...
 			else
-				owner.Stun(4)
+				owner.Stun(3)
 				owner.drop_l_hand()
 				owner.drop_r_hand()
 			return 1
 
 /obj/item/organ/external/proc/get_agony_multiplier()
-	return has_genitals() ? 2 : 1
+	return has_genitals() ? 2 : 1 //from wikipedia the free encyclopedia
```
BurgerLUA, line 309: Cock and Ball torture (CBT), penis torture or dick torture is a sexual activity involving application of pain or constriction to the penis or testicles. This may involve directly painful activities, such as genital piercing, wax play, genital spanking, squeezing, ball-busting, genital flogging, urethral play, tickle torture, erotic electrostimulation, kneeing or kicking.[1] The recipient of such activities may receive direct physical pleasure via masochism, or emotional pleasure through erotic humiliation, or knowledge that the play is pleasing to a sadistic dominant. Many of these practices carry significant health risks.[2]

Jaraci:
This is just going to push security towards using lethals over non-lethals again. This neuters stun to a big degree even against unarmoured targets -- against anything in a red voidsuit or worse, you're looking at something like six leg shots to get them down, and that's in a testing environment against a stationary target with master weapons expert.

Security has no impetus to try using non-lethals which don't adequately do their job of incapacitating a target. It's hard enough to get shitsec to use the things as it is -- this is just going to make lethals (even more of) the go-to option for antags in voidsuits/stun resistant armour, which isn't something I particularly feel is a desired change.

Jaraci:
To illustrate my point, the stun rifle -- a weapon literally _designed_ to non-lethally incapacitate targets, and the strongest of its kind (as any sec player will tell you) -- can no longer down a red-voidsuited target owing to the effect of this PR and the one that fucking slashed the ammo count in half to a pathetic five shots.

image

Higgin:
@Jaraci I'm comfortable with the RNG, but fair enough - when I was testing with the stun rifle and merc armor, the first thing that came to mind was that it could use AP readded. Here's the retweak I did tonight:

At max_damage, you still floor. It's no longer so RNG-dependent. The weaken is 2 now instead of 4 which feels a lot less game-ending but still has enough potency for it to be useful in breaking up a fight/disarming somebody. Damage under max_damage is less immediately likely to down somebody, but it can be gotten there quick with stun weapons as they are.

BurgerLUA:
Travis is failing for some reason. Rerunning

PR #666, opened 2020-04-07
Fixes my screw ups on loadout restrictions

Sparrowbird611:
I accidentally screwed up storage items like webbing, so most roles that could pick it can't any more. This allows all roles to pick them, like the PR original intended.


Sparrowbird611:
Well what do ya know, PR 666. Any one got hexadecaphobia?

PR #668, opened 2020-04-08
Travis and TGM Fixes

BurgerLUA:
Fixes a travis issue caused by another PR, and finally converts all maps to tgm.


BurgerLUA:
rip map diff renders

PR #672, opened 2020-04-08
Potential Armalis Character generation bugfix

CrackParrot11:
Adding a de-active white list spawn flag for the Vox armalis. Also specifically added in an active dagon merchant job whitelist to the Dagon to hopefully fix the bug where the armalis are de listed from character generation despite having the SPECIES_CAN_JOIN spawn flag.

### DIFF
--- a/code/modules/species/outsider/vox.dm
+++ b/code/modules/species/outsider/vox.dm
@@ -138,7 +138,7 @@
 
 	slowdown = 1.5
 	hidden_from_codex = TRUE
-	spawn_flags = SPECIES_CAN_JOIN | SPECIES_NO_FBP_CONSTRUCTION | SPECIES_FLAG_NO_MINOR_CUT
+	spawn_flags = SPECIES_CAN_JOIN | SPECIES_NO_FBP_CONSTRUCTION | SPECIES_FLAG_NO_MINOR_CUT // | SPECIES_IS_WHITELISTED
 	brute_mod = 0.5
 	burn_mod = 0.5
 	strength = STR_HIGH
--- a/code/modules/species/outsider/vox.dm
+++ b/code/modules/species/outsider/vox.dm
@@ -138,7 +138,7 @@
 
 	slowdown = 1.5
 	hidden_from_codex = TRUE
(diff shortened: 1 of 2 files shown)


code/modules/species/outsider/vox.dm
```diff
@@ -138,7 +138,7 @@
 
 	slowdown = 1.5
 	hidden_from_codex = TRUE
-	spawn_flags = SPECIES_CAN_JOIN | SPECIES_NO_FBP_CONSTRUCTION | SPECIES_FLAG_NO_MINOR_CUT
```
MarinaGryphon, line 141: Not mentioned in the PR description.

code/modules/species/outsider/vox.dm
```diff
@@ -138,7 +138,7 @@
...
+	spawn_flags = SPECIES_CAN_JOIN | SPECIES_NO_FBP_CONSTRUCTION // | SPECIES_IS_WHITELISTED
```
MarinaGryphon, line 141: You add it but it's commented out? There's no reason for this line to exist.

code/modules/species/outsider/vox.dm
MarinaGryphon, line 141: Not mentioned in the PR description.

code/modules/species/outsider/vox.dm
CrackParrot11: I commented it out because it was changed previously. By me actually  the flag NO_MINOR_CUT was not on the Armalis initially so i reverted it back to its previous state. But joining the server again the Armalis are once again back on the character gen screen so this Whole pr is redundant.

code/modules/species/outsider/vox.dm
CrackParrot11: This was a change by me from the Pariah removal pr. i added in the Flag because after looking in the Species defines dm the the flag specifically said it was for species such as vox.  removing it back to its original state may would have been able to fix it. but once again Armalis are back on the character gen screen so this pr is redundant.

PR #677, opened 2020-04-10
removes meta box

chaoko99:
🆑 chaoko99
tweak: Traitor kits use regular box sprites now. This should help traitors get away with things for longer.
/ 🆑 
Why meta protect something when you can just make it inconspicuous.


Roland410:
Could also maybe make it so you can build boxes of this sprite with cardboard so it won't go unused. (And you can totally bait metagamers with it).

PR #682, opened 2020-04-10
More Adherent Job Stuff.

Sparrowbird611:
Removes Adherent from the NTEF. If some one wants to put them back, they can. I believe it was intended that Adherent were only originally supposed to be contractors, so this fixes that.

Additionally, blacklists security cadet and forensic technician, as I had forgotten to add those on. Because I can't really see an alien servitor in a job that's whole purpose is to screw people over.


bearrrrrrrr:
adherent shouldn't be big boy roles like physician anyway tbh

Sparrowbird611:
Closing, as per the new announcement.

PR #683, opened 2020-04-10
Robotic Limb Tweak

Sparrowbird611:
Simple PR with simple changes.

Robotic limbs take 25 DMG to reach malfunction threshold (Up from 10)
Robotic limbs have a 15% malfunction chance (up from 10%)

Feedback/criticism/autistic screeching welcomed. This PR should makes limbs slightly more durable, but breaking slightly more punishing.


SilvernKell:
Sounds like an elegant solution, I approve

PR #685, opened 2020-04-11
Replaces all non-math instances of "<<" with macros.

MarinaGryphon:
May wind up replacing the bitshifts with macros, too.


BurgerLUA:
Going to need a bunch of people to review this to make extra sure that this doesn't replace things that shouldn't be replaced.

BurgerLUA:
I took 3 days reviewing this and I saw nothing odd. Should be testermerged 100% though.

ProbablyCarl:
Supposed conflicts?

PR #687, opened 2020-04-11
Kitchen + Bar Remap, Honey Bottle, Condimaster tweaks

MarinaGryphon:
* Changes the header for condimasters/chemmasters to make sense
* Adds a honey bottle sprite
* Remaps the kitchen and bar


Roland410:
Pics? Don't trust mapdiffbot too much, doesn't show all changes usually. From what it shows though, not a fan of the stool in the corner, and the removal of those vending machines unless you placed it elsewhere. The kitchen in general feels too big for the very few things it has, especially in the middle. The microwaves feel out of place as well.

MarinaGryphon:
> Pics? Don't trust mapdiffbot too much, doesn't show all changes usually. From what it shows though, not a fan of the stool in the corner, and the removal of those vending machines unless you placed it elsewhere. The kitchen in general feels too big for the very few things it has, especially in the middle. The microwaves feel out of place as well.

* The microwaves were not moved.

* There are TWO sweatmax vendors in the exercise area already, and a hot drinks vendor in cryogenics. Both are barely ten tiles from where the ones in the galley were.

* The shutters block the view, to the left of the corner-stool is the drinks showcase (which is rarely, if ever, used) and in front of them is open counter space.

* Cooking needs LOTS of prep area to work well. This was done with feedback from chef/cook mains.

Pics coming soon.

MarinaGryphon:
image

Roland410:
Yeah, that's what I'm saying that the microwaves should be moved to the tables next to disposals imo, and people don't/shouldn't be sitting in front of the showcases.

MarinaGryphon:
> Yeah, that's what I'm saying that the microwaves should be moved to the tables next to disposals imo, and people don't/shouldn't be sitting in front of the showcases.

I can move the showcases down a tile so that the bar gets 4 seats and the kitchen gets 4 seats, what about that?

Regardless, I've never seen the drinks showcase used and the hot foods display is typically only used because people aren't taking the food people make, because they can walk over to the SweatMAX in the same room and buy an MRE instead. The kitchen ran out of room on the two-tile space they had before so the display was the only option.

Roland410:
I don't mind it if you move it, but main thing is that you'd still have people sitting on stools in front of it.

chaoko99:
Cereal machine's missing.

SilvernKell:
Hi! I see where you're going! But I think it needs some changes, too much dead space for me and cramped for 2 chefs with all those tables.

https://user-images.githubusercontent.com/13442831/79039536-7d4a5600-7ba7-11ea-8c37-be3bb29cb846.png

Here are some changes/thoughts I would like to suggest.

Reasons:
2 cooking stations, (left and right upper table row) with centralized cooking machines area in middle. Both chefs can access the cooking machines without intersecting one another all the time

I do not really like the flooring between bar and kitchen, and it looks strange if it doesnt have a wall. windows are nice because you can still interact and see bar

An extra table near the jukebox because why not? It is already an open area anyway. It gives a nice place for the cutlery minilathe to sit. Also the shutter and lightswitch on that wall is not very accessible to the chef when the shutters are down, a problem. So I suggest moving it to that wall near the bar door

The disposals is next to the tagger and away from the food, and the machine has lunchboxes and jugs. I think it makes a nice lunchbox prep area, for labelling them and deliveries and hanging out near the customers, with the condiments ready to pass over to them or use when prepping lunchboxes. 

Cheesey machines are better here, allowing another table and a sink that doesnt need two tiles dedicated to accessing it, or squeezing past tables that stick out

Important stuff:
-Don't forget the cutlery lathe, it's really nice to have that dedicated minimachine specially for it!
Make sure to stock it by plopping aluminum and plastic next to it
- Please please make sure that all showcases/fridges have emergency fire shutters there, I believe air can get through them (?) and they can cause big atmos issues

Thanks!

BurgerLUA:
Is this PR WIP?

MarinaGryphon:
> Is this PR WIP?

It is now that changes were suggested, yes.

chaoko99:
The vendors are missing?

PR #675, opened 2020-04-09
Renames nearly every gun.

chaoko99:
Gimme suggestions if you don't like the names I chose.
🆑 chaoko99
tweak: Renamed nearly every gun to a [Manufacturer] [Model number] [Nickname] structure. Please keep it this way.
tweak: Mini crossbow fits in your pocket again.
\ 🆑

PR #689, opened 2020-04-11
Makes dispersive blaster (Immolator Pistol) somewhat viable

EldritchSigma:
Gives the shitty incendiary laser blaster traitors can buy a self_recharge component.

Now its an actually okay weapon for the price. Also upped the ammo count by 1 because why not.


Roland410:
I believe the main point is to have people be distracted because "OH GOD I'M ON FIRE", but good PR still.

EldritchSigma:
Also updated the item description with the "weapon rename PR", might need to add that to the "weapon rename PR" found here: https://github.com/BoHBranch/BoH-Bay/pull/675

Sparrowbird611:
> I believe the main point is to have people be distracted because "OH GOD I'M ON FIRE", but good PR still.

While you are technically correct, fire isn't exactly something terrifying. It only really stops noobs in my experience, since a skilled player will either continue to curbstomp you, or pull one of the numerous fire extinguishers off the wall and then kill you.

PR #702, opened 2020-04-13
SRV Verne Rework: functional overmap science shuttle.

CrackParrot11:
unishi 1

unishi 2

unishi 3

Adds in a functioning Overmap shuttle for the SRV Verne. This shuttle should serve as a potential escape ship or a scouting science vessel for research and exploration. Given the context of the Verne in game, Science is off the table for most of the survivors and just want a way to escape and find safety.

### DIFF
--- a/maps/away/unishi/unishi-1.dmm
+++ b/maps/away/unishi/unishi-1.dmm
@@ -813,6 +813,7 @@
 /obj/item/weapon/rcd_ammo/large,
 /obj/item/weapon/rcd_ammo/large,
 /obj/item/weapon/rpd,
+/obj/item/weapon/rpd,
 /turf/simulated/floor,
 /area/unishi/engineroom)
 "bZ" = (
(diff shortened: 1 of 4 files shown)


BurgerLUA:
```
## UNIT_TEST ##: [MAP: Simple pipes shall not face north or west]: Following pipe had an invalid direction: The pipe (/obj/machinery/atmospherics/pipe/simple/hidden) ([0x200024d]) (the titanium bulkhead) (110,129,4) (/turf/simulated/wall/titanium)
## UNIT_TEST ##: [MAP: Simple pipes shall not face north or west]: Following pipe had an invalid direction: The pipe (/obj/machinery/atmospherics/pipe/simple/hidden) ([0x200028d]) (the floor) (112,127,4) (/turf/simulated/floor/tiled/techfloor)
## UNIT_TEST ##: !!! FAILURE !!! [MAP: Simple pipes shall not face north or west]: 2 simple pipes faced the wrong direction.
```

Pipes are weird on the map. Please also update your branch as I think a travis error from an old version is caused because of that.

CrackParrot11:
This pr is done on my end, but i am not to happy with some of it. I will make a new pr when i get the map to a point where i feel ok with how it turned out.

PR #703, opened 2020-04-14
Removed Straight jackets from the Voxship + other improvments.

CrackParrot11:
Added in a disposals outlet due to it being heavily requested. Also made the Vox TEG engine alot more forgiving and harder to destroy the ship. As well as added in a few extra flair items to the Dormitories and a space minimoog. This was due to Roland removing all Straight jackets from the game except for the Voxship, so now all straight jackets are removed from everywhere now hopefully.


BurgerLUA:
Same travis issues as your previous PR. Not sure why it's causing this on your PRs but not others.

PR #705, opened 2020-04-14
BGB buff, price increase

CHEtheGREAT:
basic BGB tweak to make it actually hurt people 

after a bit of testing, even if a player stands on the same tile as a BGB it rarely kills, delimbs, or causes any damage that can't be quickly fixed by medical. it lacks any sort of use especially for its price of 30tc; the BGB can certainly be used to disorient people but it lacks any other use, and even then unless they're within a 3x3 area people will be able to get up with some minor bleeding.

Price of 30 tc adjusted to 33 tc

Explosion size increased by about 35%


Sparrowbird611:
So, on average, how much more effective if this thing going to be?

BurgerLUA:
You should take screenshots of the explosion comparison of both so we know exactly what we're dealing with. Good places to experiment bombs with would be medical, command, or security.

CHEtheGREAT:
Will do, my private server testing has so far been bugged out so I'm struggling with it. I'll try to upload them soon.

BurgerLUA:
Travis is failing.

BurgerLUA:
Your respository seems to be deleted. Will be closing this PR in a few days if that does not change.

PR #707, opened 2020-04-16
Dissuades no fear edgelords from aiming at people when they are at gunpoint.

chaoko99:
🆑 chaoko99
tweak: aiming at people while you are being aimed at triggers a reaction shot now. Stop it.
/ 🆑


neroiscute:
Makes sense, Mexican standoffs can still be done with help intent. 👍👍

Roland410:
I disagree with this. If both people have their guns drawn already, it's fair game. Obviously if they unholster after being aimed at, that's different.

chaoko99:
I actually think if you're fast enough you can do that anyway, because it takes a moment for aiming to kick in.

BurgerLUA:
Travis failed because of bad bearcat stuff. Update to upstream/dev and.

PR #714, opened 2020-04-19
Adds MREs to general loadout

Aurelien3189:
:cl:Adds MRE selection to loadouts, a 2 point option with the player's choice of vegan, standard, or protein for those hungry mil-sims.
/:cl:
image
image


BurgerLUA:
2 seems pretty low for a loadout option that has a slight mechanical advantage of not having to spend credits on MREs or going to the kitchen

Aurelien3189:
I figured 3 would be fitting, but after some thinking and discussion, I've figured the following: MREs are incredibly common both as random drops alongside being found in large numbers in crates around the ship, and with the right lunchbox you can get the same nutrition provided by an MRE, alongside having a drink.

BurgerLUA:
failing travis, currently

PR #717, opened 2020-04-20
Voxship incinerator wall melting fix

CrackParrot11:
This Makes the walls on the Incinerator wall completely unmeltable. Baystation made the default Osmium-Carbide walls impossible to use for incinerators. So this allows the incinerator to be used again without causing a gas leak

### DIFF
--- a/maps/away/voxship/Vox_material.dm
+++ b/maps/away/voxship/Vox_material.dm
@@ -1,0 +1,61 @@
+/////////
+// Material Defines
+/////////
+#define GLASS_COLOR_ALIEN    "#3a5a9a"
+#define MATERIAL_GLASS_ALIEN "alien glass"
+#define MATERIAL_STROT_ALIEN "alien material"
+
+/////////
+// Alien Burnchamber Material - This will never be used by players, aside from off-site spawns.
+/////////
+/material/glass/alien
+	name = MATERIAL_GLASS_ALIEN
+	display_name = "durable glass composite"
+	wall_name = "bulkhead"
+	flags = MATERIAL_UNMELTABLE
+	stack_type = null
+	icon_colour = GLASS_COLOR_PHORON
+	integrity = 6800
+	melting_point = 16000
+	explosion_resistance = 1200
+	hardness = 500
+	weight = 500
+	construction_difficulty = MATERIAL_HARD_DIY
+	hidden_from_codex = TRUE
+	value = 100
+
+/material/alienwall
+	name = MATERIAL_STROT_ALIEN
+	display_name = "durable alloy"
+	wall_name = "bulkhead"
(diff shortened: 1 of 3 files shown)


Roland410:
This isn't a fix, this is introducing a very unnecessary and niche material just because people are incompetent. OCP has a melting point of 12k.

If people fucking up the engine is such an issue, then the problem is in the design, as you forgot to account for player stupidity.

By the way, for reference, the Verne uses 2xH2:1xO2 burn chamber for their engine as well, with OCP walls, yet noone complains about it being melted every time.

CrackParrot11:
Well, the incinerator still needs a way to properly view inside the chamber despite all the other Burn rooms also having one i can tweak it to allow to be viewed inside at least. To save people from the Headache.

BurgerLUA:
uhhhhhhhhhhhh struggling to see how this is a bugfix PR

ProbablyCarl:
The material was introduced for dealing with incinerator issues a _long_ while back. Given it's just another matter to deal with when playing off-sites. In most cases, it's fine, but OCP is a pain in the butt to deal with, as at times it just seems to randomly fall apart. As seen on the Verne.

I'm not going to comment further on it, as I'm against baby-proofing. But it's a thing to think about, regardless.

Roland410:
OCP melts at 12k iirc. It's their fault that they don't know how to properly make and do the burnmix.

If we want to babyproof it so much, just raise the melt temp of OCP instead of introducing a gimmick material that won't be obtainable outside of mapping, as it means that if the chamber gets damaged, they can only use regular mats to fix it.

ProbablyCarl:
I agree.
However, playing Devil's Advocate, one could argue that being so stupid as to allow it to get damaged, or having it damaged indirectly, is an arguably worse punishment. 
Given you can't rebuild it normally, without admin intervention, and will have to fall back on OCP.

Keekenox:
As somebody who has been playing on the vox ship for quite a while now, I haven't seen it melt or melted it myself once, it's super duper easy not to melt and you can still get ridiculous power out of it.

CrackParrot11:
Alright, i am going put the OCP walls back in and just add in two window viewing ports to allow you to see inside. This should allow people to be able to repair the walls in case it becomes damaged.

ProbablyCarl:
> Alright, i am going put the OCP walls back in and just add in two window viewing ports to allow you to see inside. This should allow people to be able to repair the walls in case it becomes damaged.

That works. Unburnable viewing ports with the walls remaining unchanged is a good compromise.

CrackParrot11:
Vox incinerator engine fixed
So, i put back in the OCP walls and left  the windows in as a viewing port, the Windows are unmeltable but can still be broken with brute force.

PR #720, opened 2020-04-20
Fluid optimisations from Nebula

MarinaGryphon:
https://github.com/NebulaSS13/Nebula/pull/342


MistakeNot4892:
Don't merge this, Neb one has stuff I need to fix.

PR #723, opened 2020-04-21
SEA Security Access, Prospector maintenance/E-storage

Sparrowbird611:
Changes the access on the Prospector, SEA, and SGR

Prospector: Given emergency storage and maint access, because I forgot to do this the first time.

SEA and SGR: Given security access. These two are both meant to ensure the officers and the law are being enforced correctly, yet neither can get into the brig. This is rectified.


SouthyOsmanli:
I don’t feel SEA should have security access. They are a senior enlisted advisor to command for navy and marine personnel. They shouldn’t have access for the same reason Bridge Officers don’t - they can advise sec or carry down orders from command but are not their department/chain of command superior, they aren’t legal council nor should act in place of security.

Sol Gov Rep makes more sense to have the access due to his law-like nature and intention (on Bay) - I could see him having the access to act as legal council.

Sparrowbird611:
The SEA's main role is to ensure discipline and good conduct are being enforced on the ship. The main offenders to this will primarily be either security, or the people security throws in the brig, so he should have easy access to deal with this people. If it turns into an issue of SEAcurity then it will probably need some changing, but I don't believe it will turn to that.

SGR I just did this for because my god he's a worthless piece of shit with absolutely no authority or purpose. Even the CL has more reason to exist, so hopefully this will help push them to be ~~blood sucking lawyers~~ defenders of the rights of the people.

Roland410:
I'd also like to remind you that the SEA has full e-arm access, so I don't think having sec access will be an issue.

BurgerLUA:
failing tavis, currently

Sparrowbird611:
Fixed.

Sparrowbird611:
Alright, I'll start then.

The SGR is a legal man, and as such, requires access to the place where the majority of law related tasks are debated and accomplished. He's not much use to the prisoner if security decides "Hey, you're not allowed to be here," and he can't prove anything without access.

The SEA's purpose is to ensure discipline and regulations are enforced, and as such, should have the ability to enter one of the main places where regulation breakers are sent. Also, the dude has E-ARM access, so I can't even imagine why he doesn't have sec access at this point.

Shaft miner maint access was an oversight in my last access change PR. There's absolutely zero reason to deny this part, since I gave literally every other role maint access.

bearrrrrrrr:
We removed SEA's security/medical/engi access specifically because they used it to kinda powergame and valid the antags. Not a big fan of reintroducing it to them.

AdamElTablawy:
Then what's the point of the role, if they can't control security?

Just null their combat skills. Cap it at trained.

PR #734, opened 2020-04-23
makes the Ebow actually do what it's supposed to. (AKA: reverts ebow nerf)

chaoko99:
🆑 chaoko99
tweak: Ebow no longer does toxic damage. It now weakens people when it hits. Ninja ebow still exists, also reflects this now. You can use it by middle clicking a mob with the sword module selected. Betcha didn't know that.
/ 🆑


Sparrowbird611:
No objections, but a few things.

First off, you're going to want to touch the actual e bow ammo count if you're going this route. Current e bow has 8 shots and self recharge. I'd suggest lowering that to 1-2 shots.

Secondly, this probably justifies a price increase on the actual e bow itself, given weaken is a status effect that completely ignores armor as far as I know.

Thirdly, I'd give it back a little toxin damage, even if it's just 10 damage.

Roland410:
Weaken is absolutely an OP damage type.
As Sparrow said above, it shouldn't have more than one or two shots, and a fairly long recharge time along with a price increase.

ProbablyCarl:
Reduce the number of shots to one or two, or find another compromise, and I'll accept it.

ProbablyCarl:
Thank you.

Higgin:
2 seconds weaken on the regular ebow here is super weak and bad if it's only gonna have one shot.

For some perspective, a person with master CQC pushing somebody with Untrained over still at most does weaken(2) - it offers a very limited window for doing anything unless you're immediately following it up with stealing their gun or double-fisting them to death with your off-hand esword.  

I can only suggest at least two shots for the mini e-bow or raising the weaken value a bit.

Sparrowbird611:
While that is a fair point Higgins, it is also still dealing agony damage. The point of the e bow is to immediately disarm a threat, then either finish them off or disarm them and continue to pelt them with bolts. Two shots is a nice bonus, but I don't think weaken really needs to be any higher.

PR #737, opened 2020-04-23
removes mindslave implants

AdamElTablawy:
i'm removing mindslave implants because they're shit

some of you may object to this with a variety of well reasoned arguments

those arguments are also shit and i do not care


Sparrowbird611:
Going to downvote because my argument is shit.

Also can we delete Hestia it's also shit.

EldritchSigma:
I just downvoted you but what does it mean? That means I clicked that thumbs down icon in the reaction selection menu. That means a thumb down appears now right under your post. Other people may click on this icon to express their discontent.

Do you think I downvoted u in error?

U can PM me but I rarely change my mind on downvotes.

How to not get downvoted?

Dunno.

BurgerLUA:
Okay so there was a poll on this not too long ago. It was voted to kept in.

Unless you can find some stellar arguments or do some compromise, this PR will be closed.

AdamElTablawy:
@BurgerLUA According to Asystole, that poll only involved deleting it from the code entirely, not removing it from the uplink, so this PR is just removing it from the uplink and means that it's unrelated. You can go ask them what's up with that, if you want.

Otherwise, I genuinely just do not care, at all, about the poll. If the playerbase wants to be wrong, then they can be wrong and disappointed. They're guidelines to determine the playerbase opinion, not binding.

The fact of the matter is that the admin team hates these fucking things about as much as I hate everything, ever. According to Asystole, people just refuse to handle the mindslave tickets outright. You jab these things in someone for 10 tc and then they're antag, if they want to be or not. 

Wasn't the entire point of making it so that cult couldn't force convert was that forcing people to be antag if they didn't want to is bad? Well, now you have a 10 tc implant so you can make a 12-person fortnite dancing flash mob that has no choice but to follow your orders even if they don't want to be antag, and admins, like gabo, have said they hate dealing with mindslave tickets because of exactly this.

They cause issues with metashielding-either they're undetectable or they're the first thing you check for. They rope someone into being an antag, even if they don't want to be. They genuinely annoy the admin staff to the extent of literally just not wanting to deal with them, at all.

Why should they remain? What possible argument is there when there's literally just admins complaining about it at this point, and the poll itself was a narrow thing?

That's why I made this PR, and that's why I don't give any fucks about arguments to the contrary. I'm not compromising and i'm not going to compromise. If you're up in arms about the poll thing, go take it up with the head maintainer-if an antag item is so fucking bad it's making admins not want to do their jobs, clearly there's an issue.

And before you say 'but, we could rework it'-we won't. You know we won't. Everyone knows it'll never get reworked. It'll remain, like a tumor, in the uplink, pissing everyone off while everyone discusses rework after rework until the sun explodes and consumes the fucking earth.

BurgerLUA:
I'll make a poll on the matter. For balance issues, the opinions of players matter.

BurgerLUA:
I probably shouldn't have even bothered to make a second poll considering the previous one was only 2 weeks ago and this entire PR goes against what is acceptable for a PR.

AdamElTablawy:
looks like it's winning anyway

ProbablyCarl:
those arguments are also shit and i do not care

SilvernKell:
**Disgusting behaviour.**

Or should I say **shitty** behaviour.

PR #739, opened 2020-04-24
Reverts the Neon Vox 

CrackParrot11:
This reverts the neon vox on the character generation screen, The default vox sprite was changed to be completely blank to allow for any high Opacity color to be thrown onto it without giving it a restriction on the types of colors it could use allowing for really bad and ugly neon color channels. I am going to give the Vox multiple colors in the future but they need to be resprited first in order to not have absolutely retarded color gradients.


Roland410:
Absolutely based.

MemedHams:
vox colors are literally broken right now, this is less of a revert and more of a bugfix imo

PR #744, opened 2020-04-24
FBP/IPC microbattery health boost.

FGRSentinel:
Boost to microbattery health (30 to 45) to give FBPs and IPCs a slightly better chance to escape fights and (very) slightly reduce the urgency of finding a charging station (about 2 extra minutes of battery life at 5 damage) if battery damaged.


Sparrowbird611:
Seems good to me. Most guns will one shot you any way still though.

FGRSentinel:
Mostly this is meant to just give a bit more wiggle room for unarmored FBPs and IPCs to clear an area when someone starts shooting by slightly increasing how much damage they can take before the battery stops working. The way the code works this should also result in a bit more forgiving a power loss from minor damage as well, which is a nice bonus but doesn't change much outside finding someone to fix them up.

bearrrrrrrr:
i already chimed in and I *do not think* this is a good idea. As a synthetic, I can already survive a 3000c hydrogen/plasma fire with basically 0 consequences for my actions. It is literally the tradeoff. You have a disadvantage against certain weapon types in direct combat, and die to pressure while unarmored faster.

Like tbh it's kind of extremely disengenious to even imply that you have it harder than organics in this instance. like if i shoot an unarmored person in the head they go down *before* you will because of brain damage. If i shoot them in the chest their heart and lungs will take damage and they will die.

I genuinely feel like the only reason people think synths are underpowered on hestia is because they're op as shit everywhere else xd

code/modules/organs/internal/species/fbp.dm
```diff
@@ -12,7 +12,7 @@
 	var/servo_cost = 0.8
 
 	min_broken_damage = 5
-	max_damage = 30
+	max_damage = 45
```
bearrrrrrrr, line 15: max_damage = 40

FGRSentinel:
> i already chimed in and I _do not think_ this is a good idea. As a synthetic, I can already survive a 3000c hydrogen/plasma fire with basically 0 consequences for my actions. It is literally the tradeoff. You have a disadvantage against certain weapon types in direct combat, and die to pressure* while unarmored faster.

As I stated, this is _not_ meant to increase general FBP durability. It's not something that makes FBPs able to survive environmental damage better or make them able to tank entire mags. This is literally a 15 point max damage increase to bring them on par with organic hearts (which still have a 0.7 percent damage reduction that means they take 30% less damage from the same sources) to increase the chances that an unarmed, unarmored FBP/IPC can escape a gunfight they found themselves mixed up in and give them a sliver of time to find someone to fix them up before they completely depower.

To give an idea, at zero damage the 3000w battery an FBP has lasts about 62.5 minutes at full charge. Currently, taking a mere five damage drops this to about 19 and a half minutes following a formula based off what's in the code. At 10 damage this drops to a little over nine and a half and by 25 damage (the lower damage threshold where I think lungs can start having issues) the battery has 1-2 minutes of charge left. This 15 extra max damage gives about two extra minutes at 5 damage and means precious extra seconds as the damage increases, which is something FBPs and IPCs might desperately need to at least find a charging station.

SilvernKell:
Hi, this is mostly in a response to Bearr. 
The comment has some good points and IPC and FBP in general feels in a better place then it's been before, and not needing a lot of tweaks.

But I support this because of the fact that, say if an organic was shot and it damaged their heart and they die, the difference is they still stay awake for a little bit longer, and medical has better to tools to find and aid them before dying. An ipc of fbp just flat out passes out on the floor and a black screen isn't fun for anybody!

A better way of fixing this then buffs though would be doing something bay has now. On sensors you can see an fbp/ipc's battery charge %. This change doesn't really help in actual fights either, which is one of the places where just getting one shot in the chest and immediately passing out forever is the big bummer. Atleast as an organic you get to fight for survival.

Finding a different way to ensure you're not passed out forever rather then a buffing might be a better way to go.

The issue it's trying to fix is a good one, but the manner of doing so might also be better served by a different plan of attack, by simply utilizing a better way for people to come and assist none-organics in distress, just as organics rely on others to save them.

bearrrrrrrr:
@FGRSentinel you keep bringing up the charge time like it's a big thing when it honestly isn't. from my experience, i have never had to run to a recharger every 10 minutes, like...ever

even if i did have to go back to a recharger, it's not like they're *super rare*. someone who got shot in the chest requires instant medical attention, as an ipc/fbp i can literally go 'lol ok 'and walk it off

and *yes* i am aware you're not pushing this change to buff their enviromental survival, but if you push this change to do so, you are not only un-nerfing them, you are also giving them those previous advantages

literally nobody likes fighting non-human species due to their inherent 'rulebreaking'/unfairness, and it is part of the reason why prommies had such a big pushback on their buff and were eventually just flat out disabled (iirc). you're not making ipc any more fun to play, you're making it more frustrating to fight against

i've said this a few times, i genuinely firmly believe the only reason people think IPC/FBPs are underpowered is because they're so absurdly OP on other serverws lmao

AdamElTablawy:
no people think they're shit here because they're shit and everyone knows it

they're crazy vulnerable, get shot once or twice in the chest and they drop immediately

bearrrrrrrr:
bruh when unarmored sure. wear armor so you don't die. it's literally a tradeoff for being functionally immune to heat/cold/pressure/rads/hallucinogenics/gasses

i will repeat, nobody likes fighting against nonhumans

SilvernKell:
who is this metaphorical nobody
why are you acting like we're an action server
real enlightening and big brain posts bearrr

bearrrrrrrr:
We aren't an action server, which is why I am against having people be super buff in combat.

Yall don't remember when IPC's could take 7 magazines of SMG ammo but I do lmfao

bearrrrrrrr:
image

this person was still alive btw

FGRSentinel:
You're, again, overlooking a few details. The heart has a max damage value of 45 and a damage modifier of 0.7 that means it has an effective max damage of 64-65 (due to taking 30% less damage). Lungs have a max damage of 70 and even after they're popped you don't drop immediately. 15 extra points of damage is maybe one extra shot to the chest to put them down, which isn't even enough to make such a big deal about.

You can't even argue that FBPs and IPCs are easier to repair considering anyone in medical can pop an organic into a sleeper to mix tricordazine for their external injuries and there's almost always 1-4 (up to six depending on what skills the Roboticists picked) people who can fix broken bones or damaged organs, while there's usually 0-2 people who can fix an FBP/IPC since medical lacks the tools/OOC knowledge to do so easily and there isn't a guarantee of a Roboticist due to the minimal amount of work they get in the average round, which means they often have to wait so long that organic crewmembers would be dead and they can't ghost without giving up the ability to come back into the round as that character.

At present, the FBP/IPC microbattery has a max health of 30 and the amount of damage it's received (so long as it's five or more) impacts both max capacity and power drain. There's even a formula you can derive from the code to find how long the battery should last, in seconds.:

(cell.charge*(1 - damage/max_damage))/(servo_cost*(1 + 10 * damage/max_damage))

servo_cost is 0.8 and the code itself says that this results in a battery life of about an hour (my own math says 62 minutes and 30 seconds) of constant activity with zero damage and full charge (cell.charge of 3000), which drops to 19 minutes and 32 seconds at 5 damage, well over a 66% efficiency loss from realistically trivial damage, with the current max_damage value of 30. If it's adjusted up to 45, this becomes about 26 minutes and 20 seconds. That's more bearable, but I don't believe a single punch to the chest drops the efficiency of any organic organs by 50-66%.

JohnBigless:
We did it, boys. We have managed to buff IPC.

PR #751, opened 2020-04-25
Vox Map QoL

ProbablyCarl:
- - -
Map:
 - Vox ship now has the actual access requirement as it should've. This is to stop prisoners from escaping immediately, grabbing a rifle from the armory and hijacking the ship in under a minute.
 - Vox vessel rear given a touch up, primarily with the burn chamber, as well as tossing the Vox a nuclear reactor if they'd like to kill themselves indirectly. These changes are seen in the mapdiff.
 - Various other changes. Primarily adding walls.


Roland410:
Caaaaaaarl, you need to disable autoset access on doors if you set access on it, you've made Travis runtime on the voxship away site.

PR #759, opened 2020-04-26
Fixes Voxship runtime because doors

Roland410:
Fixes the Voxship runtime because of a door was set to autoset access while having access requirement set on it.


Roland410:
@ProbablyCarl Can speedmerge this as well btw, passed on my Travis already.

PR #766, opened 2020-04-26
Revert "Removes brain damage from awakening itself"

ProbablyCarl:
Reverts BoHBranch/BoH-Bay#758


ProbablyCarl:
Temporary removal, as discussed with the creator and Ikky.

PR #770, opened 2020-04-27
Revert "Revert "Makes all silicons have zero skills by default.""

Roland410:
Reverts BoHBranch/BoH-Bay#764
This was reverted for absolutely no fucking reason.

### DIFF
--- a/code/modules/mob/skills/skillset_silicon.dm
+++ b/code/modules/mob/skills/skillset_silicon.dm
@@ -1,8 +1,6 @@
 /datum/skillset/silicon
-	skills_transferable = FALSE
-
-/datum/skillset/silicon/robot
 	default_value = SKILL_MIN
+	skills_transferable = FALSE
 
 // better handling for hard resets
 /mob/living/silicon/robot/reset_skillset()


SilvernKell:
No nooo
You need surgery borgs to have surgery skills.
Don't engi borgs need construction?
I mean the only issues would be piloting (a gameplay choice whether to have it or not) and exosuit training/. . actually, I cannot think of another issue right now

Roland410:
Look at the code before you complain please, thank you. Borgs get their skills from their modules. This literally has no effect on anything else than the AI.

SilvernKell:
Who was complaining
How does it affect AI then besides flying the shuttles?

Roland410:
Sorry about that, I'm just tired in general and tired of everyone bitching about every change.
To clarify, this means the AI can't use NTnet monitoring, so hackerman need to be stopped by actual people such as the CE, SE etc. and that they can't fly ships nor use the sensors to give a scan of PoI.

SilvernKell:
I know the feeling. Thanks for clarifying.

ProbablyCarl:
Does this also have your patented code?

PR #772, opened 2020-04-27
No stale gimmick but I need a better writer

EldritchSigma:
I dislike rev.

I really do.

I cannot think of a decent flavortext that doesnt sound bland as fuck.

But the "Hibernation gimmick" gets stale and revs and loyalists should make their own gimmick.

CL:
Changes the stale gimmick of revs and loyalist being about hibernation.


EldritchSigma:
When you post a PR and forget to change the title. Its about the stupid rev flavortext. Shit gets stale because its literally the same gimmick every time.

Sparrowbird611:
May be outside the scope of this PR, but loyalists need a rewrite. Their flavor text makes it seem like they're supposed to be stopping any revolution, not being the cause for it.

EldritchSigma:
Yeah loyalist should be the reason to cause a revolution tbf

PR #773, opened 2020-04-27
Lobby Art 

CHEtheGREAT:
after the testmerges got unmerged, I closed my PR because I accidentally committed some unrelated things to it, but now i've figured out how to properly use github! 

this simply changes the current lobby art back to the vector based one that we had for about two weeks.


ProbablyCarl:
Travis is a goy and I don't care.
I'm merging this, as we had it prior.

PR #776, opened 2020-04-28
No you are not fucking Yuri while having edgelord Rahn RA2 MO combat skills

EldritchSigma:
Psionic Advisor doesnt get combat skills that are on par with sec. Its max to be trained now.

Yuri is master.


Sparrowbird611:
I was about to make this PR myself, but you should be setting it to experienced. There's no reason for him to be lower than your average crewie.

EldritchSigma:
Changed

PR #791, opened 2020-05-02
The Legal PR + Minor fixes to locker access.

Sparrowbird611:
Renames the SGR to the Adjudicator.

Adds the Bailiff, a body guard for the Chief Adjudicator.

Adds a locker for the body guard in the SGR's old office.

Gives the APA locker CL access instead of security access.

Adds an augmented tie to the SGR and CL lockers.

The SGR is a meme in his current state. It's not really obvious what his duties are, and since we use ship regulations and not sol gov law, he's not even legitimately empowered to do anything. This should rectify that. The body guard is because if the CL gets to have a personal meme man, the SGR does too.

The augmented ties are because they're funny items, so I thought they'd be good additions. The locker changes mean the CL can open his body guard's locker. Because it's kind of backwards that the CL can't open his guard's locker, but the guard can open the CL locker.

I'm new to mapping, so some one please check that, as I'm not sure I've done it right.


Sparrowbird611:
Ah frick… I screwed up map merge.

N8-Toe:
really against this, I enjoy SGR as well as I feel we should work to improve the role before removing. as well as it makes little sense to have a judge onboard, as well as it is a job for such NICHE situations. I think this is a bad change of direction

Sparrowbird611:
Wait what, map diff says it's working?

Huh. Neat.

chaoko99:
Change magistrate to Prefect. It has a more.. Legal implication.

Sparrowbird611:
Alright, so, here's the thing.

SGR is on par with chaplain in terms of uselessness, and, as far as I can see, has no interesting gimmicks to pull off. At very least, a Chief Magistrate is more credible than "politician man."

I don't have anything against the SGR, but he's beyond terrible from what I can see, and I was told that renaming him would get him on to the regulations.

SGR can be re added, and I won't say anything against it, but it serves no purpose in the current state of the game, and I find this to be a better fix for it than anything else.

Sparrowbird611:
> Change magistrate to Prefect. It has a more.. Legal implication.

I'll see about making it an alt title.

N8-Toe:
SGR acts as the diplomat, and represents one of the two major factions. they have as much "practical" use as the CL, but their use is mostly in RP. I think adding the magistrate is a poor move as the XO and CO already exist as a quality control over security. Adding in another role to muddy the water will only lead to more confusion, as well as removing a unique RP based role.

Sparrowbird611:
Alright, but here's the thing.

A diplomat for what? The only beings we interact with are the ascent and the vox, both of whom the captain interacts with because they never have time to come aboard. I haven't ever seen an SGR interact with them (Aside from that one guy who tried to sell a gyne slaves).

Here's the thing about CO and XO in my experience. They're almost always busy, especially since only one or the other usually exists at any given time from what I've seen. They're also often incredibly biased in court cases. The purpose of the Chief Magistrate is to be an unbiased court of appeal. Enforcing security is a secondary concern, but even then, he has the time to dedicate to it, unlike the XO and CO.

Sparrowbird611:
Ah, right. There's so mapping mistakes I need to fix, I'll get to that.

Sparrowbird611:
Just did away with most of the map change. Some one wants to use that room, and I like what they have more than what I was doing. Judicial locker just replaces the book case now.

chaoko99:
I did some reading, and apparently both Magistrate AND Prefect are wrong. The correct word, used for explicitly this situation, is "Adjudicator"

Sparrowbird611:
I guess I can add Chief Adjudicator as well, since that one sounds really based.

On another note, I forgot to remove the SGR race restriction on this one, so I'll do it now.

Sparrowbird611:
Alright, I just want to throw out. I'm not against SGR as a role. I like the idea of SGR. However, his current implementation and place in the lore means he's pretty worthless ingame. If some one can find a good way to make him a role with an actual purpose, I'd be all for re adding him, but until then, I'm standing by this PR.

And please don't say "Give the SGR the legal power." It's been talked about, the law/ship reg maintainer doesn't want it that way, and I'm going to follow whatever he says on this part of the matter.

BurgerLUA:
Wish this PR was a little more atomized but whatever.

BurgerLUA:
Seems like this is WIP. Tell me when it's no longer WIP.

Sparrowbird611:
Ok I swear last commit then I'm done.

Sparrowbird611:
> I mean I guess it's an okay PR? Wish this was more atomized though.

Could you give me a little more info on what this means, for next time?

BurgerLUA:
> > I mean I guess it's an okay PR? Wish this was more atomized though.
> 
> Could you give me a little more info on what this means, for next time?

Fixes and features should be separate from eachother so they can be merged faster and easier.

chaoko99:
Conflicts. Also `squash`

Sparrowbird611:
Closing, an updated version has been made.

PR #792, opened 2020-05-02
Fixes air alarm sensor settings for server and sauna

Allfd:
The Sauna is set to a default temperature of 75C however the air alarm was set to enter a warning state at 40C and an alarm state at 66C.  This changes that to 76C and 78C respectively.

The server room was supposed to have a lower warning and alarm threshold then other air alarms, however this was overwritten by the Initialize() proc, meaning that while it was set in code, it was overwritten when the Air Alarm started at the beginning of the round with its parent values.  This was fixed by moving those settings from the New() proc to its own Initialize() proc.


Allfd:
It passed the local unit tests but not when they were run by travis.  I will look into why that could be.

code/game/machinery/alarm.dm
```diff
@@ -120,11 +125,14 @@
 	report_danger_level = 0
 	breach_detection = 0
 
-/obj/machinery/alarm/server/New()
+/obj/machinery/alarm/server
+	target_temperature = T0C+10
+
+/obj/machinery/alarm/server/Initialize()
```
BurgerLUA, line 131: This has a bad Initialize call, for some reason.

code/game/machinery/alarm.dm
```diff
@@ -101,17 +101,22 @@
...
 
+/obj/machinery/alarm/cold
+	target_temperature = T0C+4
+
 /obj/machinery/alarm/warm
-	target_temperature = T0C+75
 	environment_type = /decl/environment_data/finnish
+	target_temperature = T0C+75
+
+/obj/machinery/alarm/warm/Initialize()
```
BurgerLUA, line 116: This has a bad initialize call for some reason. Try replacing ``..()`` with ``. = ..()``

Allfd:
Your recommendation fixed it, Thank you!

It is now failing with 

```
## UNIT_TEST ##: /datum/gear/uniform/suit - /obj/item/clothing/under/rank/internalaffairs/plain ('formal outfit'): Did not find a gear_tweak's icon_state 'internalaffairs' in the icon 'icons/obj/clothing/obj_under.dmi'.

## UNIT_TEST ##: !!! FAILURE !!! [LOADOUT: Entries shall have valid icon states]: One or more /datum/gear definitions had paths with invalid icon states.
```

I am at a bit of a loss as nothing in my changes touch that.

PR #803, opened 2020-05-06
Infantry Rework.

ProbablyCarl:
- - -
Items:
 - SL given the option of having a lighter kit, reintroducing the older C20B as an option. As opposed to just the Tectonic, which had replaced it at an earlier date.
 - SL has their own 'unique' variant of the Laz-Carbine.
- - -
Balance:
 - Infantry rigs given slightly better laser protection, to better represent the role of the suit. It still retains fairly weak ballistic protection, however.
 - Infantry re-added. However, they're unable to spawn without an SL present.
 - Tectonic removed from Sec's armory, as it's once again an item in the SL's possession, as prior.
 - Maint loot given a minor boost via more spawns. This, however, also increases the number of critters that spawn in maint.
 - In line with the above, two new mobs have been added to the hostile mob pool. Vox Slugs and Leeches.
 - Auto auth off ship now includes energy weapons. This is an accidental balance fix, but something that needed to be done regardless. You can see why in the bugfix section.
- - -
Map:
 - Old Aux Medical moved to old Infantry Prep area.
 - New Infantry Prep/Hangout north of the Garden on deck five.
 - All common away sites given a facelift, introducing various hostile mobs. Primarily, Syndicate agents.
 - Adds a new map planet type called "Wartorn". This is quite literally just the 'garbage' planet, but with Russians and Syndicate agents. 
 - The above comes with two 'new' ruins. One for a crashed Syndicate dropship, and another, which is just the Lodge with two Russians and a jukebox. As is tradition.
- - -
Bugfix:
 - Laser carbine issued to infantry should no longer not fire without pre-auth off the ship. This was due to the way I had it, only checking projectile weapons. This shouldn't be an issue any longer.
 - RIG QoL material applied to Infantry rigs.
 - SL should now have the correct access. Unless they don't, in which case I'm too tired to notice.
- - -
Misc:
 - Spawn message for Infantry clarified.

Probably some stuff I missed.


Sparrowbird611:
I know this is probably an unpopular opinion but, I stand by the statement that infantry is a bad role.

They detract from security by being guys with bigger guns, cooler gear, and better names.

They detract from exploration by doing similar tasks but with more equipment.

In general they're just better versions of both factions. I'd honestly rather see better armed exploration teams than infantry back, because as it stands infantry is incredibly redundant for exploration, and just "sec man but better" for security.

BurgerLUA:
Failing Travis

ProbablyCarl:
Travis is having a stroke. This all works locally, and we've tested it live.

PR #729, opened 2020-04-23
X-Ray Love

Higgin:
Gives the x-ray carbine and energy sniper a bit more to look forward to in the morning with additional AP.

PR #805, opened 2020-05-06
ThatgunPR: Various gun rebalances. Feedback needed.

Sparrowbird611:
This PR rebalances several guns. I'll list em out here. List out any guns you think I may need to change.

Laser carbine: Reduced damage to 30, (from 45). It's still a very strong weapon, but it's less likely to one shot pop your organs.

Marksman laser: Removed the weaken and stun effects from it, but doubled it's AP (It now has the same AP as the las carbine did previously, sitting as a solid 20 AP). Additionally increases accuracy, because for some reason this thing gave you storm trooper aim.

Sonic cannon (vox weapon): Decreased agony on the low mode to 50 (from 70) but increased the main mode to deal 80.

Stun revolver: Agony decreased to 35 (from 50).

Stun rifle: Agony decreased to 50 (From 80)

Additionally, feedback is necessary on another gun I'm considering touching, the infamous AMR. Considering removing it's weaken and stun, but increasing it's damage. The changes will mean that unarmored people be at an even higher risk of being injured, but armored will have a chance to remain standing if they wear specialist armor. Numbers are undecided, so if you agree with this, shoot some numbers at me.

Additionally considering touching laser cannon stats, inform me on thoughts of that.


code/modules/projectiles/projectile/beams.dm
```diff
@@ -29,9 +29,9 @@
 	damage = 25
 
 /obj/item/projectile/beam/midlaser
-	damage = 45
-	armor_penetration = 20
-	distance_falloff = 1
+	damage = 40
+	armor_penetration = 10
+	distance_falloff = 2
```
comma, line 34: fyi it's not damage falloff it's how much accuracy degrades per tile travelled

Sparrowbird611: Wait what? Are you sure?

Sparrowbird611: Ok nevermind you're right. Uhhh… not sure if I need to change that then, I'll look at the code.

AdamElTablawy:
Do not change AP on the carbine.

Just lower its damage.

Sparrowbird611:
Went with that idea. Description updated to reflect that.

Higgin:
Leave the LWAP and electrode stun weapons alone.

The stun/weaken on the LWAP are symmetrical w/ the AMR and not a big deal.

The higher agony values on electrode weapons balance them against the many other objective upsides of beam stuns: accuracy, speed, hitscan, window pen, and ammo capacity. They're supposed to hit harder.

Laser carbine nerfs/buffs I've got nothing to say about - I have some plans for organ damage besides. For reference, the small handheld laser has 25 damage, no AP - and in #729 the X-ray weapons all have around 30 damage with pretty massive AP as well.

If you want to differentiate them, I'd leave las carbine between 35-40 with 15/10 AP. That ought to also leave them useful in most scenarios.

chaoko99:
Taser revolver is not hitscan, and is thus much stronger as a result. This is intentional, it's a harder weapon to use for more gain. It doesn't need to be nerfed.

Sparrowbird611:
Tbf It does hit harder, but I see your point. Perhaps raising the revolver to 40 agony, and the rifle to 65 would suffice?

Higgin:
@Sparrowbird611 I did some more local testing and honestly I'd like to revise - I'll take a damage nerf on the carbine. 

Just leave the stun weapons and LWAP alone.

Sparrowbird611:
LWAP I'm pretty much set on, since I can't even say it's working properly with how much it misses for a sniper rifle. Stun weapons I'll change the numbers on more though. 40 for the revolver and 65 for the rifle seem good, and I'll raise the sonic cannon as well.

BurgerLUA:
Needs a poll.

Sparrowbird611:
I'm frankly looking over this again and I don't actually know what I was thinking on quite a bit of this PR when I put it up. I'll let the vote go but eh, I realize now that quite a few of these are bad changes. The only ones I can really stand by are the two laser weapons and possibly the vox cannon. Nerfing the non lethals probably wasn't the best of ideas rn when all our lethals are this strong.

ProbablyCarl:
Votes been up for a bit, and it's not looking good. I'll be closing this.
DM me if you're wanting to talk about it further - as we've discussed the subject prior to now.

PR #808, opened 2020-05-07
Add a sound for turning on and turning off welding tools

JohnBigless:
Adds two sound files for turning on and off a welding tool. Also adds some shitcode to have theese sounds play when a welding tool is turned on and off.

*Look mom, i can code!*


JohnBigless:
Whoever called the welder sounds 'ass plumbing' is based.

PR #810, opened 2020-05-07
Allows surgery borgs to put organs in again.

Higgin:
get_active_hand was causing troubles with borgs, shifted it to fix it.


code/modules/surgery/surgery.dm
```diff
@@ -214,7 +214,10 @@ GLOBAL_LIST_INIT(surgery_tool_exception_cache, new)
 			return TRUE
 
 	// Otherwise we can make a start on surgery!
-	else if(istype(M) && !QDELETED(M) && user.a_intent != I_HURT && user.get_active_hand() == src)
+	else if(istype(M) && !QDELETED(M) && user.a_intent != I_HURT)
+		if(istype(user, /mob/living/carbon))
+			if(user.get_active_hand() != src)
```
chaoko99, line 219: this could be condensed into one if statement.

chaoko99: The last two, I mean.

chaoko99: Reasoning: Short circuit eval means that if the first check fails, the second will never run. So therefore it would be fine.

AshtonFox:
This fixes the impossibility of transplantation as such, but there is a bug that causes the transplanted organ to not be properly removed from the gripper at the end of surgery. Subsequently, the same organ can be transplanted many times, even to other mobs than the one to whom it was initially transplanted (until eventually Drop Gripped Item verb will used). I saw the same bug in [an older implementation](https://github.com/Baystation12/Baystation12/pull/27605) that was rejected by the Baystation12's repository' moderators.
The problem is that it is extremely dangerous if an organ is transplanted to more than one mob, since the organ will have the same reference, and if one of the organs is damaged, same damage will be caused for any and all others in the chain, so don't be surprised if you see people die under mysterious circumstances after visiting a cyborg surgeon.

adamkad1:
Well, for now, i guess you could stash the organ in a middle of nowhere in a fridge... Wait, if you destroyed the organ while its outside of a mob, will it kill mobs with it?

adamkad1:
Also it doesnt work apparently?

AshtonFox:
I didn't every try destroying same referenced orgrans, so maybe you'll check this.
And the fix seems to be working one.

adamkad1:
> I didn't ever try destroying same referenced orgrans, so maybe you'll check this.
> And the fix seems to be working one.

Last round i played, i couldnt insert things with medborg. also i could insert robot limbs with sciborg. but thats a separate borg right? maybe you could fix that too

adamkad1:
Also, maybe the author could push that to original bay too

PR #817, opened 2020-05-10
Medal Vending

ProbablyCarl:
Adds a simple medal vendor within the Captain's Cabin. Includes all medals, including three 'exclusive' sets through a hacked interface.


BurgerLUA:
Failing travis

PR #825, opened 2020-05-11
Changes fall damage armor piercing to enable planetary drops.

Rav-Nunya:
You can fall from D1 to D5 in a sec hard suit but the same naked will still kill you, also due to me being a dumbass only two away sites can be modified to enable planetary drops.


BurgerLUA:
the hell happened with the comparisons here?

PR #829, opened 2020-05-12
Removes Telepathy from genemods

DriedCrust:
I've only seen telepathy being used for LRP purposes but there are plenty of arguments for and against it. This might get heated so I expected a vote from this.

### DIFF
--- a/modular_mithra/code/modules/mithra/traits_positive.dm
+++ b/modular_mithra/code/modules/mithra/traits_positive.dm
@@ -91,14 +91,6 @@
 // BoH Materials
 /////////////////////
 
-/datum/trait/commune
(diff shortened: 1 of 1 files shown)


chaoko99:
Here's a better idea! Give it an admin log, and we can PUNISH the LRP people. We have this cool admin team and we refuse to use it apparently.

chaoko99:
Also please provide examples of LRP behavior instead of going "Oh, but it's so LRP! Guess there's nothing we can do!"

DriedCrust:
nice thats a better fix

chaoko99:
Sorry for being so hostile, this isn't the kind of thing I was expecting to awake to.

PR #837, opened 2020-05-12
Fixes psychic whisper's "LRP" behavior features.

chaoko99:
You can no longer communicate psychically while unconscious or DEAD. Also added a psi implant check for those people who are abusing it.


chaoko99:
alternative to #829 

I really didn't expect to have to wake up to something so willfully irritating.

PR #838, opened 2020-05-12
Adresses a dumb lore concern in telepathy.

chaoko99:
Adjusts the Telepathy lore blurb.

Let's stop pussyfooting around it so people stop asking dumb questions in the lore channel.


chaoko99:
THERE, I've addressed literally everything wrong with telepathy.

PR #840, opened 2020-05-12
Fixes APA locker access.

Sparrowbird611:
Gives the APA's locker CL access instead of security access. Because the guard opening the CL's locker but having his own private locker is a bad meme.


alexpkeaton:
I do not understand how the APA can access the CL's locker but not vice versa.  Could this have something to do with it?

Sparrowbird611:
> I do not understand how the APA can access the CL's locker but not vice versa.  Could this have something to do with it?

In the olden days, the CL has security access. However, this was removed, since it felt odd for the CL to have said access. However, as a leftover of that time, the APA's locker used security access, despite the fact that the role isn't really supposed to have it. The APA however still retained his CL access, which the CL's locker also used. This basically ended with the APA locker only being usable by security personnel and the APA, while the CL's locker can only be opened by the CL or APA. It seemed backwards to me, so I've fixed it.

PR #850, opened 2020-05-13
Gives forensics EVA access

DriedCrust:
Literally everyone in security (including cadets) have access to EVA, this has been something people have complained for quite a while, this adresses this.


ProbablyCarl:
It wasn't intentional, no. As far as I'm aware. Another case of oversight, me thinks.

Or you could argue that, given Forensics is intended to be a slower, more RP oriented role, it didn't have it due to such. But it's still pretty stupid.

PR #854, opened 2020-05-13
Rank tweaks for infantry and misc. jobs, brig chief alternate title

Jaraci:
- Squad lead locked to E-6 staff sergeant. E-7 and above don't really make much sense.
- Combat technician can be E-3 now.
- Brig chiefs can (finally!) be Wardens if they really want.
- Service jobs like chef, cargo technician, janitor etc. can now be E-5s. Billet takes precedence over rank in almost every case anyway, so I really don't foresee any CoC issues coming out of this.


Roland410:
Warden alt-title is unnecessary and pointless, as we're not running a prison on the ship.
The rest of the PR is pretty bad as well, none of those ranks make sense.

Flying-loulou:
>     * Squad lead locked to E-6 staff sergeant. E-7 and above don't really make much sense.

Yes please.

Jaraci:
I didn't say "prison warden", I said "warden" -- as in, a guard/watchman and/or someone who is in charge of looking after something, namely the prisoners and the armoury.

As usual, your feedback boils down to "it's bad" without actually giving any reasoning why. I didn't pull these changes out of my ass -- E-7 and above for a squad leader is ridiculous (arguably it could be as low as E-4, and should maybe have E-5 as-is), specialists can easily post up at E-3, and so forth.

Roland410:
Technically, the SL thing makes sense, you're right about that, but why would a crewman be E-4-5?
Why is the chef E-5? Janitor as E-5? Cargo tech is borderline, but even then, why would it be an E-5?

None of these make any sense realistically and are unnecessary for gameplay purposes. This is just rank wanking.

BLUNTFORCE420:
heres a hungarian furry pretending he knows anything about militaries

Roland410:
> heres a hungarian furry pretending he knows anything about militaries

Ah yes, very cool, I hope you made an account 3 hours ago just for this. The best part is that I am not even a furry, and most people know that I only tolerate them.

BurgerLUA:
Needs additional feedback. Find more people to give you more updoots on this PR and I will merge it.

OolongCow:
Please, God, merge this. There is zero reason whatsoever for people who should be in their late 40s to mid 50s to be babysitting four junior enlisted. E-7s and above are only a relatively short time away from becoming SEAs and managing hundreds of people and their problems, they shouldn't be running around shooting things.

PR #857, opened 2020-05-14
Gun Printer nerf + Autolathe buff

Sparrowbird611:
Gun Printer part: Removes a large number of the guns from the security gun printer, along with a few nonsensical items, like rapid syringe guns, chem grenade casings, and floral somatrays. Generally, if it wasn't already in the sec arm, it can't be printed any more. 

Autolathe buff: The Floral Somatray is now available in the autolathe, and the chem sprayer, rapid syringe gun, and large chem grenade can be made in a hacked autolathe. Hopefully this means that these items finally will get the use they deserve.

Main reason for this was because ~~Screw Tracker X Trune and their combo of science and armory powergaming~~ the printer isn't supposed to be giving security these incredible end game weapons, it's there to replace any guns that may get lost or stolen.


Roland410:
Leave the 10mm haywires at least, they are barely used.

Sparrowbird611:
Eh, not a fan. Sidearms with ions are a pretty hard counter to mechanical threats, and fantastic support tools. Unlike the ion gun, they're easy to swap out with lethals. Haywire shells require a shotgun, which is bulkier.

EldritchSigma:
So the guns that are printable in there are now barely obtainable. Yeah not gonna lie, not exactly what baycode needs. More content removal

Roland410:
Sigma also makes a fair point. Although I've been on the opinion that the printer's stuff should be going back to research for quite a while.

Sparrowbird611:
The unobtainable guns are good points, and I've added it to my list to think of ways to acquire the decloner and the phoron pistol (Easiest would be to just let antags buy them, but I'm not a fan of that solution). However, the X-ray laser and proto SMG I have no issue with being rarely seen, given they're incredible armor penetrating weapons, though I wouldn't be adverse to giving the Verne a proto SMG.

I originally wanted the gun printer to be in science, but I'm pretty sure that's off the table with carl, so this is what I'm going with.

Roland410:
Actual wild idea I just had, and not sure why I never thought about it before.
Place all guns and such back to science, BUT allow security to print replacement guns and stuff like las carbines by default as-is currently.

AdamElTablawy:
Removal of bad content is still good, even on content-deficit codebases like Bay.

If you give the guns to science, sec will just take 'em anyway.

ProbablyCarl:
We'll be test merging this.

Sparrowbird611:
Something break? Not sure why my attention is needed.

Roland410:
The Travis issue doesn't seem to be related to this PR. I'd rerun Travis and write it up for one of "those" bugs.

PR #863, opened 2020-05-15
Min. age tweaks for various jobs

Jaraci:
- Combat technician min age from 24 to 22. It could potentially be pushed a tiny bit lower too, probably, but whatever.
- Engineer min age from 19 to 24. Ideally, the job would be split between full ship engineers and those who are more like just mechanics, but whatever.
- Deck officer from 27 to 25. The previous age reflects the fact that they were previously locked to high-ranking senior enlisted only; now, they can be relatively junior officers ever since the renaming of the job.


Roland410:
Technically the DC is not enlisted now, but a Warrant Officer, sooo it kinda made sense.

PR #871, opened 2020-05-16
Rig fixes and QoL

Roland410:
🆑Roland410
tweak:IPCs can wear RIGs properly now.
tweak:Added the bag QoL to the rest of the RIGs.
/🆑


ProbablyCarl:
Some of the RIGs were left without the bag option intentionally at first. A few I probably just straight up forgot, though.
Obviously, I should've given it to all in hindsight, but, it is what it is.

Anyhow, this is good. Assuming it passes Travis.

PR #874, opened 2020-05-16
Garuda QoL

Roland410:
🆑Roland410
maptweak:Gave the Garuda a CO2 tank (big one), removed the air canister stuff, as it's useless anyways.
maptweak:Gave the Garuda a proper (not weak) sensor.
/🆑
As per description, picture below:
image


ProbablyCarl:
Make sure the tank is incredibly high cap, or there's a method of refill possible.
It'll have to last a _long_ time.

PR #876, opened 2020-05-16
Removes SEA security access

Jaraci:
- MSEA and FSEA have both lost security access. This might stop every E-9 SEA in existence from trying to act like an ersatz COS/BC even when one or both of said roles are in the round already.


Roland410:
I disagree, SEAs often step in when asked by command, not on their own. I've personally not seen SEAs get in the way much if at all when playing as sec either, but some were being really annoying on sec comms, but that applies to BOs as well, when given a security key.

SouthyOsmanli:
Perhaps I come from a biased perspective - but SEAs, in the past that is which has left me personally with bad experiences, have attempted to use their rank and commands ‘approval’ to force themselves into security without even having access.

Ex; Marine SEA forcing a detective and brig chief out of doing questioning because ‘Command said I could’. Overruling brig proceedings, etc. I feel giving them access only encourages this gameplay further. 

SEAs should only be doing advising, not filling in for duties for security unless directly requested. Similar with bridge officer. The duties they typically are trying to preform are already designated to a department head or senior of the department (Brig Chief).

Sparrowbird611:
Gonna be honest.

I gave SEA security access because I feel he should have it, since it feels like part of his duty would involve going to the brig and visiting troublemakers, but I would absolutely remove the role if I could think of a good replacement.

BurgerLUA:
Poll is needed for this one.

Roland410:
Whoops, forgot to reply to this, sorry.

I mean, the above could technically happen, but if they do it because command said so, the problem is not with the SEA, but the command staff.
I often rely on SEAs, such as the excellent Sergeant Major Valeria Fraser to help me out with my duties when I am ROIC, and never had an issue with them or others, because I didn't give them orders to interfere, when the appropriate people were awake and active.
In the end this seems to boil down to:
1. SEA is incompetent and is overstepping their power, which can result with security as well (especially a lot of CoS' do that, I know from first hand experience).
2. Command forcing them to do so, and since it's usually a legal order, they don't/can't refuse it, nor do they speak up against it, which kind of goes back to point 1, them being incompetent.

BurgerLUA:
The people have spoken.

PR #883, opened 2020-05-18
Ascent Buffs

ProbablyCarl:
This PR was created in response to the current state, with a Gyne being a joke that can be killed by a laser pistol. The nerf of them was done by myself, given they used to be stupidly overpowered, with how Bay handled them, and this is to bring them back in line with the current meta.

- - -
Buffs:
 - Ascent particle lances given an overcharged mode, with the stronger being given to the Gyne's larger lance. They have a large movement delay, in comparison to other fire modes provided, and can't be burst fired. However, they have a significant increase to pen and damage. The only downside is the movement delay, and slightly less arterial damage chances.

 - Ascent RIGs given an incredible buff across the board, given that the Dagon now has an infantry compliment. Tested the buffs with both ballistics and energy/laser based weapons. Lasers are near entirely shrugged off by the armor, though ballistics such as shotguns/Tectonic and Z9s/Bulldogs still cut through them and can down a Gyne in a single mag.

 - Defence Lance given a guaranteed stun in the form of a seizure, upon a successful non-defended hit. This is to promote cuffing, or capturing, among other things. This also gives it use, given that it really had non prior. Aside from the 'seizure', it has an increased agony value, which helps quite a bit.

 - Alates given nanoblades within their rigs.

 - Ascent projected nanoblade given a 50% chance of parry by base, and the damage has been upped so it's comparable to an e-sword.

 - Basic lethal beam for the Alate particle lance given a damage increase.

 - Their speed has been nerfed to positive 0.5. What this means, is that instead of the -1 as prior, they're now .5 percent of the baseline. Slower, that is. 
This is to compensate for the above buffs.


Roland410:
I'm all for buffing the Ascent, but an instant seizure gun will be abused as fuck, I can already see it.
The overcharges also have a huge damage with insane AP as well, I think one of those values should be lowered. Either the damage, or the AP. iirc even sniper rifles don't have that damage caliber.

ProbablyCarl:
It's advanced particle weaponry for a group with already limited numbers, however. Atop the fact that the purpose of the Ascent, at least originally, and with this can be again, to be a spooky big ol' 'oh god' faction sitting in space. 
With Infantry aboard, alongside Sec, they're currently able to, within seconds, down a Gyne into DeathCrit. With this, it'll take a few more seconds. Especially if Alates are present.

OolongCow:
My main concern is that the original reason they were nerfed was to discourage them from just bumrushing the Dagon and mowing everyone down every round, as we had about two weeks straight where the Ascent spawned almost every single round and would find any excuse to come aboard and start shooting or kidnapping people.

I'm all for making them into less of a joke, but I don't know if one big PR going from 0 to 100, and merging many different buffs at once, is the right way to do it.

BurgerLUA:
Should probably testmerge this.

AdamElTablawy:
The gyne is already indestructible if it uses nab, they come with adminorazine, and they can plasmacutter walls down, but if you want them to be just indestructible, okey.

At the very least, buff the ascent drones too then.

ProbablyCarl:
We've had it test merged.
They're not nearly as overpowered as you'd imagine.

AdamElTablawy:
@ProbablyCarl 

Respectfully disagree.

Tested with Don earlier before this testmerge.

Ascent are insanely strong. It's just that our players are retarded, and don't use their shit properly.

If you try to attack head on as the Ascent, you'll get tectonic'd. Same as any antag. The issue is that all Ascent attack head on and rush down a hallway with no prep.

Ascent don't suffer attrition-crystal brothime heals incredibly quickly and treats organ damage, rigs stabilize fractures. It'd be comparable to adminorazine if it was any stronger.

Ascent have powerful armor-piercing particle rifles. The gyne's can pen ablative, even.

Alates are fast, and the gyne is virtually immune to ranged attacks due to Nab giving her a personal bodyblock meatshield. They all have thermals, so they can pick and choose precisely where they engage.

The flechette rifle penetrates walls, so the gyne can just shut down people outright when combo'd with thermals.

They all have plasmacutters.

In short, the ideal way to be a threat as ascent is hit and running from space-breach the walls with your plasmacutters and open fire, depressurizing areas as you go. Nobody will be able to fight you due to your guns ignoring armor and if you get hit, you can just mach 5 away and heal. Alates are virtually impossible to hit in the first place due to their speed unless you're a total brainlet and are inside the Dagon's tiny fucking hallways and being confronted by armed sec iin arm's distance, rather than sax'ing around in space.

Or you can just snipe people with flechettes through walls while alates guard you.

If you do that strategy, there is literally nothing anyone can do to stop you ever-fighting defensively or in space or through breaching, the Ascent are virtually invulnerable because they can't die to attrition ever because of crystal brothime. 

So they're already strong enough. The only way they aren't boogeymen that everyone is pants-shittingly terrified of is that people are just retarded and try to fight head on, and get killed by the tectonic as a result because they negate every single advantage they have. It's the same things that make traitors lose-idiocy, not game mechanics.

So yes, they are EXACTLY as strong as I think they are-completely unbeatable so long as even one of them is intelligent, and dealt with if they play stupidly. 

There's no reason to buff them to be able to facetank shit as well. That'll just invite murderbone and Ascent aggression.

BurgerLUA:
Failing travis.

PR #884, opened 2020-05-18
GAS can now be passengers

Superhats:
Literally every other species in the game (Minus vox for obvious reasons) can be passengers.


Roland410:
GAS are bred and trained for a specific job with our current lore afaik, so them just being passengers is a bit strange.

Sparrowbird611:
Not really against this, but Roland does kind of have a point. Though you could argue GAS botanist and things like that could exist.

Superhats:
Adherents are ALSO built with a specific job in mind, yet they can be passengers too. Just like an adherent investor, a GAS investor will be noticed quickly, and will probably get in trouble.

Roland410:
Adherents were built to serve, it's a huge difference in my opinion. GAS have been trained by Xynergy to perform specific tasks.
Technically GAS are I'd say slave-like, while Adherents serve people as they have nothing else to do with their life.

Superhats:
Still, I don't see why a GAS wouldn't be able to go off-duty, if anything.

Superhats:
Surely they don't work 24/7, they _are_ sapient and that would take a toll on their mental health.

BurgerLUA:
Needs a poll.

ProbablyCarl:
Poll is a resounding 'no'.

PR #886, opened 2020-05-18
Adds an new "energy Tectonic" to the SEC-ARM

Jaraci:
Oh boy. All values here are very WIP.

- Adds the BXJ-5 Culverin, a 3-round pulse rifle variant, to the SEC-ARM. It can fire 12 bursts before needing recharged.
- The Culverin has two fire modes: high-frequency fire, which inflicts burn; and experimental fire, which inflicts brute damage instead at the cost of raw power.
- The Culverin requires specific authorisation for both fire modes, like the Tectonic. It cannot be fired otherwise unless on red alert or above.
- The final shot of each burst has some dispersion to it and will occasionally go wide. The Tectonic's penetrating power meant it caused some collateral damage when used; this should help replicate this effect.

Security got a flat-out nerf with the addition of infantry. This should help rectify that without just giving them a second Tectonic again. It uses the pulse rifle sprite for now, but considering that pulse rifles are actually just flat-out not used anywhere I don't foresee this being an issue.

High-frequency (burn damage) mode is intended to be the default, with the brute damage secondary mode being an inferior alternative in terms of damage.

Feedback is very much appreciated. Thanks to @ProbablyCarl for some help with this PR.

### DIFF
--- a/code/modules/projectiles/guns/energy/secure.dm
+++ b/code/modules/projectiles/guns/energy/secure.dm
@@ -89,3 +89,6 @@
 	projectile_type = /obj/item/projectile/beam/pulse/xray
 	req_access = list(list(access_brig, access_bridge))
 	authorized_modes = list(UNAUTHORIZED)
+	firemodes = list(
+		list(mode_name="fire", burst=1, fire_delay=null, move_delay=null, one_hand_penalty=8, burst_accuracy=null, dispersion=null),
+		)
--- a/code/modules/projectiles/guns/energy/secure.dm
+++ b/code/modules/projectiles/guns/energy/secure.dm
@@ -90,5 +90,6 @@
     req_access = list(access_brig)
     authorized_modes = list(UNAUTHORIZED)
     firemodes = list(
-        list(mode_name="fire", burst=3, fire_delay=null, move_delay=null, one_hand_penalty=8, burst_accuracy=null, dispersion=null),
+        list(mode_name="high-frequency pulse fire", burst=3, fire_delay=null, move_delay=null, one_hand_penalty=8, burst_accuracy=null, dispersion=list(0, 0, 1), projectile_type=/obj/item/projectile/beam/pulse/xray),
+		list(mode_name="experimental pulse fire", burst=3, fire_delay=null, move_delay=null, one_hand_penalty=8, burst_accuracy=null, dispersion=list(0, 0, 1)projectile_type=/obj/item/projectile/beam/darkmatter/burst),
         )
--- a/code/modules/projectiles/guns/energy/secure.dm
+++ b/code/modules/projectiles/guns/energy/secure.dm
@@ -82,7 +82,7 @@
 
 //Energy Tectonic
 
-/obj/item/weapon/gun/energy/pulse_rifle/shotgun
+/obj/item/weapon/gun/energy/pulse_rifle/sec
     name = "BXJ-5 Culverin"
     desc = "A somewhat experimental product of Ward-Takahashi's energy weapons division, the BXJ-5 combines rapid-fire pulse laser technology \
     with penetrative laser power for the first time. Designed to saw through armour at the cost of raw damage."
--- a/code/modules/projectiles/guns/energy/secure.dm
+++ b/code/modules/projectiles/guns/energy/secure.dm
@@ -91,5 +91,5 @@
     authorized_modes = list(UNAUTHORIZED)
     firemodes = list(
         list(mode_name="high-frequency pulse fire", burst=3, fire_delay=null, move_delay=null, one_hand_penalty=8, burst_accuracy=null, dispersion=list(0, 0, 1), projectile_type=/obj/item/projectile/beam/pulse/xray),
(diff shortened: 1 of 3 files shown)


Sparrowbird611:
First off, why do we need ANOTHER gun to ignore armor. This is literally unnecessary.

Secondly, jeez, you gave it 36 shots? What in the world man.

Jaraci:
It is locked to three-shot bursts because it is a pulse rifle. You can only actually click twelve times.

It ignores a shitload of armour because that was the Tectonic's original gimmick -- ignoring armour at the cost of raw damage.

Sparrowbird611:
Meh. I just don't see why security actually needs an armor ignoring gun. They already have plenty of decent weapons, I see no reason to give them this one.

Though if it does go through, at least we can say that the plasma pulse projector does something now.

Roland410:
The sprite should probably be changed, as antags can still buy a pulse rifle afaik.

Jaraci:
@Roland410 fair point, although I have never once seen anyone actually use it because of how dogshit it is. I'll try to see what I can do in that regard

Roland410:
True, but you can probably do some minor changes so it's distinct from the pulse rifle.

AdamElTablawy:
this does 54 dmg per full burst

this is effectively just an infinitely better laser carbine there's no reason not to take because it's hitscan

it'll fleece anything including ascent armor and merc armor and do serious amounts of blood evaporation and lung damage, and without the usual tectonic drawback, which is running out of ammo/penetration/friendly fire

i'd be careful with this tbh

Sparrowbird611:
If this has to go through, I'd lower the damage far more, to ten at most for laser mode, and two less for the brute mode.

Actually, why even have two modes is what I have to ask. It's not really a bad thing but it's an odd addition tbh.

Roland410:
Going to agree with the above. A hitscan laser with huge AP like this combined with the damage values is a tad bit too high. 10 damage for both modes seem alright, as it still does 30 damage then.

Roland410:
Also you should probably mention that you're removing the teargas nade and the ziptie boxes to place this on the map.

Jaraci:
With lower damage the burn laser mode tends to fall apart regardless of penetration and seems to have severe issues applying organ damage at all, which turns it into little more than a flashlight. I could solve this by making it single-shot instead of a pulse gun, but I'm somewhat sold on the idea of a pulse laser as-is.

I could cut the burn laser mode entirely and drop the damage on the new primary fire without too much issue, though.

Roland410:
I still don't understand why people insist so much on every weapon having to fuck each internal organ of your target. 99.9% of the time your job is not to assblast your target, but to capture them.

Sparrowbird611:
> I still don't understand why people insist so much on every weapon having to fuck each internal organ of your target. 99.9% of the time your job is not to assblast your target, but to capture them.

I get why they want it, since organ damage is the best way to cuck most people, since it's harder to recover from than just normal stuff, but in this case, it's security. I don't feel they even need heavy penetration weapons like this tbh, let alone "ORGAN DEATH MELTORATOR CANNON."

Jaraci:
Without organ damage, energy weapons are essentially useless unless they just flat-out stun. Burn damage is 100% worthless -- it can't cause arterial or break bones or do anything except turn the target's screen white and potentially make him have a heart attack five minutes down the line. If a laser weapon can't carry over any of its damage into organs there is no point in using it; it is essentially non-lethal due to how brainmed works.

Roland410:
Laser weapons evaporate your blood and that's why they are actually dangerous, not because they microwave your organs.

AdamElTablawy:
@Roland410 All weapons must do organ damage to be useful as a weapon, due to the nature of brainmed.

If it merely applies damage without doing organ damage, then it's basically a glorified taser.

Roland410:
@AdamElTablawy The real killer in brainmed other than straight up brain damage is... *drumroll* blood oxygenation. Low blood oxygenation is caused by lack of oxygen or blood, which lasers evaporate from you.

AdamElTablawy:
@Roland410 Fair enough, but i'm not actually sure how much bloodloss this actually ends up doing.

Roland410:
Considering you get 12x3 shots, assuming you only hit even half of that, that's 18x10 (with the suggested values), which should be 180 damage, more than lethal enough by itself, and you honestly don't need to do much damage to get someone below 80% oxygenization, so I'd go with quite major.

AdamElTablawy:
Either way as it currently stands this is 'massive overkill: the gun'.

I'd have difficulty justifying taking anything else. With the amount of damage it does, it's easy to incapacitate someone, yet hard to kill them.

BurgerLUA:
More feedback is needed for this PR. Maybe a poll is needed if no additional feedback is given.

Jackofashepard:
Personally i think sec's in a fine spot with their new bullpups but this could have a place as a antag item.

BurgerLUA:
Is this PR still being worked on? It's failing tavis and has conflicts.

Superhats:
Closing this. It's conflicting, and the PR owner has left the server. Feel free to contact a maintainer to re-open it.

PR #888, opened 2020-05-19
Removes renegade from infantry roles

Jaraci:
- Infantry can no longer be renegades.

I might extend this to forensics techs, brig chiefs and the like as well, just to be comprehensive.


Sparrowbird611:
Gonna be honest, I feel renegade in general should be removed, but these at very least needed to go.

nerocavalier:
I'm against Infantry as renegades because it encourages them to valid antags.

Roland410:
I don't agree with the planned removal of renegade from BC and FT, because it gives them more breathing room to do not-so-regular stuff, but infantry is a role whose purpose is to fight offsite mobs and protect exploration.
I'd say, that most if not all antag roles are incompatible with the role (I personally disabled them as well on my infantry character).

ProbablyCarl:
#890

PR #891, opened 2020-05-19
Changes the Deck Officer's superior to the CO, other minor changes

Roland410:
🆑Roland410
tweak:Makes the Deck Officer's superior the CO, considering they basically the head of supply.
tweak:Gives the Deck Officer, Pathfinder and Squad Leader access to their departmental budget.
/🆑
As per above. They have been given the "head_position" variable which makes them spawn with the departmental budget's information in their memory, but otherwise doesn't change anything.


Sparrowbird611:
Good stuff, one thing.

Why exactly did you feel the need to swap the command position from the XO to the CO. It's not necessarily bad, but at the end of the day, the XO is the CO's hand, and the CO is over the XO. Just seems a little pointless, but otherwise fine.

Roland410:
> Good stuff, one thing.
> 
> Why exactly did you feel the need to swap the superior from the XO to the CO. It's not necessarily bad, but at the end of the day, the XO is the CO's hand, and the CO is over the XO. Just seems a little pointless, but otherwise fine.

It's in line now with other heads of staff. Technically if there is no CO, they will answer to the XO/ROIC anyways, and the XO can still pull rank if necessary. This should give them a bit more power so to say.

PR #892, opened 2020-05-19
Adds (nerfed) Skrellian voidsuits + helmets to xenowear loadout for Dagon Skrell

Jaraci:
- Adds an exchange variant of Skrellian voidsuits and helmets to the loadout menu. They have worse melee and ballistic protection than security voidsuits, but have an equal level of laser protection. They have a middling level of radiation shielding and some energy protection, too.
- These voidsuits can only be picked by command and security roles, as well as possibly a few others. Infantry Skrell might theoretically be able to pick them as well, although Skrell cannot be in the SMC so that's not really relevant (for now???)
- They can, of course, only be selected by Skrell in the loadout menu. Other species can wear them, though, I think; they just can't spawn with them.

Command level Skrell (BOs in particular) have actually had no voidsuits specifically available for them for a while now, since the command voidsuits cannot be cycled. This should help with that, as well as add a bit of flavour.


Roland410:
No, no, no. I'm fully against this. Loadout items are meant to be gimmicky stuff that don't provide any real advantage. Being able to spawn with a voidsuit goes pretty much against that. I'd recommend you resolve the issue with the cycling instead of this.

Jaraci:
The intention is for it to be a sidegrade at best to the security/command voidsuits themselves. Furthermore, I've been thinking for a while about how to differentiate Skrell from just being essentially cold humans with big heads -- unathi get their super regen and claw attacks, prometheans get their regen and gimmicky shapeshifting, and so forth -- and playing off the idea of their alternate (superior?) technology seemed like a good idea.

They're locked to roles that have easy voidsuit access anyway, and I doubt having your own personal sidegrade to what's already available in spades in storage units across the Dagon is going to be compelling enough for us to see a new wave of Skrell players. They must be one of the least-played species on the ship.

Sparrowbird611:
No real comment, though I lean towards Roland's view. Though I'm not highly against it so long as they're not back pack sized or anything.

Superhats:
Skrell can already spawn with a helmet, so the loadout isn't /exclusively/ gimmicky stuff.

Superhats:
Plus, this further cements the idea that skrell exchange personal has access to their own gear (provided by their SDTF) and helps add more flavor.

Roland410:
That may be so, I'd have to check, but that's still not a space suit though.

Sparrowbird611:
I am still against mechanical advantages in the loadout in general, but to play devils advocate here, I do want to point out several surprisingly strong bonuses do already exist in the loadout, namely augments such as the toolset implant, and the cooling unit for IPCs.

Roland410:
The only useful implant out of the two is the surgical, considering how easy it is to acquire tools, buuuut also that was added by someone here on Bastion before I was even a player. At least the damn combat augs got purged from it tho.

BurgerLUA:
Needs more feedback.

PR #893, opened 2020-05-20
Port Aurora Tajaran Sprites and haircuts

Superhats:
Tajaran now have some new sprites courtesy of Aurora

Those should color MUCH better, among other things.

They also get new haircuts — in the genemods tab — featuring **in-built ears**

This also gives them a nice autohiss in the forrrrm of rrolled rrrr's


code/modules/mob/living/autohiss.dm
```diff
@@ -26,6 +26,12 @@
 					LANGUAGE_UNATHI_YEOSA
 	)
 
+
+/datum/species/tajaran
+	autohiss_basic_map = list(
+			"r" = list("rr", "rrr", "rrrr")
```
BurgerLUA, line 32: Why is this here?

Afya212: I'm working on a PR with Weezer (Lore Master) for Taj. This may or may not be added but I say wait and see what ends up being decided.

Superhats: From what I've read of tajaran lore, they've got a mild accent. (Or at least, they do on Aurora). Considering bay has basically no tajaran lore, I don't know if that's still accurate.

Afya212: Taj lore is being worked on right now, would recommend just removing it now and waiting until then.

Superhats:
According to feedback (and because I've messed my branch up) I'll be reopening the PR without the cosmetic changes, making it 100% sprite changes.

PR #895, opened 2020-05-21
Nerfs bullpup rifle damage, buffs shotgun slug falloff

Jaraci:
- Bullpup frangible damage cut by 1/3, from 60 > 40. Why the fuck these things did as much damage as a shotgun slug while having a higher ROF, penetrating walls and mobs, only slightly worse penetration and 35 shots instead of 8 is kind of mind boggling.
- Regular bullpup damage bumped down a bit so that frangibles still technically hit harder. Maybe unnecessary.
- Shotgun slugs given much less accuracy falloff. There's no reason they should be hitting pistol levels of distance falloff.


Sparrowbird611:
All good, I think you ought to raise the fall off on the shotgun to 1.5 to 2 though.

Jaraci:
Maybe. In all honesty I just got very annoyed when I took a look at the values and saw just how outmatched the combat shotgun is by these "worse" rifle magazines. Even the buffed Ascent armour values saw an Ascent raiding party getting unceremoniously clapped by infantry with frangibles.

I'd be happy to bump it up to 1 or 1.5, which makes a slug shotgun about as accurate (or slightly less accurate) as/than the laser carbine, which is pretty much its energy counterpart.

Sparrowbird611:
> Maybe. In all honesty I just got very annoyed when I took a look at the values and saw just how outmatched the combat shotgun is by these "worse" rifle magazines. Even the buffed Ascent armour values saw an Ascent raiding party getting unceremoniously clapped by infantry with frangibles.
> I'd be happy to bump it up to 1 or 1.5, which makes a slug shotgun about as accurate (or slightly less accurate) as/than the laser carbine, which is pretty much its energy counterpart.

Agreed. I don't know what was up with the frangible meme rounds, but they had some sick damage for automatics.

Yeah, main reason for the increase suggestion was because it feels a little meme to have a shotgun be more accurate than a laser.

Jaraci:
Fair point. 1.5 it is.

PR #896, opened 2020-05-21
Buffs buckshot

DriedCrust:
Gave 10 AP to each buckshot pellet followed by increasing it's damage from 30 to 35. This should finally make it useful.

edit: toned it down to 5 AP


Sparrowbird611:
Going to say it again here, don't buff the damage. It doesn't need the damage buff, it's issue is AP.

As for actual AP, I would have started at 5, and gone up to 10 If that failed to make it viable, but I won't argue against it.

Jaraci:
Give it 5 AP, not 10.

PR #900, opened 2020-05-22
Branch/rank tweaks for diona, adherent, and prometheans

Jaraci:
- Diona restricted to civilian branches. They can still be forensics techs, though -- I kept those branches specifically open for them. No more FSEA or CE Dionae, though, because that was mind-blowingly nonsensical.
- Adherent restricted to civilian branches. On the plus side, they get a new job out of the deal. Adherent can now be doctors.
- Prometheans restricted to enlisted roles within the NTEF. They are overgrown lab experiments with what little lore exists for them; their TROOPER_ONLY status within the SMC should carry over to the NTEF here. I think civilian only is too extreme, so I cut a middle ground between giving them the full permissions of the other races (which are just human genemods, by and large, in lore, or at least stable species in their own right) and restricting them to a tiny number of jobs.


Sparrowbird611:
I like these, two slight things.

Wasn't there an issue with allowing adherent doctors last time? I tried to do this, and the PR had to be reverted for some reason.

I've heard adherent surgeon doesn't actually work properly, since they can't wear gloves, causing constant infections in their surgical patients. Can probably be fixed with spaceacillin injections, but something to keep in mind.

Jaraci:
Oh, yeah. I forgot about that. I wonder if there's any extra check I can include for that whole infection thing.

Jaraci:
@Sparrowbird611 See #901 for adherent surgery thing.

Sparrowbird611:
Nice. I don't know if the issues they had previously with lore/staff/whatever it was still apply, but I support this PR either way.

Jaraci:
As far as I remember, the only issue was that they literally could not stop turboinfecting patients due to having no ability to wear gloves.

reklaw9:
why make playing prommie even worse?

Roland410:
I don't agree with any of these changes other than allowing adherents to be doctors.
This is once again a biased PR against prometheans and dionaea, while it allows every other SMC trooper locked species to be full NTEF.
By the way Prometheans are not lab grown, but I haven't got around to piecing together their lore bits we have, as I will be reworking them anyways.

Higgin:
lore is a spook, and if somebody wants to do the slowest uniform inspection in the world as diona fsea more power to them

same goes for plasmapeeps who try to change into a formal uniform and just catch fire - i love it and it's not bad

adding adherent doctors is fine though

Roland410:
I actually halved the slowdown of dionaea a few days ago, so they go with acceptable speeds now, just can't run.

Jaraci:
@Higgin Diona are pacifists (of a sort, from what I gather they understand violence but just cannot understand _why_ it would ever be used. I really don't see how one could ever get into a military or paramilitary organisation. Forensics tech is already a bit of a meme; SEA is too far.

Also, none of this PR touches plasmamen at all, I think you're getting mixed up.

BurgerLUA:
Needs more feedback and probably a poll.

BurgerLUA:
Poll is up.

PR #901, opened 2020-05-22
Quick and dirty adherent surgery fix (maybe)

Jaraci:
- Adherents can (maybe) now do surgery without killing their patient from infection overload. This check _should_ make it so that the adherent's own `germ_level` is ignored in any surgery step, and was the only thing my tiny brain could think of implementing that was within my own level as a coder (read: not a coder) to do. I have tested this and it seems to work and also not break anything.

### DIFF
--- a/code/modules/surgery/surgery.dm
+++ b/code/modules/surgery/surgery.dm
@@ -166,6 +166,8 @@ GLOBAL_LIST_INIT(surgery_tool_exception_cache, new)
 	var/germ_level = user.germ_level
 	if(user.gloves)
 		germ_level = user.gloves.germ_level
+	if(user.get_species() == SPECIES_ADHERENT)
+		germ_level = E.germ_level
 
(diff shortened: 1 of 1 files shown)


BurgerLUA:
Can you describe the nature of the bug so I can see whether or not this fix is good or not?

Jaraci:
@BurgerLUA as I understand it, adherents physically cannot wear gloves due to having no hands. They accumulate a `germ_level` just like any other mob but can't replace it with a gloves' `germ_level` and then wash their hands to make themselves sterile for surgery.

This quick fix simply sets the organ's existing germ level equal to itself every time the germ check is called during surgery, which should be more or less functionally identical to having a doctor with just-washed hands and clean gloves doing the surgery.

BurgerLUA:
it should be 

```	
if(user.get_species() != SPECIES_ADHERENT)
 E.germ_level = max(germ_level,E.germ_level)
```

PR #907, opened 2020-05-23
Sent to Lobby

Afya212:
Adds a new button in the player panel which allows admuns to send ghosts directly back to the lobby screen.

A literal port of my PR here https://github.com/DS-13-Dev-Team/DS13/pull/368


Roland410:
I think making it auto-ghost a player would be useful, just give a prompt to the admin, that they are in the body of X.

PR #909, opened 2020-05-23
Fixes the pilot suit cycler.

Sparrowbird611:
Fixes the pilot's cycler to use pilot access, instead of mining office access.


BurgerLUA:
Failing travis.

```code/game/machinery/suit_cycler_units.dm:53:error: access_pilot: undefined var```

PR #921, opened 2020-05-27
Flashdark buffs

Jackofashepard:
:cl: Jackofashepard
tweak - TC for Flashdark reduced to 20 from 32
tweak - Item sizechanged to that of a normal flashlight
tweak - 1 tile increase to range 
:cl:


Jackofashepard:
While the flashdark is a great gimmick its very finnicky to use due to it fitting on close to no slots and costing a staggering amount considering its useless without thermal goggles, I'm hoping this PR makes it easier to use and more tempting but it still doesn't match the prowess of the chameleon projector.

I am open to input and discussion on these changes.

ProbablyCarl:
Looks fine.

chaoko99:
Flashdark is complete cancer, and this is inadvertently reducing a NERF that was made because the permanent dark, "ha you can't click me anymore" blob was cheaper than a single grenade that did the exact same thing for 3 seconds.

chaoko99:
People didn't use it much because it's expensive, but again, this is a COMPLETE COUNTER to any gun. You literally cannot shoot someone who's in the dark blob. That's why cultist constructs were so good.

Sparrowbird611:
I like the whole thing, however, I think you've made it sliiightly too cheap. I'd go at about 20-26 TC.

bearrrrrrrr:
i'd keep the cost at maybe 30tc if you want to increase the range

it is a *very* powerful item and the only reason people don't use it more is because they don't know how good it is

-1 from me til then

Jackofashepard:
In my opinion its really not that good, its really easy to figure out where the center is and if you hover over it you'll see "unknown" or whatever. Calling it a complete counter to guns is a overstatement, however it is a slight aid to not getting shot because if you move they need to figure out where center is again (doesnt take long). As for the price, keep in mind this item is useless without thermal goggles which are 24 TC bringing the actual total cost of the flashdark as it is in this PR to 40 TC.

chaoko99:
>As for the price, keep in mind this item is useless without thermal goggles 

Explain. Why does the player need to be able to see themselves to shoot others?

Jackofashepard:
Basically in order to walk you need to see where you're going and whats in your way

PR #922, opened 2020-05-27
Revert "Tajaran Resprite + Haircuts"

Superhats:
Reverts #915 

The changes in sprites has caused pretty much every tajaran player's character to become slightly darker across the board. While I thought it wouldn't be that impactful, our four tajaran players aren't happy with this.


BurgerLUA:
Honestly I don't understand why this needs to be reverted instantly. If darkness is a problem then it just needs to be retweaked instead of reverted like this.

PR #925, opened 2020-05-28
Little unathi glove tweak.

Sparrowbird611:
Just makes it so they work like normal gloves when it comes to forensics. They're just like regular gloves for the most part now.

### DIFF
--- a/code/modules/detectivework/forensics.dm
+++ b/code/modules/detectivework/forensics.dm
@@ -62,9 +62,6 @@ obj/item/var/list/trace_DNA
 		var/mob/living/carbon/human/H = M
 		if (H.gloves && H.gloves.body_parts_covered & HANDS && H.gloves != src)
 			if(istype(H.gloves, /obj/item/clothing/gloves)) //Don't add prints if you are wearing gloves.
-				var/obj/item/clothing/gloves/G = H.gloves
-				if(!G.clipped) //Fingerless gloves leave prints.
-					return 0
 			else if(prob(75))
 				return 0
 			H.gloves.add_fingerprint(M)


Roland410:
I'm fairly sure that a fingerless glove will leave fingerprints behind.

Sparrowbird611:
I made this because the only people who use fingerless gloves are unathi, and given we have no other fingerless gloves I can think of, I thought this would be the easiest fix.

Realistically, it may work that way, but I feel realistically fingerless insulated gloves would also result in you getting electrocuted, so I kind of just stopped caring.

Roland410:
Sounds like an oversight on that part tbh.

BurgerLUA:
Sounds like an oversight. I'm pretty sure fingerless insulated gloves are called "unathi insulated gloves" or "modified insulated gloves"

Regardless I don't like this PR.

Sparrowbird611:
All unathi variants are called modified gloves. Either way, I don't see the reason for making it harder for unathi to beat forensics than for any other race. No one else wears modified gloves so, this is literally just a them thing.

BurgerLUA:
> All unathi variants are called modified gloves. Either way, I don't see the reason for making it harder for unathi to beat forensics than for any other race. No one else wears modified gloves so, this is literally just a them thing.

A non-human race having drawbacks is fine.

Sparrowbird611:
Having draw backs is fine, but I'm not a fan of forcing unathi antags down the loud mode path, which is what I'm trying to fix by doing this.

BurgerLUA:
A better PR to solve this issue is to make it so that unathi gloves that you get at roundstart are modified gloves that don't have the clipped variable. I don't want to merge a PR that would make it so that fingerless gloves don't actually leave fingerprints when used.

Sparrowbird611:
> A better PR to solve this issue is to make it so that unathi gloves that you get at roundstart are modified gloves that don't have the clipped variable. I don't want to merge a PR that would make it so that fingerless gloves don't actually leave fingerprints when used.

Alright, that's fair enough burger. I'll work on figuring that out at a later date. I'll close this one for now.

PR #929, opened 2020-05-31
Changes Observatory

bearrrrrrrr:
Adds a nice lil comfy wood-tiled area to the end of D3 hallway/changes the observatory to be an actual proper area instead of a weird tumor

> why dis is goaed for gaem?

well for one it has a pair of glass windows and a door so if someone blows up the supermatter it won't vent the whole fucking hallway anymore

it also adds a comfy lil area to the dagon which is super nice :)
image

### DIFF
--- a/maps/torch/job/command_jobs.dm
+++ b/maps/torch/job/command_jobs.dm
@@ -358,7 +358,7 @@
 	selection_color = "#2f2f7f"
 	minimal_player_age = 0
 	economic_power = 7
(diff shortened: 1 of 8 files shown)


bearrrrrrrr:
wait

WHY DOES THIS HAVE MY AGE CHANGES IN IT

WHA-

Roland410:
I personally like the current design more over this, and also we have way too many "nice comfy areas" that are not used. There is the garden, the lounge/library, we have the new mess hall etc.
The huge table and the floor also ruins the experience for me, on top of making travel between the two maintenance sections awkward as fuck.

PR #939, opened 2020-06-07
Changes the cost of stims and zerk

OolongCow:
Stimpack: 3TC > 10TC
Oxycodone is very powerful, and 10 units of it mixed with 5 of inaprovaline is enough to keep someone standing long after they should have collapsed. Considering the autoinjector for some reason contains three times the volume of normal autoinjectors, 10TC is a much fairer price for what is basically three to four minutes of invulnerability inside of a hardsuit. Traitors shouldn't be able to carry eight of them in a wallet and walk around with their guts hanging out, perfectly happy.

Zerk: 5TC > 15TC
For reasons similar to my reasoning with stims, three doses of a synaptizine and hyperzine mix shouldn't be as cheap as it is, especially when there's nothing preventing you from pairing it with stims and just becoming immortal. More than one combo of that should require more TC commitment than the loose change you had after buying a gun, ammo and armor.

:cl: OolongCow 
tweak: changes the cost of zerk and stims
/:cl:


GoldenSovietFox:
I think this will balance the idea when the other types of items will have changed price. If not, it possible to code it in such that you can not use the stims with zerk? That would be also a moderate way to balance. I might not know how good they are because I have never actually used them extensivelly.

OolongCow:
> I think this will balance the idea when the other types of items will have changed price. If not, is it possible to code it in such that you can not use the stims with zerk? That could be perhaps a moderate way to balance it. I might not know how good they are because I have never actually used them extensivelly.

Stims and zerk are really just chem cocktails. 10 units of oxycodone and 5 units of inaprovaline for stimpacks, and 3 units of synaptizine (stun immunity) and 12 units of hyperzine (speed increase) for zerk. Making them impossible to use together would probably look something like making Oxycodone + Synaptizine create a poison when mixed.

Sparrowbird611:
I'd go 7 for stim, 9 for ZERK tbh.

Sparrowbird611:
> I think this will balance the idea when the other types of items will have changed price. If not, is it possible to code it in such that you can not use the stims with zerk? That could be perhaps a moderate way to balance it. I might not know how good they are because I have never actually used them extensivelly.

Stims and zerk are essentially the exact same thing, with the additions of hyperzine and synaptize that cow already mentioned in zerk.

As for how good they are, as some one who abuses them regularly, they're pretty much a must for any loud mode antagonist who actually wants to defeat security in a gun fight, otherwise, stick to ganks in maintenance with the sleepy pen.

OolongCow:
> As for how good they are, as some one who abuses them regularly, they're pretty much a must for any loud mode antagonist who actually wants to defeat security in a gun fight, otherwise, stick to ganks in maintenance with the sleepy pen.

I'd like to find a price point where they're still absolutely affordable for traitors, and something they can definitely load up on if they like, but not something they can buy an effectively endless supply of while also having other meta gear. If someone wants to buy enough stims and zerk to use for the ENTIRE average length of security hunting them, it should really start dipping into their funds and cost them the ability to buy other things. With how things are priced right now, it's way, way too easy to grab an Emag, AR with extra ammo, and good handful of stims, with just a normal uplink. The intent is to make it so that if you're dedicated to buying big box items, you have to give up a few doses, or if you're dedicated to stims and zerk, you have to maybe settle for a weaker gun than the best in the game. Johnny the unstunnable, painless, no slowdown tweaker should be using a magnum, not an AR, if he intends to stay constantly dosed.

OolongCow:
As a note, ALL of these drugs benefit from the Slow Metabolism trait, as their most prominent effects come from being in your system, not from being used up to heal damage (like bicaridine or kelotane). The length of time antagonists can stay stimmed off a single dose with that trait is disgusting even at my suggested price point.

Jackofashepard:
I'd argue stimms 3 > 6 TC and zerk 5 > 10 TC, they are good meds but dont do shit for organ damage. You die just fine as soon as your liver goes carbon

OolongCow:
> I'd argue stimms 3 > 6 TC and zerk 5 > 10 TC, they are good meds but dont do shit for organ damage. You die just fine as soon as your liver goes carbon

That's true, but the main source of organ damage from security (bullpup frangibles) are going to be blowing your insides up 50% less often due to #941

PR #938, opened 2020-06-07
Adjusts the costs of "loud" traitor items

OolongCow:
Rather than nerf good weapons, the costs of several have been increased to come more in line with equally as powerful weapons. Additionally, the costs of several weapons that see next to no use has been decreased to encourage traitors already taking "core" items (emags, parapens, etc.) to pick them up with their spare TC.

Replica Pistol: 16TC
To give a lot of context to the later changes, let me start this off by saying that while I have not changed the Mk58's TC cost at all, it's still very important. It's one of the few traitor items in this category that is fairly and intelligently-priced, and a lot of the changes I made to cheaper guns has to do with their power relative to it. This is the "baseline" gun.

Zip Gun: 4TC -> 1TC
I don't expect anyone will take this even now, but this leaves it more available for gimmicks or using up the very last telecrystal you have.

Small Energy Gun: 16TC -> 8TC
It's never used and is pitifully weak for what it is. Hopefully now it will find its way into the hands of traitors with a few TC left over.

Dart Gun: 20TC -> 12TC
An alternative to the syringe gun. Seeing as the syringe gun is easily stolen by anybody who's likely to be making chemicals anyway, reducing its price should free up more TCs for non-medical traitors to buy chems that they can load into it.

Energy Crossbow: 20TC -> 32TC
An incredibly cheap, ammo-less, self-recharging weapon should not be near-silently and instantly downing anybody it's shot into. Touching its TC cost is an alternative to simply removing the stun, which would make it worthless.

Energy Sword: 32TC -> 54TC
Let's be honest, here. The energy sword has become the weapon of choice for traitors who just want to click spacemen on par with the assault rifle. At its current cost, which is lower than a machine pistol with a single magazine, it totally crowds out all other options at its price point, and is pitifully easy to buy along with anything else a traitor could want. Inversely to the decreases in prices in this PR, which are meant to ALLOW traitors to take weaker weapons with spare TC, it's very apparent that when you pick the energy sword, you should be committing a lot of your TC to it, as is rightfully the case with the AR. Someone who buys the energy sword should be very dangerous, but not have as much wiggle room as they currently have to buy any and every thing they could ever want from the uplink. The energy sword should be the main event, not something literally every traitor buys because it's dirt cheap and stronger than everything else in the price range.

Small Silenced Pistol: 32TC -> 20TC
[...]

PR #941, opened 2020-06-07
General gun balance pass

OolongCow:
Attempting to try and encourage use of underused security and antag weapons. Make the bad guns useful and the good guns less fought over.

- Increases the number of pellets in a buckshot shell (6 -> 8), decreases damage per pellet by 5 (35 -> 30), increases range_step from 1 to 2 (Instead of going 8>7>6>5>4>3>2>1>0 pellets and losing 10% chance to hit the targeted body part per tile traveled, it now goes 8>8>7>7>6>6>5>5>4>4>etc and 10% every other tile). Adds a 2x multiplier to embed chance. This change should put the unchanged slugs back into the role of punching through armor and concentrating damage on one body part, rather than being the all-around best ammo type in every situation. Testing shows total damage per shot (pellets spread to different limbs than the targeted one dependent on range) to be on par with the bullpup at 5 tiles against medium plate.
- Hopefully makes rubbershot not an absolute joke (it legitimately did almost half the agony of a smartgun shot, on top of spread and travel time). Pellets up (4 -> 10) as well as agony (4 -> 8.5). range_step from 1 to 2 (see above for explanation). 85 agony in absolutely ideal conditions against unarmored targets is fine. Damage per pellet increased from 1 to 2. Also no longer embeds (since it was originally birdshot).
- Beanbags are now the less-than-lethal equivalent of slugs they were always meant to be. They now possess 10 armor penetration (0 -> 10), but do 10 less agony (75 -> 65). They now also do 5 less damage. (15 -> 10).
- Why did the sawnoff have a greater one-handed penalty than the double-barrel? The world may never know. Corrected. (4 -> 2)
- Non-magnum pistol rubber bullets now have 5 more agony (35 -> 40, 20 -> 25) and 5 armor penetration (0 -> 5).
- Military pistol fire delay buffed (7 -> 6). Now matches the Mk58 (Replica Pistol).
- Reduces shrapnel and internal bleeding chance by 50% for bullpup frangibles (2.0 multiplier -> 1.5). Every bullet should not be cutting an artery or getting caught in your insides. Places it on par with the AR's internal chance.
- Reduces security highcaps to 30 round magazines, down from 35. Should also affect infantry's.
- Assault Rifle accuracy bonus down (2 -> 1). The accuracy values of each weapon is, for most purposes, the equivalent of the shooter being that many tiles closer to the target when calculating accuracy. The AR before this was registering as the target being two tiles closer, (or being 14% more likely to be hit). Now it's one tile (or 7%).
- Stun rifle accuracy bonus up (1 -> 2). Unlike the AR, the stun rifle only has the default accuracy weight of 5, making this a 5% accuracy increase.
- Stun revolver electrodes travel slightly faster (0.7 -> 0.65). Does not apply to the Stun rifle.
- Added stutter and increased eyeblur to electrode weapons.
- Makes the electrolaser less of a meme and more of a super smartgun without the kill mode. Capacity up (8 -> 10), shock damage up (30 -> 40), stun unchanged.
- The Plasma Pulse Projector has had a facelift. It retains its limited range and weakness to vacuum, but should now actually be worth using. Boasting a decent amount of armor penetration and a higher dizziness amount, as well as a slightly faster projectile, anyone shot by the PPP should now actually feel like they were just shot by a one-of-a-kind non-lethal security weapon. It should also encourage security to actually use non-lethal weapons on higher levels of alert instead of immediately breaking out the shotguns and bullpups. Has also had its ammo capacity doubled (4 -> 8). Can also been worn on the back, now.
- Slight armor pen buff (0 -> 5) to ALL weapons that use smalllaser projectiles (energy guns, smartguns on kill, etc). Midlasers (laser carbines) have 20. Makes zero sense that the smaller version would have zero.
- Most sources of shock from weapons should now cause stuttering.
- Machine pistol 3-round burst fire move_delay reduced (4 -> 2). A 3-round burst from an underweight SMG should not anchor you to the floor as long as an assault rifle.
- Captain's antique laser now fits in holsters.
- Skrell sidearm pistol now uses the same projectile type as the kill setting of smartguns when set to its equivalent mode. (Damage down 5, armor pen up 5)

:cl: OolongCow
balance: changes to a large number of guns.
/:cl:


OolongCow:
Added a sentence to the end of the PPP's description that makes it clear it cannot be used in vacuum.

Roland410:
The stun rifle is fine as-is. You trade size and capacity for more powerful shots. Making it identical to the revolver means there is zero reason to use it anymore.
I'd also buff the PPP's capacity to 10 to make it more used (I was planning to PR that anyways).

OolongCow:
Realized I hadn't actually changed the stun rifle's agony value. Decided leaning into the already-existing differentiation between it and the revolver would likely be a better change. Replaced buff to capacity with a buff to accuracy. Seeing as it's a slow projectile anyway, it really shouldn't also be randomly missing people it should have connected with.

OolongCow:
PPP ammo capacity doubled (4 -> 8).

Sparrowbird611:
I like some of this, some of this I dislike, which I'll list.

Stun rifle change: The thing is gamer, not sure why you buffed it's accuracy more, but that's minimal TBH. I wouldn't touch it still though.

All smartgun related changes: AP on the mini laser I don't see a reason for, and disagree with. Buffing the shock mode I'm also against, and I have to question why you're actually nerfing it's agony damage further.

Agony buffs to all the ballistic non lethals: Unnecessary, they have significantly amount of agony already. I'm pretty sure ballistic weapons are just broken, not underpowered in terms of stats, so changing this does nothing.

Electrolaser: I disagree with buffing it's damage, it's sufficient where it is now.

PPP: I'm not necessarily against this one, but I personally think this gun should just be removed. It's a redundant gun when the stun rifle exists, and even with these changes, I find it hard to believe It'd be viable.

Electrode slowdown: Not a fan of this, current number is sufficient.

Higgin:
step_delay changes how long before you can move after firing - not how long somebody is stopped on a hit.

Consider using distance_falloff if you want to change the range/role of weapons as well.

OolongCow:
> Consider using distance_falloff if you want to change the range/role of weapons as well.

The PPP has a always had a hard limit of 4 tiles.

Sparrowbird611:
Doing some tests, step delay is uh..... literally how fast the bullets move.

I would know, I set it to 100, and I had slow motion bullet memes.

OolongCow:
Testing completed with help from @Sparrowbird611 

Several significant changes made, including a possible fix to buckshot that shouldn't break it. Please reread the original comment after the edits and make sure you stand by your original vote.

Higgin:
"Doing some tests, step delay is uh..... literally how fast the bullets move.

I would know, I set it to 100, and I had slow motion bullet memes."

me unga, got it mixed up with click delay under projectile

OolongCow:
Additional changes made after further testing and feedback.

Sparrowbird611:
I still stand by my disagreement with the smart gun and electrolaser changes, but this looks good otherwise.

OolongCow:
> I still stand by my disagreement with the smart gun and electrolaser changes, but this looks good otherwise.

I actually removed my nerf to smartguns after we did our testing, and after thinking about it some more. Mostly off the back of tator totties potentially having easier access to unsecured ones in #938 's TC reduction. Would kind of suck for them to get the nerfed version.

BurgerLUA:
failing travsi

OolongCow:
> failing travsi

Yeah, it's something totally unrelated to what I changed. I'm going to try and make Travis reattempt.

PR #947, opened 2020-06-08
Improvements to cable coils consistency + Issue #690 Fix

### LINKED ISSUE #690: Cable coil produced by autolathe is normal-sized
#### Description of issue

Cable coil, when made in an autolathe, is a normal-sized item, when cable coil that spawns normally is small/tiny.

#### Difference between expected and actual behavior

For the cable produced by the lathe to be small/tiny.

#### Steps to reproduce

Select cable coil in the autolathe menu, let it be built. Examine it/try to put it in your pockets.

#### Length of time in which bug has been known to occur

Since the autolathe's menu was changed, I'm pretty sure.

#### Client version, Server revision & Game ID

ac76e681649a45e6fcb267c7bb2734837547e310 -  - 2020-04-11

#### Issue bingo

- [x] Issue could be reproduced at least once
- [x] Issue could be reproduced by different players
- [x] Issue could be reproduced in multiple rounds
- [x] Issue happened in a recent (less than 7 days ago) round
- [x] [Couldn't find an existing issue about this](https://github.com/Baystation12/Baystation12/issues)

TheOrleans:
Hello! This is my second PR, sorry for any kind of mistake. (I plan to keep writing this even when I get to my two thousand five hundred fifty fourth PR)

What does this PR do?

**1ST Commit**

- Fixes #690 : Cable coils made in the autolathe are now small too.

- Cable coils are now always shown as "small items", even if it's just 1 piece. This was done because there is a worse issue created by the update of "item size" that allowed complete cable coils to be shown as tiny items. This is consistent with the rest of the stackable objects of the server since they don't change their examine size either, don't know why cable coils would be different. However, the room cable coils take inside the backpack keeps changing depending on the amount of cable, just like it was before, no actual gameplay change

- Fixed another issue: cable coils didn't have singular_name in their definition, meaning that, when examined, cable coils would show: "There is/are X s in the stack", instead of "There is/are X cable piece/s in the stack".

- Cable coils that have 2 units in the stack are now called "long cable piece". 1 unit cable coil is still called "cable piece", and over 2 units is still called "multipurpose cable coil". The reason behind this is that, when examined, multipurpose cable coils show (now that it's fixed, as explained in the previous point) "There are X cable pieces in the stack"

**2ND Commit**

- Everything present in 1st commit

- Cable coils, when examined, showed 2 descriptions, and showed the amount of cable twice. I don't know why cable coils were so different from the rest of the objects 🤷‍♂️ It's fixed now, when examined you will only see the amount of cable once.

**3RD Commit**

- Everything present in 1st and 2nd commit

- cable.dm code optimization. Removed pieces of code that were no longer useful or needed after the changes of the two previous commits. Special thanks to @Roland410 for helping with this

And that's it 😄


code/modules/power/cable.dm
```diff
@@ -563,8 +565,8 @@ obj/structure/cable/proc/cableColor(var/colorC)
 	to_chat(user, "<span class='notice'>You change \the [src]'s color to [lowertext(selected_color)].</span>")
 
 /obj/item/stack/cable_coil/proc/update_wclass()
-	if(amount == 1)
-		w_class = ITEM_SIZE_TINY
+	if(amount <= 2)
+		w_class = ITEM_SIZE_SMALL
```
Roland410, line 569: I think you wantes to do ITEM_SIZE_TINY here? Otherwise this piece of code makes no sense.

TheOrleans: Actually, I removed TINY to write SMALL because that's what breaks the coil

I know that lines make no sense, but I am too new in BYOND coding to be brave and delete part of the code lmfao

Roland410: You never learn if you don't.

TheOrleans: Still, I kind of want to approach this slowly. It looks like it should be safe to remove the whole /obj/item/stack/cable_coil/proc/update_wclass() part, but maybe that's related with the update of the room the item takes in backpacks... I have a lot to learn 

I know that piece of code looks "stupid" since it's unnecessary, but it's existence is not harmful in any way so... 🤷‍♂️

Roland410: You've already fucked that part where the size gets updated when you changed that. That's my exact point. This makes the size small if there is less or equal of 2, and in every other case it ALSO makes it small.

TheOrleans: As shown in Discord, I didn't fuck that part. Roland come here and repent from the great crime of questioning the greatest coder ever

jk jk

TheOrleans:
Yay

PR #952, opened 2020-06-08
Ups Colony and SDTF spawn rate

Rav-Nunya:
So, this PR changes the cost of the skrell ship and colony to .2, this was done for, A. i have gotten feedback that people enjoy colony so i think we should try it more often, and B. i play SDTF a fair bit and want to do it more, this can be said to be selfish as i havent heard word from others about it.

### DIFF
--- a/maps/away/skrellscoutship/skrellscoutship.dm
+++ b/maps/away/skrellscoutship/skrellscoutship.dm
@@ -12,7 +12,7 @@
 	id = "awaysite_skrell_scout"
 	description = "A Skrellian SDTF scouting vessel."
 	suffixes = list("skrellscoutship/skrellscoutship-1.dmm", "skrellscoutship/skrellscoutship-2.dmm")
-	cost = 0.5
+	cost = 0.2
 	shuttles_to_initialise = list(/datum/shuttle/autodock/overmap/skrellscoutship, /datum/shuttle/autodock/overmap/skrellscoutshuttle)
 	apc_test_exempt_areas = list(
 		/area/ship/skrellscoutshuttle =                NO_SCRUBBER,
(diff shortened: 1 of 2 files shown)


Roland410:
The SDTF spawns often enough in my experience, and what limits the colony the most is the planet type that gets spawned. Not all planets can spawn the colony.
If you tested this and think the colony spawns more often with it, good, but I'd keep the SDTF spawn as-is, or it'll be an almost guaranteed spawn like the Vox are currently.

TheOrleans:
I agree with Roland here, colony spawns quite often I would say. Do you know if increasing the colony spawn rate reduces the spawn rate of others (is there a limit in how many "away sites" and "special sites" can be spawned per round?)

Roland410:
> I agree with Roland here, colony spawns quite often I would say. Do you know if increasing the colony spawn rate reduces the spawn rate of others (is there a limit in how many "away sites" and "special sites" can be spawned per round?)

Technically as there is a budget, increasing the chance (lowering their cost) to spawn means it is more likely to spawn compared to other ruins.

Tename:
Colony upping is good but the SDTF ship is boring and already spawns enough

Roland410:
The thing is, as I said above, the colony spawn is restricted not just by the cost, but by the planet type, it's why you're not seeing it as often as you would want it.

BurgerLUA:
Needs more feedback.

BurgerLUA:
People don't seem to like this PR. Find more people that like it or adjust the PR so that people are happy.

BurgerLUA:
Might close this PR if there isn't a convincing argument as to why this is needed.

Roland410:
Going to close this. If you want it reopened ask a maintainer, but as I said above, it won't solve the spawn rates.

PR #953, opened 2020-06-08
The amount of shots negated by nanites are decreased to 1 from 4

Jackofashepard:
:cl:
tweak - The amount of shots absorbed by nanites are reduced to 1 from 4
tweak - Recharge time reduced to 2 minutes from 5
/:cl:

### DIFF
--- a/code/modules/augment/passive/nanoaura.dm
+++ b/code/modules/augment/passive/nanoaura.dm
@@ -15,8 +15,8 @@
 	desc = "Nanomachines, son."
 	action_button_name = "Toggle Nanomachines"
 	var/obj/aura/nanoaura/aura = null
-	var/charges = 4
-	var/max_charges = 4
+	var/charges = 1
+	var/max_charges = 1
 	var/list/memesounds = list('sound/effects/nanomachines/nanomachinesson.ogg', 'sound/effects/nanomachines/physicaltrauma.ogg')
 	var/next_regen_time
 	var/regen_delay = 5 MINUTES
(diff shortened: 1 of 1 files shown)


Jackofashepard:
This augment can be rushed in less than ten minutes and mass produced with relative ease. For this small amount of work you can completely negate any damage taken from 4 projectiles. Imagine if you shot a dude with a AMR 4 times and he just walks it off as if nothing happened.

Roland410:
You have done nothing about it's excruciatingly long recharge time, and it doesn't change the fact that it blocks 4 HITS, not X amount of damage. If you burst fire someone with an SMG or just manage to hit them 4 times even with a pistol, it means the augment is now useless until it recharges, which takes FIVE MINUTES for ONE charge.
You can only use this basically every FIVE minutes per shot negated, which means for it to fully recharge you currently have to wait TWENTY minutes.

Jackofashepard:
Reduced the charge time from 5 to 2 minutes to compensate for the change

TheOrleans:
I think you should be changing the DAMAGE stopped instead of the HITS it can stop. One AMR shot is not a 9mm pistol shot. Will definitely support this PR if that's changed

Sparrowbird611:
Not a fan of these nanites. Orleans idea is good, but unless some one is actually putting in the time to do that, I'm going with this PR.

BurgerLUA:
Needs more feedback.

BurgerLUA:
People don't seem to like this PR. Find people who do or adjust the changes.

BurgerLUA:
Closing this PR as another PR just outright removes them.

PR #956, opened 2020-06-09
Regeneration and minor balance changes.

Roland410:
🆑Roland410
tweak:All species have the ability to regenerate damage.
rscadd:Being hit, shot etc. disables your regeneration for 30 seconds.
rscadd:Regeneration now heals clone damage as well.
tweak:Space does five times more oxygen damage now.
tweak:Lungs pop at 30kPa pressure difference now.
tweak:IPCs and FBPs' cell can take four times more damage now before breaking.
/🆑
The side-effect of giving all species regeneration is that their organs actually heal at 0.25 damage per second as well, not just their external damage, and I had to implement the organheal_nut variable as that part of the code means they heal EVEN MORE organ damage if they are above 380 out of 400. Dionaea and prometheans enjoy the advantage of that being basically always active considering their nature.

### DIFF
--- a/code/game/objects/auras/regenerating_aura.dm
+++ b/code/game/objects/auras/regenerating_aura.dm
@@ -3,24 +3,28 @@
 	var/brute_mult = 1
 	var/fire_mult = 1
 	var/tox_mult = 1
-	var/oxy_mult = 0
+	var/oxy_mult = 1
+	var/clone_mult = 1
 
 /obj/aura/regenerating/life_tick()
 	user.adjustBruteLoss(-brute_mult)
 	user.adjustFireLoss(-fire_mult)
 	user.adjustToxLoss(-tox_mult)
+	user.adjustOxyLoss(-oxy_mult)
+	user.adjustCloneLoss(-clone_mult)
 
 /obj/aura/regenerating/human
 	var/nutrition_damage_mult = 1 //How much nutrition it takes to heal regular damage
 	var/external_nutrition_mult = 50 // How much nutrition it takes to regrow a limb
 	var/organ_mult = 2
+	var/organheal_nut = 380 //The minimum amount of nutriments stored when organs can heal themselves.
 	var/regen_message = "<span class='warning'>Your body throbs as you feel your ORGAN regenerate.</span>"
 	var/grow_chance = 0
 	var/grow_threshold = 0
-	var/ignore_tag//organ tag to ignore
+	var/ignore_tag //organ tag to ignore
 	var/last_nutrition_warning = 0
 	var/innate_heal = TRUE // Whether the aura is on, basically.
-
+	var/noregen = 0 //Used to count how long to toggle regen when hit.
 
 /obj/aura/regenerating/human/proc/external_regeneration_effect(var/obj/item/organ/external/O, var/mob/living/carbon/human/H)
 	return
@@ -30,11 +34,16 @@
 	if(!istype(H))
 		CRASH("Someone gave [user.type] a [src.type] aura. This is invalid.")
 		return 0
+	if (world.time < noregen)
+		innate_heal = FALSE
+	else
+		innate_heal = TRUE
 	if(!innate_heal || H.InStasis() || H.stat == DEAD)
 		return 0
 	if(H.nutrition < nutrition_damage_mult)
 		low_nut_warning()
 		return 0
+
 
 	if(brute_mult && H.getBruteLoss())
 		H.adjustBruteLoss(-brute_mult * config.organ_regeneration_multiplier)
@@ -46,7 +55,10 @@
 		H.adjustToxLoss(-tox_mult * config.organ_regeneration_multiplier)
 		H.adjust_nutrition(-nutrition_damage_mult)
 	if(oxy_mult && H.getOxyLoss())
-		H.adjustToxLoss(-oxy_mult * config.organ_regeneration_multiplier)
+		H.adjustOxyLoss(-oxy_mult * config.organ_regeneration_multiplier)
+		H.adjust_nutrition(-nutrition_damage_mult)
+	if(clone_mult && H.getCloneLoss())
+		H.adjustCloneLoss(-clone_mult * config.organ_regeneration_multiplier)
 		H.adjust_nutrition(-nutrition_damage_mult)
 
 	if(!can_regenerate_organs())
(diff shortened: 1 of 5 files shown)


Sparrowbird611:
Interesting changes, a few things I'd like to clear up though.

You've already mentioned this hits unathi regen as well, so I don't need to ask that. I assume your reasoning behind this was to make their regen something to use to recover after the fight, which seems fine to me.

I have to ask the rational behind making the lung pop threshold higher. It's one of the more obnoxious things to have happen, and I don't know the extent this PR pushes your ability to heal it on your own.

Not against buffing the IPC core, but I'd like to ask the rational behind that.

Aside from these questions, this is nice. I can't judge if it's wise to make space deal more suffocation damage, but I can't see it ending too badly, and being able to slowly heal clone damage on your own seems fine.

DriedCrust:
Hard no on space being deadlier and lungs popping by even more gentle depressurization. Enough people already complain as is on lungs being extremely frail and popping due to silly things. Being exposed to vaccum is already extremely deadly since the brute damage will quickly cripple you due to pain. Everything else is fine.

Roland410:
> Interesting changes, a few things I'd like to clear up though.
> 
> You've already mentioned this hits unathi regen as well, so I don't need to ask that. I assume your reasoning behind this was to make their regen something to use to recover after the fight, which seems fine to me.
> 
> I have to ask the rational behind making the lung pop threshold higher. It's one of the more obnoxious things to have happen, and I don't know the extent this PR pushes your ability to heal it on your own.
> 
> Not against buffing the IPC core, but I'd like to ask the rational behind that.
> 
> Aside from these questions, this is nice. I can't judge if it's wise to make space deal more suffocation damage, but I can't see it ending too badly, and being able to slowly heal clone damage on your own seems fine.

The lung pop threshold is actually lowered, now you only need a 30kPa difference for it to pop, but having a suit on even without internals negates this (although you don't have infinite oxygen, so don't try going on for long). It's also kind of necessary considering the rest of the changes.

With these changes, organs (on local) heal 0.25 damage per second, and a shitton more if you're basically full on food (380+ out of 400), I've had to make some changes to balance it.

IPC battery buff is so Decker can stop crying, and because it was a joke anyways with the tiny amount of 5.

Roland410:
> Hard no on space being deadlier and lungs popping by even more gentle depressurization. Enough people already complain as is on lungs being extremely frail and popping due to silly things. Being exposed to vaccum is already extremely deadly since the brute damage will quickly cripple you due to pain. Everything else is fine.

You heal brute, burn, oxygen, toxin and clone and organ damage at 0.25 per second with these changes. Had to make some changes for it to balance itself out.

DriedCrust:
> > Hard no on space being deadlier and lungs popping by even more gentle depressurization. Enough people already complain as is on lungs being extremely frail and popping due to silly things. Being exposed to vaccum is already extremely deadly since the brute damage will quickly cripple you due to pain. Everything else is fine.
> 
> You heal brute, burn, oxygen, toxin and clone and organ damage at 0.25 per second with these changes. Had to make some changes for it to balance itself out.

Just saying, lungs are REALLY REALLY frail in brainmed and crumble at the nearest gentle breeze. But fair point, tbh.

Roland410:
> > > Hard no on space being deadlier and lungs popping by even more gentle depressurization. Enough people already complain as is on lungs being extremely frail and popping due to silly things. Being exposed to vaccum is already extremely deadly since the brute damage will quickly cripple you due to pain. Everything else is fine.
> > 
> > 
> > You heal brute, burn, oxygen, toxin and clone and organ damage at 0.25 per second with these changes. Had to make some changes for it to balance itself out.
> 
> Just saying, lungs are REALLY REALLY frail in brainmed and crumble at the nearest gentle breeze. But fair point, tbh.

The thing is, if you also stuff yourself full of nutrients (not even counting actual meds yet), you heal about 0.75 damage per second which makes medbay almost irrelevant as you can just rob the snack vendor.

OolongCow:
This sounds like a great way to completely invalidate medical even further. Nothing but organ damage even matters as it is. And with this change, it would literally only take two and a half minutes to regenerate from being shot.

That value needs to be way, way lower.

Roland410:
> This sounds like a great way to completely invalidate medical even further. Nothing but organ damage even matters as it is. And with this change, it would literally only take two and a half minutes to regenerate from being shot.
> 
> That value needs to be way, way lower.

I'm not sure how much the average organ damage is tbf, but it heals 0.25 per second, and if I knew why the aura thing heals organ damage when it should be healing brute damage, I would change it. I suppose it has something to do with organ damage factoring into overall brute damage and it heals it because of that.

reklaw9:
in feel like this will go terribly wrong,and this will probably make medical more useless

reklaw9:
15 units of organ damage per minute,6-7 minutes are literally 100 units of organ damage healed

AdamElTablawy:
This is a bandaid fix to brainmed's issues by making damage matter even less.

If everything heals over time, then any damage that doesn't drop your blood pressure or kill your liver/kidneys is going to just be ignored-and it means that people will walk off shit pretty fast.

Healing .25 organ damage every second is going to make post-combat way less deadly, and not by actually resolving the inherent issues of 'everything damages organs, and organs are the only way to be injured that actually really matters', but instead by enabling people to just walk off shit they shouldn't while not actually being -hurt- any less.

Roland410:
Tbf this was never meant to be a bandaid fix to Brainmed or anything, and I think I remember how to fix the organ damage healing, although I'm not sure if it's a change in a good direction, but that way the original intent will be actual.

Superhats:
I honestly don't think giving **every** species such regen is a good idea. Natural regen is one of the main things that differentiates species like unathi, prometheans and diona from normal humans. 

This is gonna make it so you can literally **sit** any type of damage out by keeping yourself on painkillers, dexalin and potentially inaprovaline. I'm just curious as to  **why** that would be a good thing.

Roland410:
The regen uses nutrients up, and was meant to be for brute/burn/clone damage. It takes 4 seconds approx to heal 1 point of damage, which takes like 4 or 8 nutrients iirc. The organ damage thing was a side-effect, but I'm 90% sure I know where I messed up.
The intent was to give a small natural regen to organics, because right now even minor wounds (<10) can take ages to heal without chemicals.

OolongCow:
> The regen uses nutrients up, and was meant to be for brute/burn/clone damage. It takes 4 seconds approx to heal 1 point of damage, which takes like 4 or 8 nutrients iirc. The organ damage thing was a side-effect, but I'm 90% sure I know where I messed up.
> The intent was to give a small natural regen to organics, because right now even minor wounds (<10) can take ages to heal without chemicals.

I honestly think a better, safer, and more interesting option would be adding Tricordrazine to every nanomed. 5u pills diluted with something to stop you from just guzzling them down?

reklaw9:
humans are regenerating their wounds while being operated on,unathis,i understand,but normal humans...

ProbablyCarl:
no thanks

PR #958, opened 2020-06-09
The Bulldog can load non-frangible mags again.

DriedCrust:
It's currently a worse SMG since the frangibles do nothing the moment the tiniest bit of armor pokes them. Also, the security Bulldog is a way worse variant of the genuine bullpup rifle, it has: jam chance, no burst fire, needs commandauthorization/red alert to fire & it has overall worse stats when compared to the bullpup. This should, finally make the weapon something that isn't taken from the SEC-ARM just because it looks cool.


OolongCow:
The bullpup's strength comes from the fact that it has a 2x modifier to cause internal bleeding, which is likely being nerfed with #941 . I'd advise waiting for that to be merged so we can see where they stand afterwards.

DriedCrust:
Further testing on the thunderdome has revealed that with the merc voidsuit and heavy armor, the frangible rounds end up breaking your chest first (in atleast 9 to 10 rounds) before any arterial bleeding or organ damage is applied, this is due to the low damage and AP getting negated by the armor and thus, the modifier never comes into effect.

OolongCow:
/obj/item/projectile/bullet/rifle/military
	fire_sound = 'sound/weapons/gunshot/gunshot2.ogg'
	damage = 35
	armor_penetration = 35
	shrapnel_chance_multiplier = 1
	arterial_bleed_chance_multiplier = 1

Is what would be replacing

/obj/item/projectile/bullet/rifle/military
	fire_sound = 'sound/weapons/gunshot/gunshot2.ogg'
	damage = 50
	armor_penetration = 5
	shrapnel_chance_multiplier = 2 (soon to be 1.5)
	arterial_bleed_chance_multiplier = 2 (soon to be 1.5)

For those who don't know.

PR #960, opened 2020-06-10
Nerfs the AoE effect of the Ion Rifle and the Ion Pistol.

Dark1Star3s:
* Changed: Ion Rifle - Heavy AoE Effect from 1 to 0. - Light AoE Effect from 2 to 1.
* Changed: Ion Pistol - Removes AoE Effect

Ion Weapons do not even have to hit the target to kill synthetics, this changes it so the Ion Pistol has no AoE Effect, and requires direct hits just like any other pistol. The Ion Rifle has a light AoE Effect of 1. Which would be better used against mechs and such. However it is still deadly to synthetics. Also gives them more of a chance to dodge Ion weapons.


reklaw9:
this is a salt PR,however,i understand that its a one click weapon and that it needs to be nerfed

Dark1Star3s:
You can call it a Salt PR all you want, but I can garuntee you that, I've never played Synth, I've only played against them. It's no fun being ablt to 1 hit KO any Synth.

Sparrowbird611:
I see no downside to this. You are punished for missing with other weapons, I see no reason for this not to apply to the Ion weapons, given they can disable IPCs, energy based weapons, and blind people with thermals. The Rifle still possesses It's AoE, which seems acceptable, but the pistol frankly did not seem to need it.

You will have to present better arguments than "salt PR," as if you try hard enough, you may apply this to most PRs, given things usually aren't noticed to be OP until you either die to them or use them to kill others.

Roland410:
Jokes aside, I wouldn't care as much, if this PR fixed stuff such as the absolutely joke of a fire rate, stupidly slow projectile travel speed, or its capacity (which especially for the pistol is pretty bad iirc).
There are also no demonstrations or anything that shows how effective the weapons will be after the nerf, which is what discourages me the most about nerfing them.

Dark1Star3s:
It's hard to demonstrate an Ion Pistol and Rifle because it usually means
you have to have someone else test it with you, however the values are
changed and such, with the pistol you have to hit them directly, the Ion
Rifle you have to hit them, or the tile next to them. Although hitting them
directly will do the most damage. (Obviously) The capacity can be looked
into, the fire rate? Sure that can to, I do agree that they fire way too
slow. So I might bring the pistol in-line with the mil pistol fire-rate,
and perhaps increase the capacity to 6.

The bonus at this point would be the rifle fires slow, but has a high
capacity and has an AoE. The Pistol fires faster, lower capacity and
requires direct hits. (Something you would use in a close quarters
environment, the rifle would be used at range etc.)

OolongCow:
This would probably be better served by nerfing the effect it has on synths (to not be an instant knockout) rather than directly nerfing Ions. It's definitely something that needs to be HEAVILY scrutinized, though. Synths are already in a very weird spot when it comes to combat. They're both over and _under_powered at the same time. Large changes to the ONLY direct counter they have, which is also one of the few things keeping several other items in check, is bound to cause waves.

BurgerLUA:
Needs more feedback. Maybe even a poll.

DerFlammenwerfer:
> This would probably be better served by nerfing the effect it has on synths (to not be an instant knockout) rather than directly nerfing Ions. It's definitely something that needs to be HEAVILY scrutinized, though. Synths are already in a very weird spot when it comes to combat. They're both over and underpowered at the same time. Large changes to the ONLY direct counter they have, which is also one of the few things keeping several other items in check, is bound to cause waves.

Technically speaking Ion rifles are not the only method of generating an EMP, nor even the easiest to obtain. With a bit of mining done, it's very easy to make a chem grenade that serves the same purpose over a potentially far greater area.

Superhats:
Closing it due to lack of support from the community and staff.

PR #961, opened 2020-06-11
Makes Frag-Grenades deadly.

Dark1Star3s:
* Changed: Normal Fragment Damage from 7 to 30
* Changed: Strong Fragment Damage from 15 to 50

Frag-Grenades are now in-line with other weaponry of their time, such as EMP grenades. Wearing armor will save you from a grenade, but if you're not, you are going to die. But it WILL hurt.
Basically.

In short.

Run.

https://youtu.be/9ZmhTb_l3eA video of them in action.


Roland410:
Based and redpilled. Good PR.

PR #962, opened 2020-06-11
You will no longer be floored from any agony damage to genitals 

Jackofashepard:
In short the check checked if you had genitals, if the awnser is yes then you get floored from any amount of agony. That has been removed.


Sparrowbird611:
Wha….. is this really how it worked? I thought it was an agony modifier some where, similar to the head agony modifier that I know exists in some form.

Roland410:
There goes all the dick shots. Good PR, but it will be missed.

TheOrleans:
Why? It's a really cool feature. Downvoting with all my heart.

Abductkaz:
If a bonus to groinshots has to stay then maybe just make it a bonus instead of a flat out 100% knockdown? 1.5x or 2x or something.

Personally I don't really care either way, but I think an instant knockdown is dumb.

Roland410:
It is a 100% bonus. It only floors with stun rifles and such which have high agony values.

PR #963, opened 2020-06-11
CoS Hazard Hardsuit, Armor Buff.

Dark1Star3s:
* Changed: Projectile protection from, ARMOR_BALLISTIC_PISTOL to ARMOR_BALLISTIC_RESISTANT
* Changed: Laser protection from, ARMOR_LASER_HANDGUNS to ARMOR_LASER_MAJOR
* Changed: Melee Protection from, ARMOR_MELEE_MAJOR to ARMOR_MELEE_VERY_HIGH
* Changed: Radiation Protection from, ARMOR_RAD_MINOR to ARMOR_RAD_SHIELDED

Feedback may or may not be required. It's only a slight buff. However this will allow the CoS to be somewhat capable in the Hardsuit, instead of opting for actual armor which is better off.

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/station.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/station.dm
@@ -321,8 +321,8 @@
 	icon_state = "hazard_rig"
 	armor = list(
 		melee = ARMOR_MELEE_MAJOR,
-		bullet = ARMOR_BALLISTIC_RIFLE,
-		laser = ARMOR_LASER_MAJOR,
+		bullet = ARMOR_BALLISTIC_RESISTANT,
+		laser = ARMOR_LASER_HANDGUNS,
 		energy = ARMOR_ENERGY_MINOR,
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
I recommend these stats instead, due to recent buffs to the security voidsuits.

		melee = ARMOR_MELEE_VERY_HIGH,
		bullet = ARMOR_BALLISTIC_RESISTANT,
		laser = ARMOR_LASER_MAJOR,
		bomb = ARMOR_BOMB_RESISTANT,
		bio = ARMOR_BIO_SHIELDED,
		rad = ARMOR_RAD_SHIELDED)

While I do agree in it's current state, the suit is kinda shitty in terms of armor, it has a number of other things going for it, including slightly lowered slowdown, built in weaponry and tools, and the ability to be undeployed and redeployed at will. With it's biggest disadvantage, the inability to use a backpack gone, I'd suggest against raising it's stats higher than this.

PR #967, opened 2020-06-13
CQC and Sniper gods no more

Jackofashepard:
- Weapons skill accuracy bonus for master rank is now +1 instead of the +2 ontop of the +2 already obtained from expert+trained.

- Master weapons skill now provides a 40% to fail auto-unjam isntead of 0%

- Expert has a 50% fail chance, 63% for trained, 79% for basic and 100% for none

- Close Combat has been changed so that a expert can still resist out of a masters blue grab and so on

Input is welcome and salt comments will be ignored. Civil discussion is welcome

### DIFF
--- a/code/modules/mob/grab/grab_datum.dm
+++ b/code/modules/mob/grab/grab_datum.dm
@@ -286,7 +286,7 @@
 	if(affecting.confused)
 		break_strength--
 
-	if(break_strength < 1)
+	if(break_strength < 2)
 		to_chat(G.affecting, "<span class='warning'>You try to break free but feel that unless something changes, you'll never escape!</span>")
 		return
 
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1709,8 +1709,6 @@
 		. += 1
 	if(skill_check(SKILL_WEAPONS, SKILL_EXPERT))
 		. += 1
-	if(skill_check(SKILL_WEAPONS, SKILL_PROF))
-		. += 2
 
 /mob/living/carbon/human/can_drown()
 	if(!internal && (!istype(wear_mask) || !wear_mask.filters_water()))
--- a/code/modules/mob/living/carbon/human/human.dm
+++ b/code/modules/mob/living/carbon/human/human.dm
@@ -1709,6 +1709,8 @@
 		. += 1
 	if(skill_check(SKILL_WEAPONS, SKILL_EXPERT))
 		. += 1
+	if(skill_check(SKILL_WEAPONS, SKILL_PROF))
+		. += 1
 
 /mob/living/carbon/human/can_drown()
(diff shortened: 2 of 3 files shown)


Sparrowbird611:
Not entirely sure on the accuracy reductions, but the jam chance and grab parts are welcome.

SouthyOsmanli:
While I’m fine with the jam change I’m heavily against the accuracy change. If anything remove RNG inaccuracy all together and make the skills count towards something else.

The gunplay with RNG missing is terrible feeling as is which is why people resort to taking mastery in the first place.

Dark1Star3s:
I'm going to be honest, I'm not sure how I feel about this, considering that Master Gun Skill is what actually makes some guns worth it to use. For example, the Z8B Bulldog that security gets, has a very high Jam chance, and fires very slow.

If you remove the ability on Master for it to clear Jam 100% of the time. Sorry but, that Bulldog is now useless and I'm just going to pick up a Combat Shotty instead. Or a Carbine, or literally ANY other weapon. Obviously, that's a flaw in the bulldog its self.

Jackofashepard:
If you are downvoting this because you need to deal with miss chances then please suggest a better system or code a better one. The fact is that all our weapons are balanced with accuracy modifiers, jam chances, etc. Having a skill that negates balance variables means we might as well remove them and have all guns be garunteed hits. Also thats a issue with the guns and not the skills.

Roland410:
The only thing I find good about this PR is the grab escape thing. If you remove the accuracy modifier and halve the unjam chance (although very few weapons jam), it will be pointless to take.

AdamElTablawy:
@Jackofashepard 
> Having a skill that negates balance variables means we might as well remove them and have all guns be garunteed hits. 

Yes, we might as well.

Jackofashepard:
Alright making chances based on feedback

Jackofashepard:
Changes made

Jackofashepard:
And people, really, you need to remember this is a 2D spessman game where if you click in the general direction of the pixels your pew pew pixels will hit, thats why miss and jam exists. if point and click is what you want then the game will devolve into who has the better armour and who shoots first. This also has absolutely nothing to do with this PR unless you are purely using Master weapons for no misses and no jams.

TheOrleans:
Close combat change? Totally fine with it

Unjam change? Wrong direction. Make it 1% fail for Master, 20-30% for Exp, 50-60% for Trained, 80% for Basic, 0% for none

Accuracy change? Hm... I think it's good

Sparrowbird611:
Orlean, you're aware that only a few very specific guns can actually jam, right? These are the bullpups/dogs and the tectonic. Nothing else in the game jams, and this PR isn't changing that, it's just making it so those guns that do jam actually work as intended.

BurgerLUA:
Needs more feedback.

BurgerLUA:
Might close this PR unless people like it.

Roland410:
Closing this as the feedback is pretty negative. If you want it reopened ask a maintainer.

PR #969, opened 2020-06-13
Removes cloaking device from research.

Sparrowbird611:
*Breathes in*

~SPEEEEEEEEEEED MEEEEEERGE~

Ok but jokes aside, this thing probably shouldn't be in RnD in it's current state. Total invisibility without a way to be seen through would be fine if it was the rare cloaking suit, or ninja, but being able to print this thing is a step too far.

### DIFF
--- a/code/modules/research/designs/designs_rig_modules.dm
+++ b/code/modules/research/designs/designs_rig_modules.dm
@@ -166,15 +166,6 @@
 	build_path = /obj/item/rig_module/fabricator/energy_net
 	sort_string = "WCKAC"
 
-/datum/design/item/rig/stealth
-	name = "Active Camouflage"
-	desc = "An integrated active camouflage system, mountable on a RIG."
-	id = "rig_stealth"
-	req_tech = list(TECH_MATERIAL = 5, TECH_POWER = 6, TECH_MAGNET = 6, TECH_ESOTERIC = 6, TECH_ENGINEERING = 7)
-	materials = list(MATERIAL_STEEL = 6000, MATERIAL_GLASS = 3000, MATERIAL_DIAMOND = 2000, MATERIAL_SILVER = 2000, MATERIAL_URANIUM = 2000, MATERIAL_GOLD = 2000, MATERIAL_PLASTIC = 2000)
-	build_path = /obj/item/rig_module/stealth_field
-	sort_string = "WCLAA"
-
 /datum/design/item/rig/cooling_unit
 	name = "Cooling Unit"
 	desc = "A suit cooling unit, mountable on a RIG."


JohnBigless:
**SPEEEEEEED MEEEEEEEERGE**

Roland410:
Can we for one day not remove everything from research that is remotely fun or useful? They can only use this themselves with the shitty zero rig anyways from robotics and even that means you need a roboticist. The module also doesn't make sounds silent, and afaik it decloaks you if you try to attack. I honestly don't see any reason for its removal.

OolongCow:
> Can we for one day not remove everything from research that is remotely fun or useful? They can only use this themselves with the shitty zero rig anyways from robotics and even that means you need a roboticist. The module also doesn't make sounds silent, and afaik it decloaks you if you try to attack. I honestly don't see any reason for its removal.

Paralytic pens don't count as attacks for the purpose of decloaking.

Roland410:
I mean, true, but to parapen someone like that, you already need a rig, research to do their job (unless antag scientist), and it won't even let you into areas you couldn't access unless you sneak in, but you essentially block your way out then.
I'd say using this to parapen someone is pretty extreme and should not be grounds to removing it.

Dark1Star3s:
> I mean, true, but to parapen someone like that, you already need a rig, research to do their job (unless antag scientist), and it won't even let you into areas you couldn't access unless you sneak in, but you essentially block your way out then.
> I'd say using this to parapen someone is pretty extreme and should not be grounds to removing it.

In my honest opinion, it should be kept, but make it so they cannot activate the cloaking module unless the RIG is fully deployed, because as it stands, people just use it as a module holder, while they then put actual armor over it and then go invisible. EG: The many times Richard has got his hands on one.

Sparrowbird611:
>can we not remove fun things?

Look who you're talking to.

Jokes aside, I've never seen any of this specific behavior on the cloaking module. As far as I know, you can attach it to any rig, and it will not decloak you as far as I know when firing, though I do recall gun sprites being visible when held and invisible so, I suppose that's the same thing, though rig mounted weaponry can bypass that somewhat if I recall correctly.

I'm perfectly fine with the way it works currently, including the armor memes, but it seems rather strong to be in research with the way RnD works now.

Sparrowbird611:
I'm willing to reconsider this PR, but cloaking in general is an incredibly powerful ability, even in the hands of the unrobust, let alone gamers like plexiant. An Ion rifle is a potential counter to it, but given the fact that soon only one of the guns with have an AoE, and that one is in the EARM, I hear potential issues.

~~I also have nightmares of the day security some how manages to get fully stealth kitted soldiers.~~

Sparrowbird611:
Yeah uh, if burger or any one comes by and looks at this, don't merge this please. Looking into alternative solutions for now.

OolongCow:
> In my honest opinion, it should be kept, but make it so they cannot activate the cloaking module unless the RIG is fully deployed, because as it stands, people just use it as a module holder, while they then put actual armor over it and then go invisible.

This, pretty much.

Sparrowbird611:
I'm going to close this for the time being. There's been another suggestion I'm pursuing. If that fails, I'll reopen this PR.

PR #970, opened 2020-06-14
Hornet Balance Pass and Name Change + Fixes Tan Tactical Crate

Dark1Star3s:
* Changed Name - Name of Hornet no longer "Z2 Hornet" and is now "hornet marksman rifle".
* Changed Description - added "Z2 Hornet" in to it.
* Changed Sound - Reload sound of the Hornet modified to make more sense, and easier on the ears.
* Changed Fire Deleay - Hornet now shoots every 3 seconds instead of 4. (It's Semi Automatic.)

I feel like this change was needed after testing it in an actual round, the sprite will be changed too at some point.

-----------------------------------

Decided to fix the Tan Tactical Crate in the same PR.
* Added - Arm/Leg Guards to the Tan Tactical Gear Crate.

### DIFF
--- a/code/modules/clothing/suits/armor.dm
+++ b/code/modules/clothing/suits/armor.dm
@@ -315,7 +315,7 @@
 
 /obj/item/clothing/suit/armor/pcarrier/tan/tactical
 	name = "tactical plate carrier"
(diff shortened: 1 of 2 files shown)


Sparrowbird611:
Haven't used the gun ingame yet, but I'm sure you know what you're doing with this change.

PR #971, opened 2020-06-14
Nanites can no longer be produced 

Jackofashepard:
Title. Honestly was hoping my other PR would be preferable but here it is.

### DIFF
--- a/code/modules/research/designs/designs_mechfab.dm
+++ b/code/modules/research/designs/designs_mechfab.dm
@@ -520,13 +520,6 @@
 	req_tech = list(TECH_MATERIAL = 5, TECH_COMBAT = 4, TECH_BIO = 4)
 	id = "augment_armor"
 
-/datum/design/item/mechfab/augment/nanounit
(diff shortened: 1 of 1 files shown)


SouthyOsmanli:
Yeah, sadly nanites are incredibly overpowered currently and need a heavy rework done on them. Disabling them from production until someone wishes to do that said heavy rework is prefable.

Roland410:
Literally use big brain and shoot a burst at it once. The charge time is 5 minutes for 1 shot blocked.

Sparrowbird611:
While that is accurate roland, it still basically cucks any one not using an AR. Stealth traitor trying to run away by disabling you? Nanomachines son. Nanites essentially ruin everything short of an auto pistol, e sword, or AR, so I can't support their existence in good faith.

BurgerLUA:
I'll give this another week for more feedback.

BurgerLUA:
Okay I'm being told that these should be removed with the balance concerns. Hopefully they can be re-added in the future with a nerf.

PR #973, opened 2020-06-16
Nerfs the E-Swords Damage a tiny bit.

Dark1Star3s:
Not entirely sure how much this will effect the sword, but it's a tiny nerf. Hopefully I won't have to do another one. >_>

* Changed: Damage from 45 to 30.

### DIFF
--- a/code/game/objects/items/weapons/melee/energy.dm
+++ b/code/game/objects/items/weapons/melee/energy.dm
@@ -121,7 +121,7 @@
 	name = "energy sword"
 	desc = "May the force be within you."
 	icon_state = "sword0"
-	active_force = 45
+	active_force = 30
 	active_throwforce = 20
 	force = 3
 	throwforce = 5


TheOrleans:
No to salt prs.

Sparrowbird611:
*breathes in*

~~SALT PR REEEEE~~

Just kidding.

Jokes aside however, I'd like to point out two things. First off, I'd suggest testing sheep's new changes to traitor item prices, including stim packs and e sword. Secondly, if you are intent on this, I would not advise keeping it at the current 58 TC price. 36-42 would seem more appropriate.

Roland410:
I'm downvoting as well because you've provided no proof this changes anything about it in real situations (or even in testing), so in short, test it at the very least, and adjust it then. What Sparrow said as well, because if it does less damage, it should not cost as much either.

BurgerLUA:
Needs more feedback.

BurgerLUA:
Probably needs a poll.

PR #983, opened 2020-06-18
Blacklists adjudicator, bailiff, infantry and cyborgs from antags

Roland410:
🆑Roland410
tweak:Blacklists the Adjudicator, Bailiff, and infantry from all antags except changeling.
tweak:Blacklists cyborgs from being traitors.
tweak:Allows psychiatrist (counselor) to be a cultist again, as this is a leftover.
/🆑
As I've discussed with Carl a while ago that I'd do it for him, antagbans infantry (even if it's not in play right now), and @Sparrowbird611 asked to antagban the Adjudicator (went ahead and did it to the Bailiff as well).
I've also blacklisted cyborgs from traitor as it makes no sense to allow them to do so, since they don't provide anything really to a round. They can still be emagged and subverted with laws, but it won't let them be roundstart antags basically.


Tristan63:
more antag ban for things that never get picked anyway

PR #990, opened 2020-06-20
Revert "Regeneration and minor balance changes."

ProbablyCarl:
bugs galore

was this even tested?


Roland410:
It was actually, I just forgot about the surgery part of it (and probably shouldn't have given it to synthetics lmao).

PR #993, opened 2020-06-20
Ascent tweaks.

Sparrowbird611:
After some feedback and internal testing, I've made this PR.

Gyne armor was lowered a tier, down to BALLISTIC_RIFLE. It's still strong armor, but not as excessive as BALLISTIC_AP.

The Particle lance was tweaked. While unintentional, it seemed to cause seizures regardless of armor. It's new stats are similar to the stun rifle, but hit scan, and with a low clip that self recharges. Generally, if you miss the first shots, the foe is going to land multiple rounds into you before you can fire again.


SouthyOsmanli:
As far as I know - the bit about the lance was.. intentional. Or at least known about and not fixed for awhile.

BurgerLUA:
Failing tavis.

BurgerLUA:
Travis is failing because of another PR.

PR #998, opened 2020-06-23
Fixes Windoors

Superhats:
Fixes an issue which caused every secure windoor to have brig access.


Jackofashepard:
Travis no like

Roland410:
Travis fails are caused by the attempted fixes and are valid.

Superhats:
I can't map, so I actually don't know how to fix it.

PR #999, opened 2020-06-23
The new defualt for a failed vote is traitor

Jackofashepard:
You ever vote for merc and get denied because not enough people have their prefs on? 
You were looking for antag !FUN! but now it defaults to extended, lame!

No more, this is the PR for you, this changes the default to traitor instead of extended.

### DIFF
--- a/code/controllers/subsystems/ticker.dm
+++ b/code/controllers/subsystems/ticker.dm
@@ -11,7 +11,7 @@ SUBSYSTEM_DEF(ticker)
 	var/list/gamemode_vote_results  //Will be a list, in order of preference, of form list(config_tag = number of votes).
 	var/bypass_gamemode_vote = 0    //Intended for use with admin tools. Will avoid voting and ignore any results.
 
-	var/master_mode = "extended"    //The underlying game mode (so "secret" or the voted mode). Saved to default back to previous round's mode in case the vote failed. This is a config_tag.
+	var/master_mode = "traitor"    //The underlying game mode (so "secret" or the voted mode). Saved to default back to previous round's mode in case the vote failed. This is a config_tag.
 	var/datum/game_mode/mode        //The actual gamemode, if selected.
 	var/round_progressing = 1       //Whether the lobby clock is ticking down.
 
--- a/code/game/gamemodes/traitor/traitor.dm
+++ b/code/game/gamemodes/traitor/traitor.dm
@@ -12,7 +12,7 @@
 		but the painful reminder that space is cruel and unforgiving."
 	config_tag = "traitor"
 	required_players = 0
-	required_enemies = 1
+	required_enemies = 0
 	antag_tags = list(MODE_TRAITOR)
 	votable = 1
 	antag_scaling_coeff = 5


Hekzder:
In lieu of a removal of Extended, this is probably the next best thing.

Roland410:
This PR solves nothing and will cause more issues. Extended is the fallback mode because it doesn't need any antags and as such can be started without even any players being in the round.
As much as I dislike extended, it'll only make matters worse to have the default something that needs antags, especially since most people don't seem to have most single antags enabled from my experience.

JohnBigless:
Extended is probably a default gamemode for the sake of being even able to start a round during extreme lowpop. AFAIK, shit will break.

SouthyOsmanli:
Common issues are that no one readied-up has traitor or any sort of antag set to on - this is right next to the issue that no one readies up anyway. 

A solution is **forcing** people to play antag even if they don't have their prefs set to it like some other servers do; because the amount of times I've been on server during 30 pop to see the server trying to roll new antags - failing and having only 1 out of 5 possible antags (yet like 5 out of 5 renegades) is astounding.

Jackofashepard:
Actually, traitor requires 0 people with their traitor prefs on, in other words even in a event where there are 0 people on server or 0 with the prefs on it'll be exactly like extended and not break the server. This WILL however stop extended from being the defauly once a vote like merc fails where people actually want antags but get extended instead.

Jackofashepard:
image

Roland410:
If you check one line below, it needs at least one antag.
https://github.com/BoHBranch/BoH-Bay/blob/c7ef869fe9e262f42608689ee46a1281f0772a10/code/game/gamemodes/traitor/traitor.dm#L15

Jackofashepard:
Problem solved

Roland410:
DNM until staff decides.

BurgerLUA:
Not sure if I like this PR honestly, or at least, half of it. The default gamemode should be extended but I do think that autotratior should have 0 enemies required.

Hekzder:
> Not sure if I like this PR honestly, or at least, half of it. The default gamemode should be extended but I do think that autotratior should have 0 enemies required.

The default mode being Extended is frankly not good with rounds that are about 3h20m on average. That means a bungled vote will result in the next 3h20m being no antags AKA generally boring, **especially** for the likely majority of people that voted for an antag mode. 

Say the scenario is: Majority voted for Merc, not enough antags/players readied for Merc. I can say with almost certainty that every player who voted for Merc (the majority of players) would rather play Traitor than be stuck on Extended for the next 3h20m _minimum_. 

I've never understood the argument for Extended, the argument I've heard most commonly is "I want to RP", which doesn't make sense to me. The implication that RP doesn't occur on antag modes is patently false, obviously it's somewhat of an exaggeration, but for me personally, the best RP situations and scenarios I've had play out on this server have been on antag rounds as a result of antag actions. Because, surprisingly, conflict and unexpected things happening makes for good RP, in my opinion atleast. Sitting in your department for three hours making small talk because nothing is happening and a chunk of the server left when it turned Extended is not fun, in my opinion. 

To put it in a different perspective, the people who play/want Extended because they want to RP, can still RP on literally any mode. The people who play/want antag modes because they want conflict, antags, or just generally want more interesting rounds will rarely get conflict, definitely not get antags, and hardly get a more interesting round out of Extended. 

But yeah, that's my thoughts on Extended vs Traitor as a default mode, some of them atleast.

Roland410:
Even though I like the idea of this PR, staff decided against it, and as such I'm closing it.

PR #1001, opened 2020-06-24
Fix server crashes from syndicates endlessly shooting leeches

Lamasmaster:
Changes Leeches from faction "leeches" to faction = "syndicate".
This should prevent from syndicates endlessly shooting leeches and causing massive lag/server crash.


FunnyYellowDoge:
rip
fuckthis

PR #1002, opened 2020-06-24
Pouches no longer "heavier" than plates

Jackofashepard:
Removes slowdown from plate carrier pouches

### DIFF
--- a/code/modules/clothing/under/accessories/armor.dm
+++ b/code/modules/clothing/under/accessories/armor.dm
@@ -35,7 +35,6 @@
 	desc = "A collection of black pouches that can be attached to a plate carrier. Carries up to four items."
 	icon_state = "lpouches"
 	slots = 4
-	slowdown = 1
 
 /obj/item/clothing/accessory/storage/pouches/large/blue
 	name = "large blue storage pouches"
--- a/code/modules/clothing/under/accessories/armor_boh.dm
+++ b/code/modules/clothing/under/accessories/armor_boh.dm
@@ -1,10 +1,3 @@
-// Requested slowdowns
-/obj/item/clothing/accessory/storage/pouches
-	slowdown = 0.15
-
-/obj/item/clothing/accessory/storage/pouches/large
-	slowdown = 0.25
-
 /obj/item/clothing/accessory/storage/pouches/grey
 	name = "grey storage pouches"
 	desc = "A collection of grey pouches that can be attached to a plate carrier. Carries up to two items."


Superhats:
I personally think the slowdown should stay, and **specifically** on the large storage pouch. Removing it would mean giving people free storage with **zero** downside.

Historically, most of the "Upgraded" storage options (Rucksack, duffelbags), have had a slowdown to compensate for the extra slots.

I'd be fine with removing it on the standard pouches, but it should stay on the large storage pouches.

Jackofashepard:
Honestly, pouches are utterly for cosmetics at this point, why bother with higher slowdown than a actual armour plate provides when i can get a bandoiler that holds more than 3 times its items?

Superhats:
Then, that's an issue with the bandolier, not with the storage pouches. I'll need a second opinion, but I personally believe the slowdown is fine how it is. There's already issues of people powergaming as much storage as they can get.

Jackofashepard:
I really dont see why you would want this much slowdown on a item that mostly gets used for cosmetics as its basically useless for storage in itself, alas that is my opinion.

Superhats:
It's anything but useless. The standard storage pouches are still two extra pockets, which isn't negligible. The bandolier outclassing them isn't an argument, in my opinion. If anything, the bandolier should get nerfed.

Roland410:
Going to agree with Super on this. This is meant to give more storage and is totally not just cosmetic. The bandolier used to only work for ammo, and for some reason someone decided to allow it to store small items.

CrackParrot11:
I also agree with the other two, Having extra storage slots on with no downsides plus bullet resistant armor is a very large balance issue. It would end up causing another cascade of powerlevel creep from over buffing existing things and ruining the balances for them.

Roland410:
I'm leaving this open, as it's barely had a few hours to live, but unless you can somehow convince us to change our mind, or Burger overrides our decision, this probably won't be merged.
I don't think buffing security even more is that great of an idea.

Sparrowbird611:
I don't personally think it's as huge a deal as the others are making it out to be, but I do have to agree with them nonetheless.

Large pouches (No one uses small ones, they're shitty) grant you four storage spaces. If you decide to sacrifice your back slot for a weapon, this gives you space for items you can't store any where else (Medicinal pouches come to mind here). Four slots is still a pretty big deal for zero downsides, especially when you're probably wearing decent armor on top of it.

Roland410:
> I don't personally think it's as huge a deal as the others are making it out to be, but I do have to agree with them nonetheless.
> 
> Large pouches (No one uses small ones, they're shitty) grant you four storage spaces. If you decide to sacrifice your back slot for a weapon, this gives you space for items you can't store any where else (Medicinal pouches come to mind here). Four slots is still a pretty big deal for zero downsides, especially when you're probably wearing decent armor on top of it.

Most weapons also go on the armor's storage slot excluding stuff such as sniper rifles etc.

Roland410:
Staff has voted against this. Closing as such.

PR #1005, opened 2020-06-25
Derelict Station crashing server

Lamasmaster:
Removed ranged syndicate mobs from the map so this time it actually shouldnt crash.


Roland410:
You also need to mapmerge each map you change for future reference.

PR #1008, opened 2020-06-26
Gives CoS more access

LoreCoder:
Gives the CoS access to the Polyp and Polyp Security (because for some reason, they didn't have it but other sec members did) and Medical Lobby, because it's not fun to wait for AI/Engineer/Medical to open the door when you're a head of staff and should already have access.

Thanksies to Lamasmaster for showing a noob how to make a PR.

### DIFF
--- a/maps/torch/job/command_jobs.dm
+++ b/maps/torch/job/command_jobs.dm
@@ -266,8 +266,7 @@
 			            access_maint_tunnels, access_external_airlocks, access_emergency_storage,
 			            access_teleporter, access_eva, access_bridge, access_heads,
 			            access_hos, access_RC_announce, access_keycard_auth, access_sec_doors,
-			            access_solgov_crew, access_gun, access_emergency_armory, access_hangar, access_medical,
-						access_petrov, access_petrov_security)
+			            access_solgov_crew, access_gun, access_emergency_armory, access_hangar, access_petrov, access_petrov_security)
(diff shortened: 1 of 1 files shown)


Sparrowbird611:
I support polyp and polyp security if the other officers have it, but I'm against handing medical access to the CoS. He doesn't need it.

SouthyOsmanli:
Agreeing with Sparrow. All of sec used to have Medbay access. It was revoked due to Sec constantly and consistently abusing it and crowding Medbay. 

COS’ abused it and, personally, I see little reason in giving it unless you’re going to give the CMO Sec access just in case of medical emergencies in return.

LoreCoder:
My point is, the chief of the security department shouldn't have to wait for someone to open up when they could be dealing with an emergency. This is especially worse when there's no medical and you have to call engineering to hack open the door. But if no one wants COS to have med access then I'll remove it and leave just the Polyp access.

Roland410:
Security having access to medical was quite controversial as they often lingered around inside taking up space from medical. Not sure if giving it to the CoS (who usually is the most entitled person of all of security) the access is a good idea.

Superhats:
I'll agree with that. We already have way too many unrelated roles with access to the medical lobby, the CoS really doesn't need it, considering the ETC is two tiles away, and there should always be a doctor inside.

PR #1009, opened 2020-06-28
Kompoton balance pass

Jackofashepard:
So kompotons been in for a bit now and i personally have played around with it and watched others, i can safely say it causes heart attacks a _teeny tiny_ bit too often and tators keep dying from brain damage, which in turn came from their organ damage, while they are stunned. I have made the following changes with this in mind:

- Kompoton now does a slight less amount of agony. From 2 -> 1 a tick

PS. It still has a chance of giving you a heart attack under certain conditions and I HIGHLY reccomend taking the strongest pain killer you have with it. Aksy is also reccomended if your wounds warrant it (blood loss means brain kill)

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
+++ b/code/modules/reagents/Chemistry-Reagents/Chemistry-Reagents-Medicine.dm
@@ -481,8 +481,7 @@
 	overdose = 15
 	scannable = 1
 	flags = IGNORE_MOB_SIZE
-	var/agony_dose = 5
-	var/agony_amount = 2
+	var/agony_amount = 1
 	value = 6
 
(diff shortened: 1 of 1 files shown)


Jackofashepard:
As a sidenote i killed a line of code that wasnt doing anything too

Roland410:
I don't think allowing it to heal brain damage is a good thing. Traitors can already get Alkysine from the uplink, heal bones with Hypeross and organ damage with Kompoton.
This would mean they don't need Alkysine at all, which makes it a bit more general medicine, and a fair bit more OP.

Superhats:
I'll second that. Traitors really shouldn't have access to some super healing chem, and it healing brain damage COMPLETELY invalidates alkysine, since brain damage typically means you have damage somewhere else, which kompoton would take care off.

Jackofashepard:
Brain regen removed as suggested

PR #1010, opened 2020-06-28
Infantry is back

Jackofashepard:
As hoffmans map died and he had to take a break due to significant backlash i have taken it upon myself to make this PR.

It simply re-adds the infantry with rule and SOP adjustments which are detailed in this document https://docs.google.com/document/d/1A2XIah5dlu8YX18zeu9zrzIL0S39VBI9EzRnSW2SbJs/edit

One such adjustment is they all awnser to the pathfinder first and the SL second. They are also disallowed from valid hunting unless very specific conditions are met such as the valids being boarders or code delta being initiated.

### DIFF
--- a/maps/torch/job/hestia_jobs.dm
+++ b/maps/torch/job/hestia_jobs.dm
@@ -52,11 +52,12 @@
 
 /datum/job/squad_lead
 	title = "Squad Lead"
+	supervisors = "The Pathfinder"
 	department = "Infantry"
 	department_flag = INF
 	head_position = 1
-	total_positions = 0
-	spawn_positions = 0
+	total_positions = 1
+	spawn_positions = 1
 
 	supervisors = "the highest ranking Marine and SolGov Command"
 	selection_color = "#557e38"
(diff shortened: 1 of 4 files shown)


Sparrowbird611:
~~So it's treason then!~~

Just kidding. I still stand by all my previous statements though against this. SOP solves nothing with the issues infantry already possesses, and we're better off giving exploration better gear and improving them instead of beating a dead horse by trying infantry again.

Jackofashepard:
The new inf prep (yes merchant still works i did not extend the walls) 
image

Jackofashepard:
A warning to any mappers, there's a green wire that works and one thats a dud, dont make my mistake

Aurelien3189:
Personally, until Infantry gets a massive rework, that is, beyond a few guideline/RP changes, I don't think most of the core issues they had last time would go away. Given that they're supposed to serve as an escort, as opposed to the main participants for exploration, I feel like all the trouble made to balance them could be used to flesh out exploration more, as Sparrow mentioned, to give them proper gear to hold out themselves.

Otherwise, they seem like a niche role for mil sim or TDM players coming from a squad based server. To bring up an old point, with the way they require boarders or exploration to be active at all, I feel they're better as an off ship role, at least to eliminate the need to put tight checks on them with guidelines. 

Either way, I feel Infantry's job can be accomplished with expansions to explorer gear or security, or perhaps just making ERTs more responsive, or fixing them from breaking all the time for that matter.

Tupinambis:
inb4 infantry gets removed again.

Unless infantry have a consistent and interesting role to play every single round, the role is never going to function as intended. The only way infantry will ever work is with reliable and persistent off station military level threats. This includes stuff like vox and ascent, but they aren't consistent spawners and are not consistently threats anyway. Exploration and off-station stuff in general is very light on content as is, we don't need to bloat the server with roles that have nothing to other than be security+.

SilvernKell:
Infantry just does not work when used like this. It creates inconsistencies ic "Why can't command use them? Oh, because uhh. . [excuse for balance reasons]", They just don't work. I like the role. But it doesn't have a place in the game when used like this. It is essentially redundant when you can play SMC security.

maps/torch/job/hestia_jobs.dm
```diff
@@ -79,15 +80,15 @@
 							 /datum/computer_file/program/reports)
 
 /datum/job/squad_lead/get_description_blurb()
-	return "<span class='warning'>You are NOT Security. Ignoring this will get you job banned, or worse.</span> - You are a Squad Leader. Your duty is to organize and lead the small infantry squad to support the Pathfinder. You command Marines in your Squad. You make sure that expedition has the firepower it needs. Once on the away mission, your duty is to ensure that the worst doesn't become reality."
+	return "<span class='warning'>You are NOT Security. Ignoring this will get you job banned, or worse.</span> - You are a Squad Leader. Your duty is to organize and lead the small infantry squad to support the and you awnser to the Pathfinder. You command Marines in your Squad. You make sure that expedition has the firepower it needs. Once on the away mission, your duty is to ensure that the worst doesn't become reality."
```
Roland410, line 83:
```suggestion
	return "<span class='warning'>You are NOT Security. Ignoring this will get you job banned, or worse.</span> - You are a Squad Leader. Your duty is to organize and lead the small infantry squad to support the and you answer to the Pathfinder. You command Marines in your Squad. You make sure that expedition has the firepower it needs. Once on the away mission, your duty is to ensure that the worst doesn't become reality."
```

PR #966, opened 2020-06-13
Gun Tweaking.

Sparrowbird611:
Alright, lets do this.

First off, majorly reduces the number of laser carbines around the Dagon. Every laser carbine on bridgedeck has been removed or replaced, with the full list being these.

Byakhee: Two laser carbines. Updated version: replaced with three smart guns
Bridge emergency bunker: three laser carbines. Updated version: replaced with three smartguns
Bridge protection armory: Four laser carbines. Updated version: Replaced with one pump shotgun, and two SMGs. Comes with four SMG mags, one box of lethals for the shotgun, and one box of non lethals for the shotgun.

Next off, buffs the XO's pistol. Original stats: 45 damage, zero AP. new stats, 45 damage, 10 AP. Sufficient against normal forms of armor, but not too insane.

The thing we've all been waiting for. AMR no longer has screen shake. This  should prevent it from breaking your screen immediately after firing while scoped.

Next off, I removed the foundation's nullglass weaponry from the Dagon. Psionic threats are seldom seen, and they rely upon insane amounts of RNG to acquire a great power that'd even give them a chance against security.

I thought I'd tweak two underused picks for traitors, namely, the grenade launcher and the Ion pistol. The Grenade Launcher was reduced to 40 TC, from 60, but no longer spawns loaded with grenades. The Ion pistol was just flat out buffed to 10 TC, from 20. Before any one riots at me, the haywire shell and zip gun slug combo come out to five tc, and a current PR will likely buff the zip gun to 1 TC, making it even better **Update** Ion gun moved to 16 TC, Haywire moved to 3 TC.

Finally, infantry prep has had all of it's weapons, C4, and other weird equipment removed. It's now just an empty room for the most part.

PR #1011, opened 2020-06-28
Gun Tweaks (Definitely the original)

Sparrowbird611:
Literally everyone else: You can't just make a totally new PR with the same changes, you have to fix the original one!

me: Hahah commit button go brr- **Bonk!**

It's https://github.com/BoHBranch/BoH-Bay/pull/966 but without the foundation revolver changes and the map is fixed this time.


Sparrowbird611:
Forgive me Roland for I have sinned.

Aurelien3189:
Didn't comment on the last PR, but seems like a pretty welcome, well rounded change to flatten the curve with the less used antag weapons. Would play TM'd again

PR #1013, opened 2020-06-29
guaranteed O2 closet supplies

Aurelien3189:
Changes the RNG spawns of the o2 closets, so that they're guaranteed basic gear for one person (a spacesuit, mask, extended oxygen tank, and 2 o2 pouches), as well as some tweaks. Also changes it so that radiation closets spawn with 1 toxin and 1 rad pouch as opposed to 2 toxin pouches, given that only engineering gets it. 

This ideally should end the usual loot box esque hunt that is frantically opening O2 closets for a space suit while there's a breach, as previously suits only had a 25% chance of spawning. Changes are as follows:

image

TE4y66AXKR

image

image

image

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/utility_closets.dm
+++ b/code/game/objects/structures/crates_lockers/closets/utility_closets.dm
@@ -114,8 +114,8 @@
 
 /obj/structure/closet/radiation/WillContain()
 	return list(
-		/obj/item/weapon/storage/med_pouch/toxin,
-		/obj/item/weapon/storage/med_pouch/radiation,
+		/obj/item/weapon/storage/med_pouch/toxin = 2,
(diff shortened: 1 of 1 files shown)


Roland410:
I'd be fine with these changes, but I think making the toolbox and the first aid kit etc. guaranteed as well would be better.

Aurelien3189:
The toolbox, I think is common enough that it might as well be stanrdard, but I'll think on it, in case anyone else pitches in. First aid kit might be a tad bit much, given how common O2 lockers are, I saw it more as a bonus. It's certainly worth considering though.

Aurelien3189:
A bit late, but I haven't received much else in the way of suggestions as of late, personally I think it's well rounded as it stands described here.

Roland410:
Want me to merge it like this then?

PR #1023, opened 2020-07-01
Lasers no longer .delete organs

Jackofashepard:
- laser organ damage from 2 to 0.8 times the damage done (other factors do apply)


Roland410:
I feel lasers should still be more damaging by a lot than ballistics. Ballistics can embed via shrapnels etc. which is their long term damage. This is a pretty big nerf for laser weaponry.

Jackofashepard:
Suggested value?

Roland410:
Test with like 1.2 maybe? Ballistics should stay as-is for the above.

Jackofashepard:
From number crunching 1.2 is still really high, maybe 0.8?

PR #1027, opened 2020-07-02
Slight agony adjustments

Jackofashepard:
In this PR i adjusted the modifiers revolving around agony ever so slightly in a effort to make pain slightly less of a meme, in other words you will feel slightly less pain than normal. All changes made were incredibly slight as i'd rather not risk a massive pain nerf.


Roland410:
I personally feel like pain is in a good place right now.

Superhats:
I'll second that.

PR #1028, opened 2020-07-02
Lace splat

Jackofashepard:
laces now play a message and go splat on destruction. Also laces now have a max damage of 100 as compared to their previous default of 30


Jackofashepard:
Closing for fixes

PR #1029, opened 2020-07-02
Adds destruction effects to laces

Jackofashepard:
Splat effects when it gets destroyed

### DIFF
--- a/code/modules/organs/internal/stack.dm
+++ b/code/modules/organs/internal/stack.dm
@@ -66,7 +66,7 @@
 	gore.throw_at(get_edge_target_turf(src,pick(GLOB.alldirs)),rand(1,3),30)
 	gore = new /obj/effect/decal/cleanable/blood/gibs(get_turf(owner))
 	gore.throw_at(get_edge_target_turf(src,pick(GLOB.alldirs)),rand(1,3),30)
-	owner.visible_message(SPAN_WARNING("[owner]'s neck explodes in a shower of strange blue liquid and metalic fragments!"))
+	owner.visible_message(SPAN_WARNING("[owner]'s neck explodes in a shower of strange blue liquid and metallic fragments!"))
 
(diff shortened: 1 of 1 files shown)


Roland410:
I personally prefer laces to be more vulnerable. As it stands that was at least one drawback you had. Your lace goes pop, you die. With it having 100 health that's a lot harder to achieve.

Sparrowbird611:
A fair point, but isn't current lace HP only 25? A smartgun matches that number, and could probably pop a lace in two shots.

Roland410:
Yes, but you're more likely to hit the brain, and you'd get a lot of damage reductions anyways, because the skull counts as encasing, which reduces the chance/damage I don't remember by at least half iirc? On top of that the brain has a size of like 60 or something while the lace now has 25.

code/modules/organs/internal/stack.dm
```diff
@@ -58,6 +58,16 @@
...
 
+/obj/item/organ/internal/stack/Destroy()
+	. = ..()
+	var/obj/gore
+	playsound(src, "shatter", 70, 1)
+	gore = new /obj/item/weapon/material/shard(get_turf(owner), MATERIAL_GLASS)
+	gore.throw_at(get_edge_target_turf(src,pick(GLOB.alldirs)),rand(1,3),30)
+	gore = new /obj/effect/decal/cleanable/blood/gibs(get_turf(owner))
+	gore.throw_at(get_edge_target_turf(src,pick(GLOB.alldirs)),rand(1,3),30)
+	owner.visible_message(SPAN_WARNING("[owner]'s neck explodes in a shower of strange blue liquid and metalic fragments!"))
```
Roland410, line 69:
```suggestion
	owner.visible_message(SPAN_WARNING("[owner]'s neck explodes in a shower of strange blue liquid and metallic fragments!"))
```

BurgerLUA:
oh lmao I no longer have access

PR #1076, opened 2022-02-01
Removes a couple of nonsensical skrell-related Xenowear loadouts.

Kaksisilma:
**Removes the following items from the loadout options:**
Skrellian Holo-Badge
Skrellian Uniform
Skrellian Exchange Voidsuit (Also deleted from code. SSV has another version with better qualities)
Skrellian Exchange Voidsuit Helmet (Also deleted from code. SSV has another version with better qualities)

**Why?**
All the items above are directed towards skrell on SDTF exchanges with humanity. With the war, it is assumed that exchange programs got cancelled, and therefore, no skrell onboard the Dagon would be able to have those.

This does **not** affect the Skrellian Helmet, as it is one of the few options of decent headgear for skrell security.

This does **not** affect the SSV offship, who will still spawn with holo-badges and uniforms regardless if whether or not they are taken on loadout.

SDTF Skrell have no reason to be onboard independent ships such as the Verne or the Bearcat, and thus it should not obtainable for skrell in those positions.


ProbablyCarl:
Good stuff.

PR #1079, opened 2022-02-01
Added The Cuchulain Foundation faction

Noobetski:
The name. Also i'm still somewhat new, how to make changelog?


ProbablyCarl:
Don't worry about changelogs. We don't utilize them as of current.

Noobetski:
It's bay wiki copy-paste, but it should do until lore regarding foundation is rewritten

Noobetski (commented):
I fixed error, yay

PR #1091, opened 2022-02-02
fuck you bay

Vizoplez:
removes god-awful cancer mechanic that resulted in nearly everyone dying to awakening their psionics

https://www.youtube.com/watch?v=g7_I6kj9fTc


Noobetski:
If it's caused only by Three Eye it's a feature so people would actually ask help from psionic advisor with awakening, if it works on any kind of awakening it's fucked

Vizoplez:
> If it's caused only by Three Eye it's a feature so people would actually ask help from psionic advisor with awakening, if it works on any kind of awakening it's fucked

this is false and I do not believe you have read the code.

This applies to all types of awakenings, including brain damage awakenings.

Squigger:
Should really be capped at a threshold rather than removed, so it's the more dangerous option than asking the PADV for assistance. Chugging drugs to go fight interdimensional Egyptian deities should hold some risk.

Vizoplez:
> Reference in new is

done

ProbablyCarl:
Good stuff, with not complete removal.

PR #1107, opened 2022-02-04
Adds an overmap hazard which now causes the gravity failure event.

anconfuzedrock:
Adds dark matter influxes, overmap hazards that force ships with gravity generators to turn them off. Pretty simple, but I think it's a nice hazard for three reasons:

First, if you're not the one piloting, it usually doesn't really feel like the ship is moving. This hazard is harmless enough that the ship might consider flying through it (like the ion storm), and you *will* feel the effects of it (unlike the ion storm), which can make the movement of the ship feel more important.

Second, shuttles don't always get as much use as they could, which is boring for pilots. After all, why let your little bugger fly somewhere when you can just let big one go there itself? Well, this doesn't change the dynamic _much_, but as a hazard that's harmless to shuttles but harmful to the main ship, it might encourage you to let them go on ahead.

Third, the gravity failure event is neat, there's no arguing that, but there's not really much of a chance to... play around it. You never know when one's going to happen, and when one DOES happen, you never know how long it'll last, so trying to take advantage of it is way too risky. Now it's a little bit more predictable, and you might even be able to cause this on purpose for whatever reason. And yes, you can intentionally drive into and out of this hazard repeatedly to knock people over.


ProbablyCarl:
Absolutely based PR. Good stuff.

PR #1109, opened 2022-02-05
Ejector windoor fix

Dimasw99:
Fixes access for the engine ejector windoor located in the engine monitoring room. Its access was set to brig and now is ACCESS_ENGINEERING.


ProbablyCarl:
Don't think you've done this correctly. Might want to close this and try again in order to make the singular edit.

Dimasw99:
Probably. I am not sure what type of mapping tool you are using. I tried both DM and FASTDMM with both having weird results

PR #1113, opened 2022-02-05
Windoor SM eject fix

Dimasw99:
Today I found out that TGM exists. This PR fixes the windoor access for the SM ejection.


ProbablyCarl:
Good stuff

PR #1125, opened 2022-02-06
Adds manual landing

Anton-Kr:
You can now manually land all shuttles. Works for exoplanets so far, but not stations etc. Downside of this is that the tiles reveal structures but imho it's a worthwhile trade off.


ProbablyCarl:
Looks good at a cursory glance. Might be worthwhile testmerging this for a bit.

Vizoplez:
This isn't forced but would be nice.

-Fixing the AI camera bug (now just a typepath issue)
-Ensuring it now works.

Frankly AI is mega low-priority and I'll fix it myself sometime in the future, though. Works well.

Vizoplez (approved):

Anton-Kr:
Im too stupid to fix any of that so if you could handle it, I'd highly appreciate it.

PR #1130, opened 2022-02-06
Fuck you. *Removes abusable traits*

Vizoplez:
Destroys some genemodding traits due to their inherently terrible design, and buggy nature. Also attempts to fix the spawn over limit issue.

This is part one of a potentially four-part removal series, up into disabling the replacement of the chest, and brain, as a genemodder.

Do not do this again. You know who you are.

### DIFF
--- a/modular_mithra/code/modules/mithra/traits_positive.dm
+++ b/modular_mithra/code/modules/mithra/traits_positive.dm
@@ -6,7 +6,7 @@
 	name = "Agile"
 	desc = "You can move faster than everyone else."
 	cost = 4
-	var_changes = list("slowdown" = -0.5)
+	var_changes = list("slowdown" = -0.2) //No.
 
 /datum/trait/darksight
 	name = "Tapeta Lucida"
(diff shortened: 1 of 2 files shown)


Vizoplez:
Waiting for Carl's comment on this, along w/ whoever wants to pitch in.

Pariah919:
Should just delete genemodder traits alltogether, most of them seem pretty innane for nearly no downside.

Species in general need an balance pass badly probably.

modular_mithra/code/modules/mithra/traits_positive.dm
```diff
@@ -6,7 +6,7 @@
 	name = "Agile"
 	desc = "You can move faster than everyone else."
 	cost = 4
-	var_changes = list("slowdown" = -0.5)
+	var_changes = list("slowdown" = -0.2) //No.
```
NotRanged, line 9: this is still coded backwards and makes you slower instead of faster lol

(or maybe that was some cryo randomness one time, idk actually)

ProbablyCarl:
Removal of character freedoms is always unfortunate.
If we could, I'd prefer to keep telepathy, as we highly limited it prior. It's pretty much only worthwhile in an RP aspect at this point.

That aside, this is good. I'll merge this shortly.

modular_mithra/code/modules/mithra/traits_positive.dm
ProbablyCarl: Nadda. This is actually correct. If it were the opposite, it'd be a slowdown.
Negative slowdown increases the speed.

PR #1131, opened 2022-02-06
organ sprites and hud implants

Kaksisilma:
ported directly from bay

Adds new organ sprites by Azlon PC and some by PebblesThe2nd

Also adds Hud Implants by SealCure


Vizoplez:
Seems to remove ascent organs we might still be using.

Kaksisilma:
> Seems to remove ascent organs we might still be using.

I'm not sure how you came to this conclusion, but this is not what happens. You're free to download the branch or the specific DMI to check if you want, though.

ProbablyCarl:
Doesn't look like it negatively touches Ascent organs.
@Yawet330 is mistaken in this case.

PR #1136, opened 2022-02-07
Schizoshift

Vizoplez:
Sprite shifting. For ERP.


Vizoplez:
credits to vesta or w/e

honestly should fix a ton of our broken stuff because our interface was jacked

PR #1137, opened 2022-02-07
Polyp Military Refit 

Mannybrado:
Refits the Polyp to house the infantry squad. Repainted and somewhat remodeled, the Polyp has never been so useful!

Note, that the thrusters do NOT work, and that you cannot detach from the Dagon without being stranded in space. The Dagon must be near the location you wish to go for successful deployment.

The oldnew infantry prep location has been deleted entirely and made into an extension of the observation deck. The stairs leading down still exist for quicker access to the main docking and escape shuttle ports.

PLEASE TESTMERGE FIRST
Open to feedback.
polyp


N8-Toe:
They shouldn't get an OR. The combat medic shouldn't be doing surgery. I just don't think them being on the polyp makes sense? why would you have them on a detachable part of the ship.

Squigger:
Why is there so much on it that invalidates interacting with the rest of the ship? Cells, so infantry are encouraged to act like Security more? Stocked Medbay? Their own jailbroke lathe so they can just print endless ammo without interaction? I'm sure there's more tucked away in there that I can't see at first glance, but this seems like massive creep to make infantry entirely self-sufficient and more powerful than they already are in their own fortress.

Mannybrado:
> Why is there so much on it that invalidates interacting with the rest of the ship? Cells, so infantry are encouraged to act like Security more? Stocked Medbay? Their own jailbroke lathe so they can just print endless ammo without interaction? I'm sure there's more tucked away in there that I can't see at first glance, but this seems like massive creep to make infantry entirely self-sufficient and more powerful than they already are in their own fortress.

good feedback! Thanks.

PR #1147, opened 2022-02-09
adds one use RPGs to the code

Pariah919:
One use RPGs, they fold and then unfold.

Code bad, RPGs are still baycode, I need to convert them to guns in my next PR.


Pariah919:
i'm rewriting rpg code

PR #1149, opened 2022-02-09
Redoes RPGs to be projectiles, makes RPG projectiles gib on hit, redoes RPG rockets to be projectiles, One use RPG added.

Pariah919:
## About The Pull Request
Rockets work off projectile code so they have travel time, as otherwise they're hitscan (Which is obviously a fact that shouldn't really be a thing.)

Rockets gib on hit, as they already remove several of your limbs on hit. And I believe this is more satisfying for their incredibly high damage, plus they're easy to dodge considering-

Rockets now have windup to fire, its short and .5, but this stops you from shooting at point blank range, which should be obviously a bad idea.
Rockets have a new sprite in the air (though you couldn't see the old one anyway.)
Adds one use rockets with inhands, they're not in the game but they exist, they use normal rockets.

https://streamable.com/expbqj

## Why It's Good For The Game
Makes rocket launchers more unique rather than whatever they are as hitscan instahit weapons.

## Changelog
:cl:
Codeimprovement: Rockets aren't poopy coded launchers, and instead work off projectile code.
Add: One use RPGs, plus marine version, not in the game yet.
Balance; Rockets gib on hit, rocket launchers take half a second to fire. Rocket launchers have a high one handed penalty
Fix: Rockets are no longer hitscan.
sound: Rockets have vastly improved sound effects.
/:cl:


ProbablyCarl (commented):
Check annotations going wacky again? Regardless, good stuff.

PR #1161, opened 2022-02-10
Modular computers no longer give off light when turned off.

anconfuzedrock:
PDAs, tablets etc, now only give offflight when the screen is on, allowing you to sneak around a bit better. Thanks to @Hubblenaut for letting me use their coode.


Kaksisilma:
Looks good to me.

PR #1166, opened 2022-02-11
Makes revolvers break open to be reloaded and eject casings.

Pariah919:
Makes revolvers require to be broken open to be reloaded, and makes them drop all casings when opened.
Adds sprites to support this.


ProbablyCarl:
based

ProbablyCarl:
Looks like the standard sprite for the sidearm revolver is open by default?

Pariah919:
>Looks like the standard sprite for the sidearm revolver is open by default?
>
forgot to reset something when editing it, fixed.

PR #1183, opened 2022-02-12
Removes Diona and Prommies from any form of sec/SolGov roles.

Vizoplez:
No.


ProbablyCarl:
We'll be going ahead and test merging this briefly to see opinions.

ProbablyCarl:
On further inspection, I'm afraid we can't merge this. Please blacklist the specific roles that are issues, rather than blocking the entirety of them. We've quite a few people who don't powergame in them as is. Punishing 'em as a whole is not the way to go.

PR #1184, opened 2022-02-12
Removes unused areas

Kaksisilma:
Removes one unused area flipping CI off.


Kaksisilma:
Bit of a crackpot way to fix this. Alternatively, we can comment out the baton and re-enable it when we need it. @ProbablyCarl

ProbablyCarl:
The stunbaton requires an area that's called elsewhere, and it likes to screech when it's separate from what I recall.
That's why the area in the file exists. It _should_ be safe to remove, but Github/Byond may have a fit.

Kaksisilma:
Eh. One block of code in an area file won't do much harm, and we get one step closer to a green CI.

PR #1189, opened 2022-02-13
Speeds up surgery steps slightly

Kaksisilma:
Severe injuries happen very frequently in this codebase, and as a result I have observed that medical gets incredibly bottlenecked on most rounds where a couple of people start to die.

This PR attempts to alleviate the issue by making most surgery steps slightly faster, and also buffing the Incision Manager System to make it a more interesting research option for Science.

I'm all ears if anyone has any suggestion regarding the new surgery times.

### DIFF
--- a/code/modules/surgery/bones.dm
+++ b/code/modules/surgery/bones.dm
@@ -23,8 +23,8 @@
 	)
 	can_infect = 1
 	blood_level = 1
-	min_duration = 50
-	max_duration = 60
+	min_duration = 40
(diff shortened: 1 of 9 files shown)


Vizoplez:
Min-durations should be lower still, but the max-durations are now good. Having a super-skilled surgeon rn is pointless

Kaksisilma:
Most of the feedback I got from this was 'It's fine as it is tbh' so I'm not merging it.

PR #1200, opened 2022-02-14
Renames Passenger to Off-Duty and removes most of it's alt-titles

Mannybrado:
Its a Sol Gov Military ship. No more will there be random bums laying about doing nothing. If you like sociologist or psychologist, just play counselor and RP being a researcher.
Keeps journalist and botanist because journalist is cool and botanist has utility.
EDIT: Passenger has been readded as an alt-title for rescued crash survivor RP or whatever.

### DIFF
--- a/maps/torch/job/misc_jobs.dm
+++ b/maps/torch/job/misc_jobs.dm
@@ -30,7 +30,8 @@ Civilian
 	announced = FALSE
 	alt_titles = list(
 		"Journalist" = /decl/hierarchy/outfit/job/torch/passenger/passenger/journalist,
-		"Botanist")
+		"Botanist"
+		"Passenger" = /decl/hierarchy/outfit/job/torch/passenger/passenger)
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
These are all titles currently in use, actively, by various players.
While some of them don't thematically make a lot of sense, it'd be wacky removing them while people are getting genuine RP outta 'em. That's my take on it.

Mannybrado:
> 

The only one I've seen used is journalist, which is still present. If some people could come forward to say which ones they use the most, I'll gladly keep them. I mostly want to trim fat.
I'm not going to compromise on independent observer or investor though. Those ones are the worst ones. I think I will go and readd passenger regardless as an alt title to this for the continued RP of being a crashed survivor or whatever.

Kaksisilma:
CI is failing and the overall premise of this PR is not very good. I'll go ahead and close this.

PR #1201, opened 2022-02-14
Fixes CI

Kaksisilma:
Fixes one of many checks. Probably.


Kaksisilma:
I'm going to gut the next person who merges any PR that breaks CI. This is ready to be merged either way

PR #1207, opened 2022-02-15
Adds pools

Anton-Kr:
Adds pool to D5. Rearranges the D5 fore area slightly to give it the pool and a sandy gym area. Gives 2x2 pool to CO. Changes around some fluid code to accommodate change.


Squigger:
Why?

Anton-Kr:
@Squigger because its a good change

PR #1217, opened 2022-02-16
GAS Job Restrictions

IvanIvanovIvanovsky:
Allows all GAS to be civilians now that they are citizens under the lore. Adds Bartender and Chef to grade A, as before it was only grade B. This is my first PR in over six months so I'm keeping it simple. Playtested, looks like it works fine to me. May create another more expansive PR for reworking the shitcode that is GAS grading. I'd like feedback on what jobs you think GAS should or should not be in so feel free to comment it here incase I do make that PR.

### DIFF
--- a/maps/torch/job/torch_jobs.dm
+++ b/maps/torch/job/torch_jobs.dm
@@ -58,7 +58,7 @@
 	LAZYADD(valid_jobs, /datum/job/scientist_assistant)
 	..()
 	
-	/decl/cultural_info/culture/nabber/New()
+/decl/cultural_info/culture/nabber/New()
 	LAZYADD(valid_jobs, /datum/job/assistant)
 	..()
 
--- a/maps/torch/job/torch_jobs.dm
+++ b/maps/torch/job/torch_jobs.dm
@@ -1,7 +1,7 @@
 /datum/map/torch
 	species_to_job_whitelist = list(
 		/datum/species/nabber = list(/datum/job/ai, /datum/job/cyborg, /datum/job/janitor, /datum/job/scientist_assistant, /datum/job/chemist,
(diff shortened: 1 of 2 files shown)


Kaksisilma:
Idk about passenger GAS but eh. Additions to A are fine.

Also:
Capture

maps/torch/job/torch_jobs.dm
```diff
@@ -57,6 +57,10 @@
 /decl/cultural_info/culture/nabber/New()
 	LAZYADD(valid_jobs, /datum/job/scientist_assistant)
 	..()
+	
+	/decl/cultural_info/culture/nabber/New()
```
Kaksisilma:
```suggestion
/decl/cultural_info/culture/nabber/New()
```

Kaksisilma (changes requested):
This will probably fix it.

IvanIvanovIvanovsky:
Once I downloaded a copy and it didn't work I edited it in DM until it did, to save time, and so I copied the changes by memory to the github version, seems like I made an error.

Vizoplez:
Good for merge if @Kaksisilma agrees

benj8560:
wanted to suggest giving them prospecting. Always kinda made sense they would suit that kinda work. They got RIGs now and their strength makes them well suited for any labor job. I believe prospectors already get EVA so them getting their suits is already taken care of.
also no doubt some funny memes will come of having a GAS in a small enclosed space on the Gaunt *chitter

Kaksisilma:
> wanted to suggest giving them prospecting. Always kinda made sense they would suit that kinda work. They got RIGs now and their strength makes them well suited for any labor job. I believe prospectors already get EVA so them getting their suits is already taken care of. also no doubt some funny memes will come of having a GAS in a small enclosed space on the Gaunt *chitter

Sounds good on paper but on practice I think it'll be a nightmare.

Code-wise, this PR is OK so I'll go ahead and merge it. If push comes to shove we can remove passenger GAS later.

PR #1218, opened 2022-02-16
Revert "Removal of GAS from Antagonistic Roles"

ProbablyCarl:
No longer needed, as per discussions in Discord with the implementation of 'Species' bans.


N8-Toe:
I think even with the species bans, Thing such as GAS ling should not be a possibility

PR #1224, opened 2022-02-16
Dexalin plus changes

Pariah919:
Dexalin plus no longer gives you 80% oxy, now only 70%.
It metabs much faster.
Removes oxy loss from the body.


Kaksisilma:
Might be interesting to testmerge this.

ProbablyCarl:
I'll go ahead and test merge this.

PR #1230, opened 2022-02-17
Psi should no longer risk instantly killing people

Vizoplez:
New dex+ changes made it so if merged, psi was impossible to survive.
For some reason braindamage apply procs take into species resistances and were making even regular humans take 80 brain damage per faculty.


ProbablyCarl:
This will be test merged alongside #1224

PR #1234, opened 2022-02-18
Shamelessly buffs drop pouches

anconfuzedrock:
Drop pouches are now on par with webbings instead of having 1 less slow. Many people think drop pouches are way sexier than webbings, they have personally gotten me four girlfriends. No one should have to choose between 1 storage slot and a girlfriend.

"but rock, drop pouches are slotless storage, making them slightly better!"
Do you have any idea how little that effects? The only jobs I can think of that would get benefit from this are like, doctors carrying 3 syringes and cargo techs carrying both stamps- But there's plenty of other things with slotless storage, like wallets or YOUR BACKPACK. All backpacks and boxes having slotless storage means the advantage is, at most, a convenience.

"but rock, why don't you just make webbing slotless as well?" Main reason being that it'd be snowflakey as fuck. the code for webbings is ass, so it'd be snowflakey as fuck, none of them are subtypes of the rest. you'd have to squeeze this proc in like 4 times. And again who CARES about slotless storage.

### DIFF
--- a/code/modules/clothing/under/accessories/storage.dm
+++ b/code/modules/clothing/under/accessories/storage.dm
@@ -80,7 +80,7 @@
 	slots = 5
 
 /obj/item/clothing/accessory/storage/drop_pouches
-	slots = 4 //to accomodate it being slotless
+	slots = 5
 
 /obj/item/clothing/accessory/storage/drop_pouches/create_storage()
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
based rock pr

PR #1236, opened 2022-02-18
Pouches

Technobug14:
It works but they could use some polish. Currently you can't grab from the pouch by just clicking, you'd have to click-drag it to your character to take stuff out without taking it out of your pocket. I'd like this to be the other way around but I'll do this at a later date. Or somebody else can do it, I hate Bay code.


ProbablyCarl:
Incorrect indentations.

Anton-Kr:
I'd prefer pockets to remain pockets unless you can resolve the clicky behavior you mentioned. If you can fix that, I will happily merge this PR. Hail the CMfication of our code!!!

Anton-Kr (changes requested):
add clicky behavior

PR #1242, opened 2022-02-18
gives fleet +1 athletics and eva skill (might actually be a good idea?)

anconfuzedrock:
2 reasons:
1. athletics is probably a more logical skill for them to have than even the guns skill.
2.Marines get +1 athletics, weapons, cqc, AND eva. that's 4 skills and that's 6 points worth of them. before this fleet only had 3 worth of points in +1 eva and +1 weapons. this evens it out a bit while still having a distinction between the ranks.

### DIFF
--- a/maps/torch/torch_ranks.dm
+++ b/maps/torch/torch_ranks.dm
@@ -178,8 +178,9 @@
 	)
 
 	assistant_job = /datum/job/crew
-	min_skill = list(	SKILL_WEAPONS = SKILL_BASIC,
-						SKILL_EVA     = SKILL_BASIC)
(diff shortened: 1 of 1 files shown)


Kaksisilma:
Seems fine code-wise.

Skillpoints are in a good situation RN and we don't need anymore buffs. (If anything, we should remove the extra points Marines get) I'll leave this open until tomorrow but unless Carl wants this merged, I'm going to close it.

ProbablyCarl:
Nah, you're right on that. Other than potentially removing the skillpoint buff to Marines (kind of?), it's fine as is. I'll be closing this now.

PR #1243, opened 2022-02-18
Significantly lessens laser damage.

Kaksisilma:
Lasers are very strong right now, with both big damage and big AP values. This nerfs most beams back into sane damage values.

Open to feedback.


ProbablyCarl:
This is good. Permits us to have them standard until the rework is done. If it's _ever_ done, anyhow. I'll merge this.

PR #1256, opened 2022-02-19
Protein bars have iron, the ones in protein mres have no sugar

anconfuzedrock:
iron is good for you, it gives you blood, good in an emergency. Sugar does a couple of things, but it's bad for unathi, which compose a large enough portion of the crew that emergency supplies probably shouldn't poison them. So this replaces the sugar with iron.

edit: that and protein bars are found in the PROTEIN MRE, which unathi are supposed to be able to eat.

### DIFF
--- a/code/game/objects/items/weapons/storage/mre.dm
+++ b/code/game/objects/items/weapons/storage/mre.dm
@@ -93,7 +93,7 @@ MRE Stuff
 	icon_state = "meatmre"
 	main_meal = /obj/item/weapon/storage/mrebag/menu10
 	startswith = list(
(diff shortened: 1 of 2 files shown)


Kaksisilma:
1) Keep the sugar. They're human-made after all and not necessarily unathi-friendly.
2) Feel free to keep the current amount of iron you added.
3) Make a variant of the protein bar WITHOUT the sugar and add it to the protein MRE.

anconfuzedrock:
bwuh it's called a protein bar but ok

anconfuzedrock:
I have done as you've asked, my lord.

PR #1258, opened 2022-02-19
Replaces the old exploration ranks

Mannybrado:
Replaces or removes the old exploration ranks, as the expeditionary corps is not a thing anymore. Now, exploration uses fleet ranks, mostly petty officers.
Additionally allows the pathfinder to be a Lieutenant and removes the junior exploration rank from all of the trainee ranks, deck tech, and scientists. I didn't replace the scientist rank with anything because a scientist not being an officer is retarded.

ALSO, gives deck officer 2 more skill points to be equal with bridge officers.

### DIFF
--- a/maps/torch/job/supply_jobs.dm
+++ b/maps/torch/job/supply_jobs.dm
@@ -26,7 +26,7 @@
 	                    SKILL_PILOT       = SKILL_BASIC)
 
 	max_skill = list(   SKILL_PILOT       = SKILL_MAX)
-	skill_points = 18
+	skill_points = 20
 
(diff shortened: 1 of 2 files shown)


ProbablyCarl:
Intentionally left it as Explorer, as it's a 'sub branch'. Not too sure on this one, but it's fine otherwise.

Mannybrado:
I personally hate referring as explorers as "senior explorer" or "junior explorer" in rank. In fact, I just ignore it and call them by name, as everyone else does. It really doesn't add anything other than being somewhat frustrating when addressing an explorer seriously.

PR #1266, opened 2022-02-20
Removes GAS antags.

Kaksisilma:
We gave it some time after reverting the GAS antag ban and it didn't really work out. Balancing GAS is impossible without taking away some of their more unique features, so the best way to cull murderbugs is to remove the possibility to acquire antag statuses, which, for GAS are A) OP as fuck, B) Completely senseless from an IC standpoint, and C) Always horribly executed.

### DIFF
--- a/code/modules/mob/living/carbon/human/human_species.dm
+++ b/code/modules/mob/living/carbon/human/human_species.dm
@@ -59,6 +59,7 @@
 
 /mob/living/carbon/human/nabber/New(var/new_loc)
 	pulling_punches = 1
+	status_flags = NO_ANTAG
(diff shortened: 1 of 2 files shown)


Kaksisilma:
ok idk where to put the no_antag flag for it to work without causing CI to cum. @Yawet330 help 🥺

ShadeAware:
I support this, though I still believe GAS should have antag roles of some kind - however, it would likely require an entire concept centered around them and them only, which is outside the scope of anything any one of us really wants to do right now.

ProbablyCarl:
Looks good so far?

PR #1281, opened 2022-02-21
Update nabber.dm

Vizoplez:


Kaksisilma:
ok but what does it do?

PR #1282, opened 2022-02-21
Adds stealth module to traitor uplink, removes them from everything else.

Knightfall5:
Stealth modules are extremely powerful, yet can currently be found for free in maintenance, bought from merchants, or worst of all unlocked in the R&D lathe by simply deconstructing the 10 TC chameleon kit. 

This will remove stealth RIGs from maintenance/merchant drop tables and remove the stealth module from R&D. Cloaking can now only be acquired from a traitor uplink/Ninja.

Stealth modules will cost 130 TC, leaving 50 to spend on whatever else with a default uplink. Mercenaries will not be able to buy these because an entire squad of literal stealth ops would be amusing exactly zero times.


ProbablyCarl:
Good shit

Vizoplez:
Terrible.

PR #1284, opened 2022-02-22
[Ready] Rewrites Unathi codex stuff, makes their background selection better etc, makes their minimum age 25

anconfuzedrock:
This rewrites all the lore blurbs for unathi and sorts them out, they're a bit chunkier now, but they're made to 
A. Give you a better idea of what unathi are actually like if they're all you read
B. Be accurate to our lore (e.g.Solgov is now mentioned based on how they effect unathi here). Aside from solgov changes, though, our lore isn't changing much. Just what we put in the lore blurbs is.
The way this restructures unathi factions/culture etc (thanks for the energy deadmon) makes it more sensible if we end up changing it in the future. 

I suppose this basically rewrites veteris lore, since their lore was literally 3 sentences long. This just moves their origin to an island, as there were no "ash plains" on moghes. Their lore's the same, they're just slightly more moist now.

Also makes unathi minimum age 25 because that's adult for them, you can say "but unathi adult age varies between clans!" well human adult age varies in real life and we stick to 18 minimum. We're doing 25 minimum.


anconfuzedrock:
whoops, hope this is what was wrong and I fixed it.

anconfuzedrock:
damn bro, what's wrong with the pr?

(seriously I could use some help, I can't tell why it's failing the indent check, if there is a reason).

anconfuzedrock:
sorry to do this but I think I remember *sometimes* you can retry the checks by doing this
edit: I did this and it still fails the indentation check

ProbablyCarl:
Indentation strokes. That's about it.
Otherwise, this looks good.

anconfuzedrock:
> Indentation strokes. That's about it. Otherwise, this looks good.

please help I can't find it
edit: ok I have an idea of what it might be but I can't fix it for 7 more hours
edit edit: actually wait no that's almost definitely not it I'm still lost
edit edit edit: ok I know what's wrong now, we good.

anconfuzedrock:
should be fixed, also adds in the age thing because this is an 18+ server and I'm not having it.
happy with this as it is now

PR #1285, opened 2022-02-22
Reverts Yawet's Stealth Buffs to Monarch Serpentids

BurpleBineapple:
About five hours ago, according to the ascent_maps.dm, Yawet committed a 12 point increase to all monarch serpent job points directly to the dev branch without a PR. Because I'm not a fan of developers going behind everyone's back and buffing their choice of species, I'm making that change public and putting a pull request to revert it.

A 12 point increase for no particular reason is frankly insane. There's no reason that the monarchs should have more points than a gyne on top of the already strong pre-set job skills for Ascent. I don't care what your 'lore accuracy' is. This is terrible from a balance perspective and an extremely shitty thing to do by committing this without telling anyone.

image

No pull request, no prior alert to this, absolutely debased.

### DIFF
--- a/maps/away/ascent/ascent_jobs.dm
+++ b/maps/away/ascent/ascent_jobs.dm
@@ -174,7 +174,6 @@
 	title = "Serpentid Adjunct"
 	supervisors = "your Queen"
 	total_positions = 2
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
Cringe as fuck on their part.
Thank you. I wouldn't have caught this otherwise.

PR #1288, opened 2022-02-22
Monarch proper buffs (now that its not 5:06 AM)

Vizoplez:
Failures to note MSW's and MSQ's inherit the GAS debuff.

Also they have forced combat skills that eat 12 points anyway, please see how we handle forced skills.

Meant to PR it anyway, did a funny and hit push to master due to 5 am, unfucked

(also, as a later edit, I figured out why I did this in the first place.

Every other off-ship except skrell gets 44-48)

### DIFF
--- a/maps/away/ascent/ascent_jobs.dm
+++ b/maps/away/ascent/ascent_jobs.dm
@@ -174,6 +174,7 @@
 	title = "Serpentid Adjunct"
 	supervisors = "your Queen"
 	total_positions = 2
+	skill_points = 46 //Monarchs live longer and take longer to grow compared to Gynes and Alates.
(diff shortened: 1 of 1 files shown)


Vizoplez:
Oh yeah, for reference.

This is with the 42 point skill change.

image


You can barely take experienced in medicine and anatomy, and trained in chemistry, even if you sacrifice every other category for it.

This is just because Ascent had less skillpoints to use than things like survivor.

image


Looks the same? Right?

Thats because stranded survivor gets 44 skill points (as do vox, as do bearcat, as do unishi crew)

I'll probably just fix this at some point. Skills momento

PR #1289, opened 2022-02-22
Pulls Skrell skills more into line.

Vizoplez:
Read https://github.com/BoHBranch/BoH-Bay/pull/1288.

Essentially; Skrell and Ascent were having issues involving skill points and specialisation due to the fact they got (10 or so) less points than everyone else. 18 less in the case of the Vox in some instances.

No clue why this was the case. Carlman bad.


Kaksisilma:
SSV skillpoints are OK rn. Medic-oriented roles suffer a bit but it's not buff-worthy.

Rayfod:
Could we give them the skills instead of the points for chemistry and anatomy?

PR #1290, opened 2022-02-22
Allows GAS to be surgeons.

Vizoplez:
image

They can be almost every other medical role.

Also, this is only because both Medical tech and Lab Tech are capped at trained anatomy.

### DIFF
--- a/code/modules/culture_descriptor/culture/cultures_serpentid.dm
+++ b/code/modules/culture_descriptor/culture/cultures_serpentid.dm
@@ -59,7 +59,7 @@
 
 /decl/cultural_info/culture/nabber/a
 	name = CULTURE_NABBER_A
-	valid_jobs = list(/datum/job/chemist, /datum/job/roboticist, /datum/job/bartender, /datum/job/chef)
+	valid_jobs = list(/datum/job/chemist, /datum/job/roboticist, /datum/job/bartender, /datum/job/chef, /datum/job/senior_doctor)
 
 /decl/cultural_info/culture/nabber/a/minus
 	name = CULTURE_NABBER_AMINUS


Anton-Kr:
Please no. GAS are restricted lore wise to technician levels of training for roles. This implies that they could go to college for 6+ years and become a physiciain.

Vizoplez:
> Please no. GAS are restricted lore wise to technician levels of training for roles. This implies that they could go to college for 6+ years and become a physiciain.

The same is implied for chemistry and medicine, of which they can both take master. Lore-wise, GAS are capable of surgery if trained correctly. Up to Carl though. It just feels strange they're capable of /some/ surgery, but unable to be trained further.

anconfuzedrock:
Them being able to do *some* surgery is the error.

Anton-Kr:
>The same is implied for chemistry and medicine, of which they can both take master. 

Conceptually they are able to understand simple repetitive concepts which means procedural med TECHING or chem dispensing. Being a physician puts them on the same level as officers which is also bad.

Vizoplez:
> > The same is implied for chemistry and medicine, of which they can both take master.
> 
> Conceptually they are able to understand simple repetitive concepts which means procedural med TECHING or chem dispensing. Being a physician puts them on the same level as officers which is also bad.

Physicians can be contractors. Also, this doesn't make sense; that would imply they're enlisted, which they can't be.

Also, bartender GAS would fit the bill to be removed if this was the case. Serving a variety of drinks prepared and mixed in multiple glasses at specific temperatures would be odd.

> Them being able to do _some_ surgery is the error.

Technically false? Sealing arteries/etc makes sense. It's a repetitive action that is easy to identify and treat in theory.

Kaksisilma:
No. GAS Roboticist and GAS nurse are already pushing it in terms of lore, not to mention the PR is disliked. You can bring this up to Carl if you want, but for now I'm going to close this.

PR #1293, opened 2022-02-22
Weapon Fab access update.

Jul-Jul:
My first PR so no major changes. Simply adds Science/Research access to the weapon fab. Allows Science to access and print weapons from the weapon fab which they can make for the sake of furthering research, and supplying Infantry/Sec in emergencies. Since this is now a military vessel, this now gives Science another little purpose for them to exist as they can supply inf/sec with experimental/rare weaponry when required.


Kaksisilma:
VERY wary about this since it incentives the cringe gunsci gameplay loop. Gonna defer to @ProbablyCarl 's decision here, but again, everyone and their mother is armed in the ship now so eh.

Jul-Jul:
> VERY wary about this since it incentives the cringe gunsci gameplay loop. Gonna defer to @ProbablyCarl 's decision here, but again, everyone and their mother is armed in the ship now so eh.

Whilst I understand that this would of been a problem in the past. The weapons that fab even provides is very lacking, and like you even mentioned that nearly everyone is packing on the ship anyway. I am simply trying to find more reasons for Science to be useful, as I might add other low end smallarms to the weapon fab but make them very costly. Since you can already order weapons crates from Supply like ARs, DMRs, LMGs, etc already.

ProbablyCarl:
Weapon fab only exists in the Hard Armory as of current, and can't be accessed physically by Science without assistance. It'd be wise to provide a lathe in their department, should you think this is appropriate.
Until then, this has little to no effect as far as I'm aware, and only permits easier use for traitor Science.

Jul-Jul:
Attempted to create a clone of the weapon fab, however I will be honest, I doubt this would work. It needs to be looked over.

PR #1296, opened 2022-02-22
Revert "makes D100 have 100 sides"

Rafflesia7793:
Reverts BoHBranch/BoH-Bay#498

This breaks the D100. The code for the d100 cycles the icon state based off of the result. The d100 only has icon states for 1-10. So if the result is greater than 10, the dice switches to an icon state that does not currently exist. To fix this the d100 would need 100 icon states.

### DIFF
--- a/code/game/objects/items/weapons/dice.dm
+++ b/code/game/objects/items/weapons/dice.dm
@@ -45,7 +45,7 @@
 	name = "d100"
 	desc = "A dice with ten sides. This one is for the tens digit."
 	icon_state = "d10010"
-	sides = 100
+	sides = 10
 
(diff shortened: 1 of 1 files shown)


Rafflesia7793:
Oh boy, it looks like the previous state might've been broken too.

ProbablyCarl:
Shit's fucked, Captain.

Kaksisilma:
You have warnings in your build. This is probably because of the commits Yawet did directly to master. You'll need to rebase your changes to clear it, unless the warning comes from this PR aswell.

Anton-Kr:
I'll just make another revert PR to handle this.

PR #1297, opened 2022-02-23
Light Rig added to MechFab

Jul-Jul:
Honestly, this is me mostly testing code, and getting the grips with the spaghetti code that is SS13 development. This simply adds the Light Rig to the MechFab. The light rig is not EVA proof, and provides very low protection alike to the Null suit that the MechFab also provides. Its only slightly better then the Null suit, therefore has a research requirement, and slightly higher material cost. Again throw ya feedback at me, if this is shit then please say so. I am simply doing this to learn.


anconfuzedrock:
This might be slightly more of a balance change than it appears, as the null suits are intentionally unable to be equipped with weapons, but the light suit can be equipped with them. Robotics isn't really supposed to be a hardsuit factory. We *could* change that though.

Jul-Jul:
That is the same code for restricting the Null suit from allowing to use combat mods.

ProbablyCarl:
Category of the suit doesn't exist, causing issues.

Jul-Jul:
Removed the combat mod limitation, along with fixing my dumb ass spelling mistake.

ProbablyCarl:
Might need to pull from latest to get the fixes I made to CI.

Jul-Jul:
I believe I just did so. That should be sorted now.

Anton-Kr:
seems reasonable

PR #1301, opened 2022-02-23
Removing CorpoSec

Jul-Jul:
God help me, I am prepared for to be ass fisted for this. This PR removes CorpoSec from the Security. Considering our current lore regarding the Corpo War, and that fact that the Dagon is now a military vessel, it does not make sense that we would have Corpos in one of the most important jobs on the ship. Especially considering why would Sol Command hire expensive Corpo Guards when they have a supply of loyal, cheap military personnel to fill such roles.


Kaksisilma:
Fucks over too many corposec characters for an arguably arbitrary reason, so I'm going to close this. You're free to ask carl to re-open it and possibly merge it if it aligns with his view.

PR #1302, opened 2022-02-23
Purchasable KOSMAG Slugs

InfinitelyThinRectangles:
Hallo,

Spotted someone mention as a ghost that you can't purchase KOSMAG slugs, and I'm fairly sure that's just because they weren't added to the supply menu when they were implemented. This change makes them purchasable at the same cost as any other slug, as well as makes the name of the supply packs they come in a little nicer to read and comments out (NOT deletes) the ability to order ammunition for ship weapons that physically exist on the Dagon but can't be fired due to true Ship to Ship combat not being implemented.

Should be an inoffensive enough change! :)

### DIFF
--- a/code/datums/supplypacks/munitions.dm
+++ b/code/datums/supplypacks/munitions.dm
@@ -1,12 +1,12 @@
 
-//The weapons that fire this charge aren't functional and they don't have any other use, so there's no reason for them to be accessible until that changes. - ITR//
 
 
-///decl/hierarchy/supply_pack/munition
-//	name = "Ship Munitions"
-//	containertype = /obj/structure/largecrate
-//	containername = "mass driver munition crate"
+/decl/hierarchy/supply_pack/munition
+	name = "Ship Munitions"
+	containertype = /obj/structure/largecrate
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
you commented out the parent

InfinitelyThinRectangles:
Fixed. I'm really curious about the "ap_slug" and "md_slug" items. 

image

There's not a whole lot about them. I also see that both the INDARRA and the KOSMAG are both OVERMAP_WEAKNESS_EXPLOSIVE. Does this mean they have the same effect? The INDARRA and KOSMAG charges seem to have a different method of calculating their effect, the INDARRA uses this:

image


While the KOSMAG does this:

image


Which seems to be the same except it also manually creates an explosion?

ProbablyCarl:
> 

They have identical use overmap wise, but the KOSMAG charge is specifically for S2S and shouldn't be wasted, as it's far better for that.

The slugs, however, that seemingly have no use, are leftovers from Bay itself, IIRC.
They'll likely be removed with the new system coming in.

PR #1292, opened 2022-02-22
Expands the /me window into a full paragraph.

ShadeAware:
Does what it says on the tin. Rejoice, ye paragraph posters.

This took way too fucking long because the window is handled by the **typing indicator function.**
Kill me.

PR #1308, opened 2022-02-24
/me Window Expansion Attempt 3

ShadeAware:
See #1292

### DIFF
--- a/code/modules/mob/say.dm
+++ b/code/modules/mob/say.dm
@@ -12,7 +12,7 @@
 	remove_typing_indicator()
 	usr.say(message)
 
-/mob/verb/me_verb(message as text)
+/mob/verb/me_verb(message as message)
 	set name = "Me"
 	set category = "IC"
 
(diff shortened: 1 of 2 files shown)


Anton-Kr:
Were you intending for this to have multiline support? Right now everything gets reverted into a single blob. Further while typing in `me` works, typing in` me "` without an emote brings back up the single line interface.

ShadeAware:
> Were you intending for this to have multiline support? Right now everything gets reverted into a single blob. Further while typing in `me` works, typing in` me "` without an emote brings back up the single line interface.

@Anton-Kr No, this just expands the window. Please re-read the purpose of this PR.
As far as the single line interface goes, typing in things is a different method than the verbs. The things I have edited are the only plausible mentions of the emote verb and function I could find; if you want the paragraph interface, use hotkeys **(the /me hotkey works regardless of hotkey settings)**. BYOND text command inputting is not supported and likely cannot be supported or edited.

Anton-Kr:
Consider this a test merge. If there is substantial desire to revert this I likely will.

PR #1310, opened 2022-02-25
Makes Sword Sheathes wearable as Outerwear

InfinitelyThinRectangles:
Hiya,

This one simply enables the sword sheathe to be worn as outerwear in its own right, since right now the places you can wear it are really awkward and often taken up by other equipment. This in no way adjusts them, adds more of them, makes them more accessible or anything of the sort, purely makes it a little easier for those who already have access to a limited number of them to carry it.

I'm still not entirely happy with how this works right now and will probably try to rework it into a holster accessory for innerwear clothing, same as how you can get holsters for firearms, but this is a start on making the things a bit less clunky. It seems strange to me that they're so restricted in terms of where you can put them, especially when other really nasty melee weapons like the machete or surgical saw that are of a broadly similar size fit in a bag with no issues:

image

Whole thing with the sheathe just feels like it's there for it's own sake.

Recycles the mob overlay for when they're on a belt for when they're in the outerwear slot.

Also fixes a typo in the comment.


N8-Toe:
i think it is because the sword is meant to be an item for dress wear, not used as an actual weapon like people have been.

Anton-Kr:
Any chance you can make the sheath itself attachable? Might be unable to with how accessories work, but still.

Anton-Kr (changes requested):
Not too happy with it being in oversuit, given that no other storage item exists there to my knowledge and that it would conflict with dress uniforms. Best option forward if you want to continue with PR would likely be to separate the following items in the uniform vendor:
- Sword itself without sheath
- Current sheath
- Uniform sheath preferably with a modified sprite to distinguish it from current sheath.

anconfuzedrock:
The machete and saw are actually significantly worse than the sword IIRC. sword's nasty, but it's really not meant to be carried with your gear, it's for funerals and shit, you're gaming if you use it with anything else in your belt.

Sparrowbird611:
I would advise against doing this as outerwear. I don't really care if you use it to gank people, that's the purpose of a sword and if you can access a sword you probably have access to half decent guns minimum, but it's very weird to handle this as an outerwear item. I'd recommend putting a bit of time into making it a holster item instead, or alternatively making a variant of a regular belt that can hold a sword, such as a sheathe belt similar to the machete and batons.

Pariah919:
Personally think the sword being forced on belt/back is okay balance/theme. personally.

Though I think suit storage needs to be reworked since its kinda stinky.

InfinitelyThinRectangles:
Hiya,

Sorry about the tardyness, real life came knocking for a little while.

I'm going to take Anton's advice and add the following to vendors:

    Sword itself without sheath
    Current sheath
    Uniform sheath preferably with a modified sprite to distinguish it from current sheath.

I'll do that in a separate PR to keep the decks clear while I work on this, feedback much appreciated and honestly I think I agree.

See you all next time!

PR #1312, opened 2022-02-25
Adds Tasers to Security's Kit Options

BurpleBineapple:
A simple addition after asking over on the Discord's suggestions page. This PR aims to add non-lethal, low capacity tasers as an option for security's equipment kits and provides the COS one with their 15mm magnum.

The reason I think this is necessary is that at the moment, security has painfully bad non-lethal options at a range. Weapons like rubbers and beanbag shotguns still land people in the medbay more often than not. Infantry of all departments has tasers and are more effective at taking people down non-lethally than security because of that.

To avoid complaints about security having lasers, these tasers are strictly non-lethal. They only have stun and shock settings with no way to be set to lethal. Smartguns aren't going to be made irrelevant because of this.

:cl:
tweak: Security's equipment kits now have a non-lethal taser as an option. The Chief of Security's locker also spawns with one due to them not having kits.
/:cl:


ProbablyCarl:
The lack of stun weapons is intentional. The return of them to Infantry was done by someone in a speedmerge. I'm sure you can guess who. I'll be reverting that now, regardless.

We'll take a look at this momentarily.

PR #1316, opened 2022-02-25
Tricord Tweak

InfinitelyThinRectangles:
Hi hi,

As discussed in VC, this change removes the requirement for a phoron catalyst when mixing tricord, halves its effectiveness in overall healing and prevents it from mixing inside a body by requiring a temperature between 50 and 100 Celsius. 

I didn't see an easy way to just prevent Inaprov and Dylo from combining if and only if they're in a body, but the temperature requirement does the same thing!


ProbablyCarl:
Closing this due to internal discussion. We're leaving Tricord as is.

PR #1317, opened 2022-02-25
fixes sauna shutters being sorta-open

anconfuzedrock:
Previously sauna shutters started open, but obscuring view. This was pretty confusing, Sauna has like 3 seperate ways to obscure vision with the windows and that did not help. a bug in any case.

### DIFF
--- a/maps/torch/torch3_deck3.dmm
+++ b/maps/torch/torch3_deck3.dmm
@@ -3954,7 +3954,8 @@
 	density = 0;
 	icon_state = "shutter0";
 	id_tag = "saunashutters";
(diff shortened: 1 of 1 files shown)


Anton-Kr:
Can you confirm opacity is restored on closing?

anconfuzedrock:
> Can you confirm opacity is restored on closing?

5000% sure. this is the same way other shutters work. Hell, before I did this pr you could workaround it by closing and reopening the shutters.
(even if it didn't restore opacity it wouldn't matter because you can also close sauna curtains and tint windows for the same opacity

PR #1318, opened 2022-02-25
readds arm and leg guards as a sec thing

anconfuzedrock:
Arm and leg guards were for sec roles _only_, which was fine because sec actually has a vending machine that vends arm/leg guards. It's just a convenient timesave for them.
Nevermind on adding medium plates, that's more of a balance thing, out of scope.


Jul-Jul:
Thank you, because my britbong wording confused Carl.

anconfuzedrock:
It breaks my heart to tell you this, but considering security only has 6 spare medium plates, that's more a balance change and as I didn't ask carl about that I'm not putting it in this pr.

PR #1322, opened 2022-02-26
Gives Basic EVA and Medical to SolGov Fleet characters

Ctr-Truescreen:
Makes sure that no fleetie makes it onto the Dagon without passing their EVA and First-Aid training.

image


ProbablyCarl:
based

PR #1327, opened 2022-02-26
High-vis Jacket suit storage oversight fix.

Ctr-Truescreen:
Gives the high visibility jacket's suit storage the same permissions as the hazard vest.

Not tested on localhost, but compiles fine and seeing as this is a simple line change, it should be okay.


Anton-Kr:
>Not tested on localhost

famous last words

PR #1329, opened 2022-02-26
SGR + Guard fixes and edits.

Jul-Jul:
This PR allows for SGR to have a larger range of civilian clothing, whilst also adding the (Chad) Marine Embassy Guard to the SolGov Guard.


Anton-Kr:
why not

PR #1331, opened 2022-02-26
Changes to the sec revolver to distinguish it from the pistol / buff it (more ap and accuracy mostly)

anconfuzedrock:
This gives the revolver: 
-an intermediate 12mm cartridge that has 10 ap and has slightly less damage falloff, but is less likely to cause internal bleeding/shrapnel. So a bit more reliable and consistent
-increased shortrange accuracy.
-MUCH more fire delay, 50% more fire delay than the pistol
-a more distinct sound from the pistol.

All this should shake out to making the revolver more consistent than the military pistol, while the pistol is more spammy and scary to be hit with because of the shrapnel/internal bleeding. The pistol can shoot faster, and 15 bullets with 0 ap is still going to do more damage through armor than 6 bullets with 10 ap, but the accuracy should make the revolver fun to use in its own way. willing to shuffle these values around as needed.


anconfuzedrock:
Hang on, I may have done this in a very stupid way, maybe I'll end up redoing the whole pr in a sec

PR #1334, opened 2022-02-27
Changes the revolver to make it more distinct from the pistol / less ass (mostly a buff)

anconfuzedrock:
This was made with the thought in mind that having only 6 bullets as opposed to 15 in its first clip and every spare ammo thing making it far inferior to the pistol.
How it's different:
+The revolver has +1 extra accuracy and -0.5 distance falloff, meaning it does slightly more damage at range and is more accurate at range (equivalent to your target being one tile closer)
+the revolver's shots have +10 armor penetration, meaning its shots have +10 armor penetration
-the revolver is WAYYY slower. 14 delay compared to the pistol's delay of 8 means you can spam the pistol way harder. 

Also changes the unused penetration multiplier to a modifier so that it works (better overall, really) as well as adds a damage falloff modifier.

### DIFF
--- a/code/modules/boh_misc/firearms.dm
+++ b/code/modules/boh_misc/firearms.dm
@@ -51,9 +51,11 @@
 	ammo_type = /obj/item/ammo_casing/pistol
 	fire_sound = 'sound/weapons/gunshot/revolver_small.ogg'
 	desc = "The Lumoco Arms' Mk12 is a rugged revolver for people who don't keep their guns well-maintained. Unlike its cousin, the Mk59 'Jhen Bothus', it has no issues with reliability."
-	accuracy = 1
+	accuracy = 2
 	bulk = 0
-	fire_delay = 9
+	fire_delay = 14
+	penetration_mod = 10
+	falloff_mod = -0.5
 
 /obj/item/weapon/gun/projectile/revolver/medium/sec/pepper
 	ammo_type = /obj/item/ammo_casing/pistol/rubber/pepperball
--- a/code/modules/projectiles/gun.dm
+++ b/code/modules/projectiles/gun.dm
@@ -90,8 +90,10 @@
 	var/combustion
 	//damage multiplier. Multiplies damage. 0 does nothing, by the way. Higher means more damage, lower less.
 	var/damage_mult = 1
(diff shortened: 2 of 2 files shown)


code/modules/boh_misc/firearms.dm
```diff
@@ -51,9 +51,11 @@
 	ammo_type = /obj/item/ammo_casing/pistol
 	fire_sound = 'sound/weapons/gunshot/revolver_small.ogg'
 	desc = "The Lumoco Arms' Mk12 is a rugged revolver for people who don't keep their guns well-maintained. Unlike its cousin, the Mk59 'Jhen Bothus', it has no issues with reliability."
-	accuracy = 1
+	accuracy = 2
 	bulk = 0
-	fire_delay = 9
+	fire_delay = 14
+	penetration_mod = 10
```
BurpleBineapple, line 57: I would recommend testing how this works with rubbers, because unlike regular pistol rounds rubbers have slight AP to properly apply agony though armor.

I expect that such a large penetration buff will have some unintended consequences on how lethal rubbers are for revolvers. Possibly even more lethal than FMJ on armor.

anconfuzedrock: I'm pretty sure you don't know how armor pen works, but here's the thing, I don't either, so I'll check just for u

anconfuzedrock: I tested it, very simply, did 6 shots with the revolver vs 6 shots with the pistol at point blank (which I believe increases dmg) with both guns, with rubbers. Both lead to moderaate blunt trauma but no internal bleeding or fractures or anything. I'm pretty sure armor pen has no effect on unarmored targets. on armored targets I also tested it, same result, they don't do enough damage to be way too lethal (though the revolver has less shots so it's less likely to annihilate someone)
It's not meant to be a large buff, this still isn't enough armor pen to break through light plates lol. It's just meant to make it slightly punchier than the pistol, but buffing its damage would've made it really unfun to get shot with if you're unarmored.

Anton-Kr:
I don't super like same bullits doing different shit in different guns, but we can chalk it up to a "longer barrel" and go from there. I don't see this being too significant of a change.

PR #1337, opened 2022-02-27
swaps  bonus point of eva skill for fleet with sci skill

anconfuzedrock:
I didn't want to do this, but there's a bug I'm not sure how would be best to fix (maybe some other time) where because these skills are applied first, a job that gives you a lower level of eva (e.g. most jobs) overrides the higher level eva, meaning the trained isn't applied to a lot of jobs. and even with that they had 1 point less worth of skills than marines. sci skill costs 2, so it evens it out, and it's a thing? can fix it later but here's an alternative to the fucky wucky behaviour for now.


anconfuzedrock:
I am not a clever man, but I found out how to fix the bug, and I'd rather do that actually.

PR #1339, opened 2022-02-27
SGR & guard edits

Jul-Jul:
This hopefully fixes the SGR clothing selection. Also increase SGR guard skill points since they're supposed to be veteran/specialised individuals for protecting diplomatic individuals and VIPs.


maps/torch/job/command_jobs.dm
```diff
@@ -449,7 +449,7 @@
 		"Sol Federal Assistant",
 		"Sol Federal Chaperone",
 		"Marine Embassy Guard")
-	skill_points = 20
+	skill_points = 32
```
Anton-Kr: This is far too excessive. Reduce this to 24, and we'll go with that. Otherwise this will encourage far too much shit behavior that we already saw one round.

Anton-Kr (changes requested):
Too many guard points.

Anton-Kr (approved):
thanks!

PR #1340, opened 2022-02-27
Improvements to the Body Marking System and Hair Fades

BurpleBineapple:
Several ports from baystation12's improvements to the body marking system, mainly from:
https://github.com/Baystation12/Baystation12/pull/30270

Also guts the weird color_blend_mode garbage from Mithra code. No idea why that exists. It does absolutely nothing.

Body markings can now be assigned by draw_order to layer properly. Hair fades will work with any hair and do not clip. This PR is going to remain a draft however until I can fix the Mithra icons, which at the moment are not compatible with the change in how markings are applied.

Other than that, this should be functioning fine.

Examples of the hair fades:
image
image
image

Several different styles exist, and they should work with all hairs.

To do:
- Recolour the mithra accessories so that they work properly with the new code.


code/modules/client/preferences.dm
```diff
@@ -261,8 +261,8 @@ datum/preferences
 
 		for(var/BP in mark_datum.body_parts)
 			var/obj/item/organ/external/O = character.organs_by_name[BP]
-			if(O)
-				O.markings[M] = list("color" = mark_color, "datum" = mark_datum)
+			if (O)
```
BurpleBineapple:
```suggestion
			if(O)
```

BurpleBineapple:
After testing with the icons, this will take some time as I will manually have to change them all from white to black icons. Not difficult, just time consuming.

BurpleBineapple:
Pull request isn't stale, just taking time.

BurpleBineapple:
Still being worked on, albeit slowly.

BurpleBineapple:
Closing this for now. Not being activity worked on.

PR #1341, opened 2022-02-27
Fixes a Skill Issue

anconfuzedrock:
Previously if your job skills were lower than your branch skills but existed (e.g. paramedics got basic eva) you would get the lower job value instead of the higher branch value. On account of this being fucking stupid, this is changed. Now you get whatever skill value is higher.

image


Anton-Kr:
I fucking love baycode

PR #1342, opened 2022-02-27
SGR Guard locker fix,

Jul-Jul:
Carl removed the SGR's guard access to their own locker. Simply adding it back. Also, I am not sure why SGR and guard have had most of their access rights removed when they were near complete clones of the CL and guard's access rights.


Anton-Kr:
speedmerge

PR #1343, opened 2022-02-28
The giant storage upgrade for skrells.

Rayfod:
This makes the weapons of the skrell able to be put into the belt like most other weapons.
The pistol is able to be put into the holster, so they have a trusty secondary.


Kaksisilma:
The pistol is a subtype of the nuclear gun, which is why it doesn't fit ordinary holsters but can specifically fit the skrellian holster belts. (You don't lose storage space by holstering it there btw. You can fit the same amount of items with or without a gun.)

I'm fine with the QX-2 being carried around in a belt but adding the ZT-8 there feels a bit like a stretch. Skrell have strong waists and swimming muscles but they're not that strong.

Anton-Kr:
Looks fine. Was gonna mirror above change requests as well, but this seems resolved.

PR #1347, opened 2022-02-28
Reverts the Charge Boost to Vox Flux Cannon

BurpleBineapple:
After a recent round, it's clear that whoever made this change did not consider how recharging works.

The current issue with the Flux Cannon is that it's a weapon that fits in a satchel and has the power of a laser cannon in a mode that can be fired almost ever 3-5 seconds without needing to be charged. While yes, the original PR's intention was to increase how many shots it has, it also increased the rate at which it charges. It's a little less than optimal to have a gun that can be fired almost forever and instantly stop people's hearts.

Other than that, I think the Vox situation is fine. This was the only major meme.

### DIFF
--- a/code/modules/projectiles/guns/energy/vox.dm
+++ b/code/modules/projectiles/guns/energy/vox.dm
@@ -30,9 +30,9 @@
 	one_hand_penalty = 2 //a little bulky
 	self_recharge = 1
 	firemodes = list(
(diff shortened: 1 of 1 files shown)


Rayfod:
If you need to change the recharge time, add "recharge_time = 8" to double the standard "recharge_time = 4".
Changing the number of shots they can shoot makes the weapon very bad.
Or you can add "recharge_time = 12" to have it three times the original value if you feel the need for it.

BurpleBineapple:
Closing this after talking with Rayford in discord, he's going to tweak the recharge rate in his own PR.

PR #1346, opened 2022-02-28
Vox nerfs the big umpf

Rayfod:
Vox nerfs: Kosmag removed, Armalis changed to one, and the sniper rifle nerfed so that it **cannot** stun anymore while giving it back its original ammo.
Nerfed the dark matter gun, to charge a focused shot, every 12 seconds.
These changes were talked about with AntonKR before they were made.
Final confirmation from Carl would be good.
Let's see how this goes.

PR #1349, opened 2022-02-28
The best Vox nerf PR yet

Knightfall5:
Drastically reduces the volume of Vox jump packs, they now carry around 45% more gas than a Human jetpack instead of around 1800% more. Seriously they can just fly around forever without running out.

Brings down the flux cannon to be slightly better than a regular energy rifle, and self-charging. As it is right now it penetrates any set of armor in the game, has a damage flag that makes your organs melt even faster than a regular gun, is hitscan, fires rapidly, and self-recharges. Even if you removed the self-charging completely it is a broken gun. Also tones down the eight shot secondary burst fire mode damage since even THAT was stronger than a laser rifle. It still penetrates Infantry RIGs and security voidsuits so no need to worry about getting face tanked by them. Ablative armor on the Dagon, however, not so much.

Update: Consolidates https://github.com/BoHBranch/BoH-Bay/pull/1346 into this PR, which increases recharge time on the Flux Cannon since it was way too low, changes Armalis slots to 1, and removes the stun from the self-recharging Vox laser DMR.


ProbablyCarl:
The packs having so much gas was intentional.
The Skrell and Ascent have infinite gas for their packs. Vox do not.
The Vox are intended to be a relatively painful enemy group to deal with, and the pack is fine as is.

Anton-Kr:
Following discussion, all of this is going to be a no, EXCEPT for the Armalis slot reduction. I think the issue is overall with the number of Vox and how difficult an Armalis is to deal with, so we will test how it plays out with a single Armalis. Mind you, other Vox nerfs will be forthcoming in both a nerf (but not removal) to the bodyshielding mechanic, as well as a significant reduction of their OFD s2s abilities. Thank you!

Anton-Kr (changes requested):
Remove all except Armalis slot reduction.

Knightfall5:
So we are just gonna leave in the... ah yes, 50 damage 65 pen rapid fire 4 second recharge hitscan laser, and the self-recharging 10 second stun DMR? Might as well not nerf them at all then since balance isn't a concern!

PR #1351, opened 2022-02-28
fixes 2 more maint doors

anconfuzedrock:
These silly maint doors required the wrong access, blocking off half of d1 maint. this fixes it.
image

### DIFF
--- a/maps/torch/torch5_deck1.dmm
+++ b/maps/torch/torch5_deck1.dmm
@@ -7064,8 +7064,11 @@
 	d2 = 8;
 	icon_state = "0-8"
 	},
-/obj/machinery/door/airlock/maintenance,
 /obj/machinery/door/firedoor,
+/obj/machinery/door/airlock/maintenance{
+	autoset_access = 0;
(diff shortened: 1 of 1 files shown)


Anton-Kr:
Any chance you can unbolt the two randomly bolted doors around dorms on D3? Pleeeease

anconfuzedrock:
just for you sweaty, hang on a sec.

anconfuzedrock:
map merger is fucking with me so that'll have to be another pr

Anton-Kr (approved):

ProbablyCarl:
Dorms airlocks were locked to prevent unauthorized access (anything not maintenance work that can unbolt doors) from running side-to-side. This allows Security to actually contain a threat inside, should such a thing be required, as Dorms were intended, initially at least, and still in spirit, for antag purchases. The same is extended to the sauna.

As for the doors next to gas runoff, that's actually my bad. Apologies.

In any case, both can be changed without issue.

Anton-Kr:
Very well, disregard my comment requesting that additional change in that case.

PR #1354, opened 2022-02-28
Tablets on Hand, my Beloved

AniciaBelle:
Add sprites to the tablets (normal and command) when having them in your hand. More immersion, tablets are very useful ?)

image
image


Kaksisilma (approved):
Couple of CI fails but they seem to be unrelated. Looks good to go.

PR #1366, opened 2022-03-01
Powercreeps the Rawl

anconfuzedrock:
Well, more accurately tweaks it, but here's the biggest balance change:
-Energy knife has a parry of 10, because combat knives have 15 but this thing's more unwieldy.
-The energy knife's damage is reduced to 20, down from 35. That was a placeholder till I got the next addition in, which iiis:
-Gives them a war pick as well so they have another melee option. It's actually weaker than the energy knife was, doing 30 damage instead, **its attack delay is 60% longer than that of most weapons**, and its parry is still worse than the esword or officer sword (and dmg is worse than the officer sword against unarmored) but 50 pen is 50 pen, 20 parry is pretty good, and the +5 accuracy bonus is rare (though officer swords have 10).  That and it doesn't have a blade so it doesn't leave as nasty of cuts. On the bright side, it can cut down door bolts like a welding tool can for "Salvaging", albeit slightly slower.
**Thanks to Pawn for the sprites!** 
(in order to make it not edged I changed melee energy code a bit, doesn't cause any issues, just new optional vars so you can make non-sharp energy weapons.)

The Rawl voidsuits have been buffed a bit, on account of them sucking compared to hestia's infantry rigs and sec voidsuits. They're not supposed to be great, so the new ones are still a bit worse than the sec voidsuit (the heavy one is about on par) but  this way they're not super squishy with their limited medical facilities.

Veteris can now also join the rawl.

-Instead of some hobo rags for clothes, the rawl crew now has "custom" Dohruk flightsuits, practically just brown versions of existing flightsuits (and literally, you can still spot a faded NT logo). If for some ungodly reason you want to wear it as a shuttle pilot, it's also an option.

-the spawn cost is increased from 0.2 to 0.5, Idk why I didn't have it at 0.5 in the first place, it must've been for testing? It's not that cheap.

-The pilot/captain/jhoge now starts with some bureaucracy to help distinguish them.

-Buffs the pilot's ham radio


Anton-Kr:
Seems reasonable!

Anton-Kr (approved):

PR #1374, opened 2022-03-01
Changes the Envoy's and KS-40's shell capacities back to their original ones

DriedCrust:
Likely some legacy Hestia update that happened for some reason or another. This fixes this.


Spummm (commented):
e

Kaksisilma (approved):
Looks OK.

PR #1377, opened 2022-03-01
Fixes Bearcat's CI by reverting the Rework.

Kaksisilma:
Not a massively popular PR in the first place and one that came with a myriad of issues. Let's see how this goes.


Kaksisilma:
This fixed ALMOST EVERY SINGLE ERROR!

Who would guess?

PR #1378, opened 2022-03-01
Fixes ascent ci

Kaksisilma:


Kaksisilma:
ow oof my branch is fucked brb

PR #1380, opened 2022-03-01
Removes a 1.1 mod to Ascent alate hardsuits

DriedCrust:
What in the god damn. Why is this here.


Anton-Kr:
This is merged once CI passes less because of the balance change and more because of how godawful of a coding practice this is that actively breaks examine text.

Anton-Kr (approved):

PR #1381, opened 2022-03-01
All my vox balance changes in one place

Rayfod:
The Darkmatter cannon is acting a bit stranger with the DAM_DISPERSED trait, so I don't know if we should remove it, or what we should do there. 
The recharge time is the same as with a normal AEG.
**The Sniper was changed to "only" have a "stutter", which makes people walk randomly. The original 4 shots were added back in, with a longer recharge time between shots.**


Anton-Kr:
Title changed because these aren’t really bug fixes.

Rayfod:
Opening this for it being ready for after the 24-hour delay for balance PRs

PR #1383, opened 2022-03-01
Stealth RIG Module

Kaksisilma:
Only the Ninja gets stealth modules now. While it is definitely a very interesting RIG augment, it presents very little counter play when readily available to other roles in the form of random maintenance loot or to antagonists through uplink purchases.

This will give us time to think about a solution while still ensuring Ninja is a viable antagonist.


Anton-Kr:
@ProbablyCarl seems like a solid change. Any issues with it?

ProbablyCarl:
No. If we can find some manner of reworking it back in later, great. I'm just happy we're keeping it with Ninja, for uniqueness if nothing else.

Anton-Kr (approved):

PR #1387, opened 2022-03-01
Removes some runtimes from away sites testings

Kaksisilma:


Kaksisilma:
Remaining runtimes are on the Voxship.

Anton-Kr (approved):

PR #1388, opened 2022-03-01
Makes FBP Charge relevant

Vizoplez:
FBP/IPC charge crit is now based off of their actual charge, versus organ damage

For some reason, it was using the damage to determine if the organ was dead, which stopped it from discharging entirely.

Now, this was an issue, as the organ_broken for IPCs was 15, resulting in them getting the incorrect message + causing a balance hell.

This simply moves it to be based on their charge, so you don't get situations where they have 2000+ charge but are still in crit because they have had exactly 1 more damage to their micro than 15

This now also makes microbattery damage increase power usage (shorting mechanic), encouraging repairs lest ye collapse.

Open to suggestion.


Anton-Kr (approved):
Interesting change. Lets give it a go. Admittingly not super familiar with IPCs but the reasoning seems reasonable.

PR #1395, opened 2022-03-02
The Last(tm) Vox ship update!

Armyguy:
This is mostly just a QoL pass and one last try to fix the lasers. And removing the template flag.


Armyguy:
CI does not like atmo memes I guess

Anton-Kr (approved):

PR #1401, opened 2022-03-03
Vox mindslave implant

Rayfod:
Vox mind slave implants are organic and can get people captured instead of killing them.
A lockbox was made with a few implants in it.
A new sprite for when you take the implant outside of a person, it wiggles!
Funky code, that is bad but works.
You can see the laws in your notes if you ever get confused.

"You are the property of the voxes."
"You are worth a lot as a property, so make sure that you stay away from harm."
"You are not allowed to use any device that is considered dangerous. This may include using weapons, delaming a supermatter, or activating a bomb, as an example. "
"If you are sold to someone, then you are to treat them as your new owner."

Skærmbillede 2022-03-03 221546

**I will add the lockbox in a separate PR**


ProbablyCarl:
Map merge this. Use TGM if needbe.
Don't want it like it is currently.

Rayfod:
**Do wait a bit to merge it**

Anton-Kr (changes requested):
Since the implant will seriously impact a person's round and may put them in a position of feeling uncomfortable with continuing to play in the round, please add an option for the host to reject the implant and face the consequences of the Vox knowing that the implant was rejected.

Rayfod:
**Do not merge. Let me see if this works**

Rayfod:
I will just make a new PR, for this is useless to work with right now

PR #1413, opened 2022-03-04
Sec Primaries | Stuff

ProbablyCarl:
- - -
Balance:
 - Security Officers, not Cadets, given access to a 'primary' gun box. This gunbox contains three choices, giving the selection of a 'new' SMG, the beanbag shotgun and the taser carbine.
 - Provides both SMGs, old and new, with automatic fire. This applies to the Sword and Pariah.
 - 'New' SMG given a restriction to a new cartridge type, but has a faster fire rate to compensate.
 - Newer cartridge is a much weaker version of the rifle frangibles, intended for disabling people in the same manner birdshot does. Will absolutely kill someone without armour, but that's not too heavily a concern.
 - Guarantees shrapnel on birdshot.
- - -


anconfuzedrock:
image

PR #1419, opened 2022-03-05
Xeno Gamemode Hotfixes

ProbablyCarl:
- - -
Bugfix:
 - Life wasn't processed for some reason regarding larva, and I imagine it was to do with the file I had enabled. It was previously unloaded, now it's not. Lets see if this fixes it.
- - -
Changes:
 - Infestation made exclusively Xenophage.
- - -


ProbablyCarl:
Fixes it locally. Speedmerging this. Fails on the check aren't related.

PR #1421, opened 2022-03-05
Nerfs swords, buffs machetes

Anton-Kr:
The swords were overwhelming for the amount of force they were given for essentially free. They shouldn't be an outright upgrade to machetes. This nerfs them to a more appropriate (but still useful) force, and gives machetes the game intended (unless I am reading code wrong) force.


anconfuzedrock:
note the other advantages of sword, most notably having a rare 10 melee accuracy bonus.

InfinitelyThinRectangles:
Just a thought, but when I wanted to make swords easier to carry and fit in bags like machetes can, I was told that part of the reason it was okay that they're difficult to wear was how powerful they are. If they're no longer going to be so strong, I'd be really appreciative if they could fit into bags like the very machete they're now functionally a reskin of, since the entire reason they could not has fallen away.

Anton-Kr:
@InfinitelyThinRectangles they are still at the strength of pathfinder's machete, with an accuracy bonus. The expense here is obtainability and inventory clunkyness.

anconfuzedrock:
that and +10 armor penetration on the sword.

ProbablyCarl:
Good to merge?

PR #1424, opened 2022-03-05
Re-enables Four Species

ProbablyCarl:
- - -
Currently, Vulpkanin, Vasilissan, Sergal and Akula were playable through abuse of genemodding. Because of this, I've gone ahead and just reenabled them outright, since they already have stuff 'written' for them, to explain their presence.
Sergals and Vasilissan are a tough sell, but after some debating I've decided to include them too.

This has been discussed internally, and likely to be speedmerged™.
- - -


Mannybrado:
lol

PR #1428, opened 2022-03-06
Removes Infantry Polyp Access

N8-Toe:
at request of staff member HighAdmiral. Polyp access revoked for all infantry roles


N8-Toe:
fucked up, hold on

PR #1434, opened 2022-03-07
Raises Exoplanet Count | Raises Off-Site Budget

ProbablyCarl:
- - -
Balance:
 - Exoplanet count raised by two, bringing it up to three. Previously, it sat at one, not leaving room for variety in our semi-lengthy rounds. This is likely to heavily impact performance; as such, we'll test merge it first.
 - Away site budget raised from eight to twelve. If you don't know what this does? Don't worry about it.
- - -


N8-Toe (approved):
Looks good, just a simple number change. Will have to just keep an eye on performance

PR #1437, opened 2022-03-07
Do you have any idea how much this gear costs, son? - Infantry Edition

BurpleBineapple:
Adds Azlan's new sprites commissioned by Quill as the infantry fighting hardsuit. It shares the exact same armor values and modules as the original infantry hardsuit for now, with a slightly different version for the SL for the sake of identification.

Regular Infantry 
image

Squad Leader (And other species)
image

Comes with its own unique thruster sprites
image

In a future PR, I will be adding hardsuit weapons that can be selected for this suit as sprites for a mounted bulldog, finger laser, and mounted TVP were also provided. For now, these are just the new suits to replace the old.


N8-Toe (approved):
Sprite work is great, code and implementation looks good as well, approved

MichaelMaCleod007:
I do not care for the color, seems to bright and reflective for an infantry unit and you may as well remove the helmet lights as they don't even show up on the helmet.

PR #1439, opened 2022-03-07
Nerfs Makeshift Guns Somewhat

Anton-Kr:
Rifle less straight up upgrade
Shotgun no longer fits in backpacks
Removes unused comments.


code/modules/projectiles/guns/projectile/zipgun.dm
```diff
@@ -120,11 +119,11 @@
 	accuracy_power = 5
 	has_safety = FALSE
 	bulk = GUN_BULK_RIFLE
+	jam_chance = 40
```
ShadeAware, line 122: In favor of the rest of the nerfs, but hard no on the jam chance. None of these weapons were supposed to jam, and adding this essentially invalidates it as something anyone should use.

PR #1450, opened 2022-03-08
Fixes Infantry Suits for Xenos

BurpleBineapple:
Quick bugfix. Modular mirtha has some weird override stuff that prevented species who should be able to wear the infantry suit from wearing it. This fixes that.


N8-Toe (approved):
Code looks good!

PR #1461, opened 2022-03-09
Electrolaser carbine changes

Pariah919:
Electrolaser has a higher fire delay, up to 1 second between shots. A .4 second nerf.

The Carbine has a very high damage, and ability to down people. I think this is fine aslong as the firedelay is low to differentiate it from the taser outside of big boy damage.


Anton-Kr:
Code looks good and I agree with the balance change. I'll leave it up to Carl to merge.

Anton-Kr (approved):

PR #1464, opened 2022-03-10
Lets Skrell, Unathi, and Ascent dock properly with the Dagon, as well as making their landing zones exclusive!

anconfuzedrock:
Fixes:
Skipjack can now use its intended docking port, thanks to the shipen- I mean, uh, comically large docking port. This also allows the Skrell to use this port, which they were intended to be able to use.
The Rawl no longer has a "ghost landing site" show up. It was dragging around the original landing zone but unable to go back to it, this is fixed now.
Whoever  designed the airlocks of the lower ascent shuttle screwed up hard. It's way less spaghetti now. If it has any other problems,
Features:
Functional docking for the skrell and ascent shuttles! The Ascent shuttles previously didn't have docking ports, but now they do- might give the upper ascent shuttle the ninja docking port later, but right now it has one port, and the lower ascent ship has two ports, the same ones the skrell ship has. They each only have *automatic* docking at one of the ports, and it needs to be manually done for the other port (fore for ascent is automatic, aft for skrell). Fortunately they'll be able to see this because
-Docking ports for these 3 shuttles are now less confusing to use, slightly! For these 3 ships the docking information says if the docking port is valid for docking or if you can just rub up against it. And these docking locations are now _restricted_ for the ships, meaning you won't accidentally go to "docking port" when you wanted to go to SSV docking port, because "docking port" is properly restricted to the ERT shuttle.

I didn't give the ascent accesss to the merchant ship docking port because that's the only place the merchant ship can go (and having it done this way increases the chances of a skrell-ascent war being done in one singular torch alleyway.

### DIFF
--- a/maps/away/ascent/ascent-2.dmm
+++ b/maps/away/ascent/ascent-2.dmm
@@ -52,7 +52,7 @@
 	},
 /obj/machinery/door/airlock/external/bolted/ascent{
 	airlock_type = "Internal";
-	frequency = 1333;
+	frequency = 1380;
 	id_tag = "ascent_starboard_inner"
 	},
 /turf/simulated/floor/ascent,
@@ -512,7 +512,7 @@
 "bq" = (
 /obj/machinery/door/airlock/external/bolted/ascent{
 	airlock_type = "Internal";
-	frequency = 1333;
+	frequency = 1380;
 	id_tag = "ascent_dock_starboard_internal"
 	},
 /obj/structure/cable/cyan{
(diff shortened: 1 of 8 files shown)


anconfuzedrock:
hol on, maybe I can actually make these shuttle changes and stuff it into this pr.

anconfuzedrock:
FUUCK YEAH LETS GO

anconfuzedrock:
huh, that's a runtime..

anconfuzedrock:
ok ,that's one runtime..

anconfuzedrock:
Well, I guess that's where that old airlock controller was. fucky as hell. I hope no more weird bugs or merge conflicts show up because I need to get some sleep.

PR #1466, opened 2022-03-10
Vox Missiles | SMG Frangible Buff

ProbablyCarl:
- - -
Balance:
 - Missile casings can now be constructed with 25 plasteel. This is, as per the note in the code, temp until we figure out how we want procurement to work.
 - SMG frangible penetration upped to 35, from 5. Smaller projectile, higher velocity. Or something. I 'unno. Lets try this briefly.
- - -
Map:
 - Voxship provided with a rudimentary missile bay, with two preconfigured missiles and four blanks. They have everything they need with construction through plasteel, too.
- - -
Misc:
 - Target lock noise added to missile consoles, courtesy of CookieJarvis. Thanks!
 - Pariah uncommented, but properly restricted unlike prior. Now we can shunt players into them again.
- - -


ProbablyCarl:
Maaan.
That failed check isn't even related to this. Wacky stuff.

PR #1475, opened 2022-03-10
Supply/Munition update!

Jul-Jul:
Description: This PR gives Supply (Deck Technician, Deck Officer) access to the newly added missile systems of the Dagon. The ordering, making, loading, and firing of the missiles in the case of ship to ship combat now falls to Supply. Deck Officer, and Deck Technician have also been given a few brand new alt titles to show this change, however it also includes the removal of Civilian Deck technicians...considering civilians wouldn't surely be manning a military vessel's missile weaponry.

To also nip this in the bud now before it comes up or is made a problem on the server. The Deck Officer is now in command of the missile systems, not the Bridge Officer/Tactical Officer. However, if a Deck Officer is not present, but Deck Technicians are (or aren't), the Bridge Officer/Tactical Officer has the right to command the missile systems. The Bridge Officer/Tactical Officer however remains their command over the OFD as normal. 

Another massive question also is. Why Supply, and not any other department to handle the missiles? The reason that Supply was selected for this instead any of the other departments is because in the case of a ship to ship battle, and most cases in normal rounds, Supply simply doesn't have much to do. Some may say, why wouldn't engineering handle the missiles? which whilst a good point does not take into account that when in ship to ship combat the engineers already have a lot on their plate including the engines, power, shields, dealing with breaches, etc etc.

Possible future changes: Within the future with further feedback, and more discussion with the devs, the Supply Department may also be given a different name to show off this change of responsibilities and duties. Other changes may be that Deck Technicians and Gunnery Technicians might also be made their own separate roles instead of simply the same role with alt titles.  

If you have feedback, wanna complain at me, or suggest other things to add to Supply then message or @ me on discord. "✦ Ali'nar ✦"

### DIFF
--- a/maps/torch/job/supply_jobs.dm
+++ b/maps/torch/job/supply_jobs.dm
@@ -41,7 +41,7 @@
 							 /datum/computer_file/program/deck_management,
 							 /datum/computer_file/program/reports)
 	
-	datum/job/qm/get_description_blurb()
+	get_description_blurb()
 	return "You are the Deck Officer, and the Supply department, hanger, and munitions are your domain. Your responsibilities include ensuring the ship is supplied, the flight records are filled, shuttles are fuelled, missiles are made, and the ship is always combat ready for ship to ship combat." 						 
 
 
@@ -84,7 +84,7 @@
 							 /datum/computer_file/program/deck_management,
 							 /datum/computer_file/program/reports)
 	
-	datum/job/cargo_tech/get_description_blurb()
+	get_description_blurb()
 	return "You are a Deck Technician, you answer directly to the Deck Officer. Your responsibilities are to ensure supplies ordered are delivered, shuttles are fuelled, munitions are made and loaded, and to man the ship's missile weaponry in the case  of ship to ship combat."
 
 /datum/job/mining
--- a/maps/torch/job/supply_jobs.dm
+++ b/maps/torch/job/supply_jobs.dm
@@ -55,7 +55,7 @@
 	minimum_character_age = list(SPECIES_HUMAN = 18)
 	alt_titles = list(
 		"Supply Technician",
-		"Gunnery Technician")
+		"Munitions Technician")
 	outfit_type = /decl/hierarchy/outfit/job/torch/crew/supply/tech
 	allowed_branches = list(
 		/datum/mil_branch/expeditionary_corps,
--- a/maps/torch/job/supply_jobs.dm
+++ b/maps/torch/job/supply_jobs.dm
@@ -41,7 +41,7 @@
 							 /datum/computer_file/program/deck_management,
 							 /datum/computer_file/program/reports)
 	
-	get_description_blurb()
+/datum/job/qm/get_description_blurb()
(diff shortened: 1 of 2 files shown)


Jul-Jul:
Not quite sure why the tests are chimping. If someone can figure out then please say.

maps/torch/job/supply_jobs.dm
```diff
@@ -1,15 +1,17 @@
 /datum/job/qm
 	title = "Deck Officer"
-	department = "Supply"
-	department_flag = SUP
+	department = "Support"
+	department_flag = SPT
```
Cupax3:
```suggestion
	department_flag = SPT|SUP
```
Will put them in Command Support and Supply. Probably. So it does not break with convention of having a head in its own department.

Kaksisilma (dismissed):
Looks mostly OK. A few issues I would like the opinion of @Anton-Kr , @N8-Toe and @Cupax3 on:
- Supply should not OPERATE the weapons. This duty falls to Bridge Officers. Supply should instead be charged with re-arming them.
- The Gunnery Technician alt title would be better as Munitions Technician, but admittedly this is more of a personal nitpick.
- Keep Civillian Deck Techs if possible. It is entirely feasible that reloading and resupplying weapon systems would be one of their duties. This kills some existing characters for no real benefit.

Anton-Kr:
Concur on all above points.

Jul-Jul:
Let me just go through your points.

1. If Supply Crew (which to be soon made into a different department) were not meant to man/operate the missile systems, then what's the point making them do everything else? Secondly, it's already confirmed by Carl and other developers that 2 new roles (3 slots) to be made solely for the operation of the missiles within the near future. All I've done is simply added a placeholder until that is implemented, as this gives Deck officer/techs something else to do aside from order a few crates and sit around. Though if it's a issue, I agree that normally techs shouldn't be manning/operating the missiles only making and loading, however I still think the Deck Officer should be able to or I think we should simply leave it to IC discretion.

2. I prefer that title too honestly.

3. If it was contractor or some kind of PMC contractor, I would agree but it was just civilian which doesn't entirely make much sense.

ProbablyCarl:
As per previous discussion on Discord:
image

Having something unrelated to that is absolutely fine, as it was just the product of a five second conversation.
Feedback provided on Discord, as I'm not going to step on the toes of those who've already contributed here.

N8-Toe:
i think a dedicated munitions team would be best. but I can work on that myself in the near future. I just eeehhhh on the idea of non military crew working the weapons

Anton-Kr (changes requested):
Due to lack of of attention to this I'll request the following
- Return of civilians to supply
- Change in description to be something along the lines of. "For military members of the department, the job also includes handling and assembling missiles, and prepping OFD charges as required by command." 

I am fine with civilians having access to the job, and I think it being IC enforceable is reasonable. This allows civie cargo techs to continue to just exist as civie cargo techs. On completion of these changes I will merge the PR.

Anton-Kr (approved):

PR #1478, opened 2022-03-11
Taser Carbines Tweak

Kaksisilma:
Following some comments on the Discord, I decided to look into the issue of Taser Carbines being overpowered.

Taser carbines are well-documented as being very powerful in a game ruled by stun combat (Groin shots are instant floorings and the agony damage they deal is no joke). This PR removes them from the SEC primary gunbox and adds the Electrode Rifle in their place. Two taser carbines were added into SECarm aswell.

Furthermore, the BOH variant of the Taser Carbine was removed. 24 Shots felt uneeded and a bit in the big side. I would suggest a rifle-sized smartgun or an energy SMG if we are going with laser primaries for security, and potentially re-enabling smartguns as a secondary alternative, though I think this should be deferred to Carl's judgement.

Also adds one ion rifle to the SEC's armory because those are heavily lacking in the Dagon.


Anton-Kr (approved):
Code checks out and I agree with this balance change.

PR #1486, opened 2022-03-11
Boot knife buffs

Ctr-Truescreen:
Reduces boot knife draw speed significantly, as requested by the community.
Also changes the sheath/draw sounds for something more fitting.


Ctr-Truescreen:
Tested on `localhost`. Works good.

PR #1487, opened 2022-03-11
Slightly Lowers Engineering Age Minimum

BurpleBineapple:
While looking through the jobs, I found it extremely strange that out of every department, for some reason Engineers must all be 25 or older. For a military this is extremely strange because it means we have E-3s at 25 which would be the age of something more like a staff sergeant or PO1. The age minimum has been lowered to 19, the average age of an E-3 in the US Military.

I also slightly lowered the robotics age to 22, the average age of an E-5 in the US military. This should more accurately represent our ranks in game to ages.

The engineer trainee job on the other hand still has its age limit at 18, as they are an E-2 or new E-3 learning the ropes.

### DIFF
--- a/maps/torch/job/engineering_jobs.dm
+++ b/maps/torch/job/engineering_jobs.dm
@@ -55,7 +55,7 @@
 	supervisors = "the Chief Engineer"
 	economic_power = 5
 	minimal_player_age = 0
-	minimum_character_age = list(SPECIES_HUMAN = 20)
+	minimum_character_age = list(SPECIES_HUMAN = 19)
 	alt_titles = list(
 		"Engine Technician",
 		"Damage Control Technician",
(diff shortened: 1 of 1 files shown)


maps/torch/job/engineering_jobs.dm
```diff
@@ -55,7 +55,7 @@
 	supervisors = "the Chief Engineer"
 	economic_power = 5
 	minimal_player_age = 0
-	minimum_character_age = list(SPECIES_HUMAN = 24)
+	minimum_character_age = list(SPECIES_HUMAN = 20)
```
BurpleBineapple: Upon rechecking, the average age of an E-3 is 19. So I'll be changing this to 19 to match.
```suggestion
	minimum_character_age = list(SPECIES_HUMAN = 19)
```

N8-Toe (approved):
looks good!

Ctr-Truescreen:
Engineering lore-arc characters rejoice.

PR #1489, opened 2022-03-11
Mantid Weapon Lock Bugfix

ProbablyCarl:
- - -
Bugfix:
 - Firearm lock no longer applies to the Queen. Because for some reason defines are wacky and I'm dumb.
- - -
Change/Bugfix:
 - Ascent drone provided with a near identical weapon to the standard particle weapon, although it lacks overcharge.
- - -


N8-Toe (approved):
looks good to me

PR #1512, opened 2022-03-12
buffs breacher suits

anconfuzedrock:
## About The Pull Request
First, breacher suits have half the slowdown. 3's still a lot. Next, the default breacher's torso has been buffed in laser armor and rad armor to be in line with the infantry suit. Lastly, its armor values are now higher on the helmet, and lower on the arms and legs.
## Why It's Good For The Game
First off, the breachers fucking suck. The makeshift one is supposed to be a worse version of the good one, yeah, but even the good one fucking sucks. The breacher should be cool, and maybe there's simple ways to make it cooler without making it op. It doesn't need to be as good as an infantry suit, but it should be competitive with other found rigs, so the slowdown is reduced. As for the helmet and limb guards, it's a little bit because it'd be neat for balance (It's frustrating to get instantly melted by a headshot so buffing armor there would be good, but being able to cronch their legs would mean breacher guys don't feel invincible) and a little because it'd make sense for Unathi, because they can regrow limbs but not brains. Also the head of the breacher suit was already special because it has a horn for making headbutting better but it's still a bad idea, more armor there makes it a less stupid idea.

Big breacher's admin only right now, so not sure what to do with it really, but I changed it in line with the other so it's still better.
## Did You Test It?
Yes
## Authorship
Rock
## Changelog

:cl:
tweak:breacher suit slows you down less and has a tougher helmet but weaker limbs.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/rig/suits/alien.dm
+++ b/code/modules/clothing/spacesuits/rig/suits/alien.dm
@@ -3,17 +3,9 @@
 	desc = "A makeshift Unathi battle-rig. Looks like a fish, moves like a fish, steers like a cow."
 	suit_type = "\improper makeshift breacher rig"
 	icon_state = "breacher_rig_cheap"
-	armor = list(
-		melee = ARMOR_MELEE_MAJOR,
-		bullet = ARMOR_BALLISTIC_RIFLE,
-		laser = ARMOR_LASER_HANDGUNS,
-		energy = ARMOR_ENERGY_RESISTANT,
-		bomb = ARMOR_BOMB_RESISTANT,
-		bio = ARMOR_BIO_SHIELDED,
-		rad = ARMOR_RAD_MINOR
-		)
+	armor = 0
 	emp_protection = -20
-	online_slowdown = 6
+	online_slowdown = 3
 	offline_slowdown = 10
 	offline_vision_restriction = TINT_BLIND
 	allowed = list(/obj/item/weapon/storage/)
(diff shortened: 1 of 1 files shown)


Kaksisilma:
Fine from a code standpoint.

From a balance standpoint, the armor values look very, very overtuned, and this is on top of the brute mod that Unathi get. This would be fine for the 'Fancy' version since it is adminspawn only, but even the makeshift values are really high. You currently might as well take no damage to the head or to the torso, and although legs and arms are slightly less armored, it is still a lot and they are even harder to hit.

So, here's what I propose: Lower the slowdown or even remove it, but tone down the armor values.

Alternatively, balance them in-line with Ascent, SSV or Vox suits but make them limited to the Unathi offship and/or derelict offship spawns.

If an on-ship Unathi antag gets a makeshift breacher rig, creth and oxycodone (Which aren't really that hard to get.) there'll be no way to stop them.

anconfuzedrock:
The torso armor is literally the same as the infantry rig. If anything they’re more vulnerable than that since the legs and arms are weaker and you can outrun them. I’m not changing it because that’d be silly and it’d be like any other generic rig. Please reconsider.  By this logic anyone with the infantry rig or cos rig on those drugs is also unkillable, and the cos rig is easy to steal. Any antag with trained electrical can hack it.

anconfuzedrock:
I will say, if I had to pick one, I'd pick the latter, making it rarer. I think it's pretty rare as it is, though.

Anton-Kr:
Honestly the changes seem fiiine.

PR #1523, opened 2022-03-13
reduces tiro skill points

anconfuzedrock:
## About The Pull Request
This reduces Tiro skill points from 50 to 36, (from 58 to 44 after the general job skill bonuses) the default for away maps like the bearcat and verne. 
## Why It's Good For The Game
58 skill points isn't just the highest skill point value in the game, it's 10 higher than the second highest. Why should they have this? Because, according to the code comment. it's equivalent to what roboticists can afford. first off, this is very misleading- roboticists have an incredibly high skill point total anyways. Second off, the difference is that the Tiro can spend those skillpoints on whatever they want, such as getting master combat, master weapons, AND master athletics, and still having 12 points left over. Even the COS can't do that. The brig chief can barely afford that, and they have the most spare skillpoints of any job on the dagon as well as combat skills.

Why should a tiro have this many points? There isn't much of a reason. A tiro could be anyone? Well, that's the theory with the bearcat and escape pod, pod survivors can be anyone, but they don't have 58 skillpoints. They have 44, and 44 is still quite a bit. 

Why SHOULDN'T they have this many points? Well, there's a few reasons, it being horribly imbalanced is one thing, but With all the free points, and the suits, and not having to deal with the complexities of being an alate, Tiro is far more appealing of a role than alate. and that isn't inherently a bad thing, but the ascent isn't just a bunch of humans in purple suits, and with a skillspread like this, they can do everything an alate can just fine.
 
If 44 points isn't enough for your tiro then that's a skill issue. as in, skills are too expensive, and we should fix it for everyone, or at least every offship role, and not just for the tiro.
## Did You Test It?
1. look at that code and tell me it needs to be tested. LOOK AT IT
2.Yes, actually, I did test it!
## Authorship
Rock
## Changelog

:cl:
tweak: Tiros have less skill points.
/:cl:

### DIFF
--- a/maps/away/ascent/ascent_jobs.dm
+++ b/maps/away/ascent/ascent_jobs.dm
@@ -258,7 +258,7 @@
 	blacklisted_species = list(SPECIES_VOXPARIAH, SPECIES_VOX, SPECIES_VOX_ARMALIS, SPECIES_ADHERENT)
 	whitelisted_species = null
 	loadout_allowed = TRUE
-	skill_points = 50 //Just *about* the # for a Roboticst at default, counting their preset skills. We have no min-skill level for this role since anyone could be deemed "interesting".
+	skill_points = 36
 	min_skill = list()
 	requires_supervisor = "Ascent Gyne"
 	set_species_on_join = null


Anton-Kr (approved):
This is reasonable

Cupax3:
DNM to give those in opposition of this PR time to formulate a response.

Pariah919:
```

Why should a tiro have this many points? There isn't much of a reason. A tiro could be anyone? Well, that's the theory with the bearcat and escape pod, pod survivors can be anyone, but they don't have 58 skillpoints. They have 44, and 44 is still quite a bit.
```

I can see both sides of this argument, Tiros are probably atleast somewhat exceptional people already, considering they have had to grab the interest of a Gyne. I know next to nothing about Ascent lore.

But the skill points are absurd, so I agree with this PR, I think 36 is fair.

anconfuzedrock:
I do think some things are a bit more expensive for offships than they should be (particularly expert anatomy, offships are the only people who *can* take it yet it's prohibitively expensive for them)  but if there's an issue of offships being unable to take stuff because of that, then giving _one_ offship excessive skillpoints instead of solving the problem directly won't get to the root of it or help those other jobs.

N8-Toe:
I think this looks good as well

Anton-Kr:
Merging on basis of two maintainer consensus.

PR #1528, opened 2022-03-14
body shielding is directional

Anton-Kr:
## About The Pull Request
Body shielding is now directional and works in a _cone_ like way. If you really want you can expand the coverage area to be bigger following the way this code is written. 
I'd like another maintainer to test this first but like legitimately it took me three fucking minutes to code this and two minutes of that was waiting for DM to work, why did no one do this in the past 2 months that this has been an issue when it was legitimately trivial to fix... unless it doesn't fix it and Im wrong but I digress 

## Why It's Good For The Game
Body shielding less cringe

## Did You Test It?
Yes

## Authorship
Me

## Changelog

:cl:
tweak: Body shielding is now directional
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -263,7 +263,7 @@
 		if(istype(A, /mob/living))
 			//if they have a neck grab on someone, that person gets hit instead
 			var/obj/item/grab/G = locate() in M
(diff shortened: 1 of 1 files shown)


ProbablyCarl:
It's almost as if it was a trivial problem that people only complained about for the sake of it.
Thank you for this.

PR #1531, opened 2022-03-14
nerfs firing speed of all pistols and adds jam chance to p20 (WATCH THE VIDEO IN THE DESCRIPTION GOD PLEASE)

anconfuzedrock:
## About The Pull Request
not quite a revert of #1463 but somewhere in between.
## Why It's Good For The Game
First off, the p20 just has jam chance to make it a reasonable trade between it and the mk49.
Second off, please see this video. SEEEEEEEE THIS VIDEO.

https://user-images.githubusercontent.com/18406892/158237997-2a7defa8-1d89-45a4-9e3b-a967d5282d6c.mp4

TELL ME the entire crew should have weapons that shoot that fast. I've tested this on dummies in ballistic armor (highest non-admin armor rating) and magdumps were able to cause internal organ damage, internal bleeding, and fractures. Magdumps take 4-5 seconds, as can be seen from the video. I Don't know if it would KO the person being magdumped or not but I'm pretty sure they'd be stunned by the pain. SMGs and rifles are pathetic in comparison. let me restate this, the SMGs shoots noticably slower on full auto. And has weaker bullets.

also, uh, thirdly, I guess, the mk49's advantage was supposed to be shooting faster than the p20. considering they both shot as fast as you could fucking click, that advantage was meaningless. increasing the gap means that upside may have an effect.

This mk49 fire rate value is still faster than previous, as is the revolver. just a little bit slower than FULL AUTO SMG, now. 
## Did You Test It?
Yes
## Authorship
Rock
## Changelog

:cl:
tweak: handguns shoot at a somewhat reasonable rate again.
/:cl:

### DIFF
--- a/code/modules/boh_misc/firearms.dm
+++ b/code/modules/boh_misc/firearms.dm
@@ -8,7 +8,7 @@
 	magazine_type = /obj/item/ammo_magazine/pistol/double/pepperball
 	fire_sound = 'sound/weapons/gunshot/pistol_mk59.ogg'
 	jam_chance = 5 //Cheap firearm. Chance of jamming
-	fire_delay = 2 // Fires faster than usual
+	fire_delay = 5 // Fires faster than usual
 	damage_mult = 0.9 // Damages a bit less than peers.
 	origin_tech = list(TECH_COMBAT = 4, TECH_MATERIAL = 2, TECH_ESOTERIC = -1) //bandaid fix to prevent r&d from getting free esoterics
 
@@ -57,7 +57,7 @@
 	desc = "The Lumoco Arms' Mk12 is a rugged revolver for people who don't keep their guns well-maintained. Unlike its cousin, the Mk59 'Jhen Bothus', it has no issues with reliability."
 	accuracy = 2
 	bulk = 0
-	fire_delay = 5.5
+	fire_delay = 8
 	penetration_mod = 10
 	falloff_mod = -0.5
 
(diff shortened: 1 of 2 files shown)


Pariah919:
>First off, the p20 just has jam chance to make it a reasonable trade between it and the mk49.
Second off, please see this video. SEEEEEEEE THIS VIDEO.
>
why does the mk59 now have no jam chance and the P20 now has a jam chance?
What is the purpose of this change of reversing the entire (probably stupid to be fair) gimmick behind both of these firearms? Just give them no jam chance at this rate and make them equivalent, or leave the jam chances alone.

0.8 for the revolver makes it bad again against any half decent target, probably
0.6 is a bit too high, in my opinion.
I think 0.2 and 0.25 is too high for Hestias balance/game speed which is agreeable by most people. Though the SMG i use as a basis isn't the one Sec gets (Which is 0.2 and not 0.4, though I don't get why the sword can't just get reduced to .3 itself).

I'd move them personally up to .3 and .35 respectively, Which while not sounding like much of an increase, is a pretty big increase in fire delay considering its in deciseconds (The difference between 5 shots a second and 3 for the MK59, and 4 and 3 for the P20) Leave the revolver as is at .55. Its fine there.

You could also reduce the pistol bullet damage overall in the future, as I think higher ROF with less damage is more conductive to gameplay, but that's for another argument or day.

anconfuzedrock:
You read the code wrong. both have jam chance now. I don't have anything to say about everything else, bullets usually stun from the pain anyways, this fire rate is absolutely excessive. This "feeling bad" for you is less significant than the feeling bad of being killed in 2 seconds from a magdump from the pistol the entire crew gets.

N8-Toe:
We do not allow reactionary PR's unless the issue at hand is egregious. Lets let this sit for awhile and see where we in awhile

Carl has signed off on this PR'ing being closed. Feel free to reach him or I if you have an issue or wish to discuss this further.

PR #1549, opened 2022-03-16
Citizens & Feds

ProbablyCarl:
- - -
Additions:
 - Provides a new category to IDs and player characters. This is citizenship within SolGov. Your rank determines where exactly that sits, if anywhere at all. This is likely to need adjustments in the future, but it worked in testing.
- - -
Changes:
 - Detective required to be a Federal Agent now, as per the change above. They, the Captain and SolGov Rep are the only First Citizens aboard at round start.
- - -


SilverCelestine:
Sounds good to me and it'll fit in well with an upcoming change to Dagon Regulations and the creation of Sol Federal Law.

Jul-Jul:
Don't forget the SGR guard is a Federal Agent too.

ProbablyCarl:
That's fair. It's set by rank, so it should be fine.

Sgtwok:
Removing marine, contractor, and fleet detective? Why?

PR #1554, opened 2022-03-16
adds the ability to purchase missile parts through supply

anconfuzedrock:
## About The Pull Request
You can now order all missile parts, including casings, through supply. Missile parts come in packs of five for about 40, depends on the type of missile components. Missile CASINGS are 1 for 75.
## Why It's Good For The Game
Missile supplies should be at least slightly replenishable, this makes sense, and the prices are high enough that it won't overshadow a more complex missile construction option if one is ever added. This also makes munitions actually related to supply.
Talked with Antonkr on prices. Casings are expensive so missile factory would be really hard to pull off.
## Did You Test It?
yes
## Authorship
Rock
## Changelog

:cl:
rscadd: you can now order missile components from supply.
/:cl:

### DIFF
--- a/code/datums/supplypacks/munitions.dm
+++ b/code/datums/supplypacks/munitions.dm
@@ -46,3 +46,54 @@
 	name = "OFD-Launched Drop Pod"
 	contains = list(/obj/structure/closet/odst)
 	cost = 30
+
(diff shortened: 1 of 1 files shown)


Cupax3 (approved):
code should be fine

PR #1577, opened 2022-03-17
Adds a non-self-charging version of the compact smartgun as a sec sidearm

anconfuzedrock:
Worth a shot, anyways.
## About The Pull Request
This adds a compact smartgun as a sec ~~primary~~ Secondary. Instead of rapid self recharging and having 3 shots, it has to be manually recharged and has five. The electrolaser has 9 shots now.
## Why It's Good For The Game
~~Gives a unique spin on the sec primary options. Sec primaries are kind of weird, They vary in lethality but all of them are at least a little bulky and you might consider chucking them out if you ever get new guns. If you like that, great, but if not, it feels awkward.~~

~~This compact smartgun's meant to be the opposite of the other options, so you have something convenient and simple, but not very powerful.  It fits in pockets and pouches so it's easy to carry, but it's not going to do anything crazy. The 5 shots makes it complementary to your other small arms, and it gives you an option for shooting carp through windows besides the electrolaser, so you can feel comfortable taking a different one of the handgun options.~~

~~Compact lasers do even less damage than the mk49 does with its reduced damage (as well as no shrapnel or internal bleed chance) and this gun shoots a bit slower too and has a third the ammo, so I don't think this is going to be a significant balance change.~~

The electrolaser's a bit poopy with 5 shots you can't really reload, and it's also just so ugly. 9 Shots means that it goes by in 3 seconds rather than 2, so you might stop and think about actually making the shots count.

If you like having fuckall for ammo, the compact smartgun is like if the electrolaser felt good to use. Looks nice., has five shots but that doesn't feel as shitty because it fits in your pockets.

Lastly, I just miss the lil' guy. I don't think I'm alone in this. 🥺
## Did You Test It?
Yes
## Authorship
🥺Rock🥺
## Changelog

:cl:
rscadd: Adds a manually-recharged compact smartgun to the sec secondary kit options.
/:cl:

### DIFF
--- a/code/modules/boh_misc/boxes.dm
+++ b/code/modules/boh_misc/boxes.dm
@@ -67,6 +67,7 @@
 //	options["Energy - Smartgun"] = list(/obj/item/weapon/gun/energy/gun/secure)
 //	options["Energy - Stun Revolver"] = list(/obj/item/weapon/gun/energy/stunrevolver/secure)
 	options["Stun - Non-Lethal Taser"] = list(/obj/item/weapon/gun/energy/taser)
+	options["Small Energy - Compact Smartgun"] = list(/obj/item/weapon/gun/energy/gun/small/secure/sec)
 	var/choice = input(user,"What type of equipment?") as null|anything in options
 	if(src && choice)
 		var/list/things_to_spawn = options[choice]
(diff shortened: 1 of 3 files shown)


Anton-Kr (changes requested):
I think this belongs better in security sidearms. Also buff the taser to accommodate this change. Some 8-10 shots would make it a reasonable alternative.

anconfuzedrock:
okie dokie

Anton-Kr (approved):

PR #1591, opened 2022-03-18
Ascent clothes shuffle

N8-Toe:
Ascent clothes shuffle

Implements sprites for Monarch Queen over/under clothing
implements sprites for Monarch Worker over/under clothing

as a side effect adjusted how Gyne/Alate clothes work, should have no player facing difference


Shirumic:
<3

PR #1597, opened 2022-03-19
Changes the Merc SMG to actually be an accurate description.

BurpleBineapple:
Merc SMG now has a description that actually is accurate to it.

## About The Pull Request
Changes the Merc SMG (10mm) description and name.

## Why It's Good For The Game
For some reason, the base merc_smg is called the MA-C20v and says it fires 7mm? Even though it doesn't? Now it's just a normal NT C-20r as it always was.

## Did You Test It?
Yes.

## Authorship
PurplePineapple#0001

## Changelog

:cl:
bugfix: merc_smg no longer says it fires 7mm. It's been renamed to the NanoTrasen C-20r and has an accurate desc.
/:cl:


BurpleBineapple:
Me when bearcat tests don't run

PR #1610, opened 2022-03-20
Hestia: Slightly-Less Armored Edition

Dark1Star3s:
## About This Pull Request: (Also edited because apparently you need a format to read big text.)

Slightly nerfs the armor values on all exo-suit platings. EM Values were too high across the board and were lowered. Watch out for ions!

## Why it's good for the game.

While we like to have fun there needs to be some fun had while using mechs, mechs are still VERY strong and you shouldn't take one on alone unless you're superman or something.


Kaksisilma:
Please follow the description format

PR #1611, opened 2022-03-20
Seizure moment

Lamasmaster:
## About The Pull Request
Removes seizure from defensive lance but in compensation gives 5 more agony damage and 4 more shots.

## Why It's Good For The Game
Seizure has absolutely no counterplay as is, it ignores any armor and any chemicals you can throw at it. So what essentially this turns it into is stronger stun rifle which security already has. Willing to work out details of this but I just don't see why it should just stay in the way this is (and before someone says this is a salt PR read stats of stun rifle).

## Did You Test It?
Yes

## Authorship
Me

## Changelog

:cl:
tweak: Defense Lance
/:cl:

### DIFF
--- a/code/modules/boh/ascent2/firearms.dm
+++ b/code/modules/boh/ascent2/firearms.dm
@@ -5,7 +5,7 @@
 	name = "defence lance"
 	desc = "A modified variant of the Ascent particle lance, usually carried by rearline combatants."
 	force = 6
-	max_shots = 2
+	max_shots = 6
 	burst = 2
 	move_delay = 2
 	one_hand_penalty = 0
--- a/code/modules/boh/ascent2/firearms.dm
+++ b/code/modules/boh/ascent2/firearms.dm
@@ -5,7 +5,7 @@
 	name = "defence lance"
 	desc = "A modified variant of the Ascent particle lance, usually carried by rearline combatants."
 	force = 6
-	max_shots = 6
+	max_shots = 2
 	burst = 2
 	move_delay = 2
 	one_hand_penalty = 0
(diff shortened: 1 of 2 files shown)


ProbablyCarl:
The entire purpose of the weapon was to give a no-counterplay capture item to the Ascent. Without it, things generally turn into slug fests and negate any potential for capture.
While I can understand not wanting to be smacked with it, this is intentional behavior.

ProbablyCarl:
Actually, I'll go halfway here.
After very brief internal discussion, with some of the team, we believe a warmup would be a better alternative to this. If you can make it function identically to the RPG in terms of firing delay, I'll go ahead and speedmerge it.
Otherwise, this PR is kind of unfortunate in that it sits in a grey area.

Sparrowbird611:
I'm just kinda curious about what's wrong with this variant of the defense lance. The stun rifle that security gets is a 5 shot 80 agony stun rifle that can one tap people with groin shots currently, with this variant the defense lance gets 3 shots before needing to recharge, deals 90 x 2 agony. It's arguably better at capture gameplay right now, since it can mow down multiple people before needing to recharge, with it's only downside that I can see being that it works off electrical armor which is a questionable choice IMO.

I'm don't care much either way but I'm just wondering as to why you think this variant would suck at capturing people, since it's for all intents and purposes a hitscan, self recharging, double shot stun rifle. It's not like the majority of people the ascent would want to capture are pain resistant, and if the concern were IPCs I thought they were immune to seizures.

Lamasmaster:
Alternative that I could see that would actually make some sense is making it paralyze the person it hits instead of seizure since seizure is get hit once and die, paralyzing someone results in it doing the same thing except it won't kill the person in single hit. Also not sure what you mean by RPG fire delay, like the windup time ?

Rayfod:
You can use this for reference for the wind-up code.

```
	var/windup = 5 // Rockets take half a second to prep to fire, set to zero to invalidate completely.
	
	
/obj/item/weapon/gun/projectile/rocket/afterattack(atom/A, mob/living/user, adjacent, params) // RPGs snowflake this due to windup, okay? Else ancient baycode gets angry and starts erroring.
	if(adjacent) return //A is adjacent, is the user, or is on the user's person

	if(!user.aiming)
		user.aiming = new(user)

	if(windup > 0)
		if(do_after(user, windup)) // Do the windup.
			Fire(A,user,params) //Otherwise, fire normally.
	else
		Fire(A,user,params) //Otherwise, fire normally.
```

ProbablyCarl:
> Alternative that I could see that would actually make some sense is making it paralyze the person it hits instead of seizure since seizure is get hit once and die, paralyzing someone results in it doing the same thing except it won't kill the person in single hit. Also not sure what you mean by RPG fire delay, like the windup time ?

The windup time, yes.

Lamasmaster:
Removed my other changes but now it has windup time just like RPG and paralyzes the target for (actually something around a minute depending on server tickrate and time dilation) which should still fill the niche of capture weapon where it now even affects synthetics but has some actual counterplay possible against it.

ProbablyCarl:
Up that paralyze time and we'll merge it. Two seconds is very much two seconds. Both beams connecting will give four, unless I'm mistaken, which is far less than you're thinking.

PR #1615, opened 2022-03-21
removes hair options from Unathi

anconfuzedrock:
## About The Pull Request
This removes Unathi's ability to take human hair in their head accessories. Or beards.. Veteris still can because they don't have the offset issue.
## Why It's Good For The Game
Now, although it doesn't make sense for 99.9% of Unathi, I'm not removing their genemodder option. But no one on the PLANET wants Unathi to have hair options. They're not shaped for unathi faces, so basically all hair options just look like a mess. And this actually causes a problem because it makes it hard to find the actual unathi options (for horns and frills).
## Did You Test It?
yes
## Authorship
rock
## Changelog

:cl:
rscdel: non-veteris unathi can no longer take hair or beards.
/:cl:


modular_mithra/code/modules/sprite_accessories/_accessory_hair.dm
```diff
@@ -1,9 +1,9 @@
 
 /datum/sprite_accessory/hair
-	species_allowed = list(SPECIES_HUMAN,SPECIES_UNATHI,SPECIES_OLDUNATHI,SPECIES_VOX,SPECIES_AKULA,SPECIES_VULP,SPECIES_VASS,SPECIES_TAJ,SPECIES_IPC,SPECIES_CUSTOM,SPECIES_HUMAN2,SPECIES_PROMETHEAN)
```
Anton-Kr, line 3: is OLDUNATHI having them intentional?

anconfuzedrock: Yep! I think the hair options are dumb, but at least they WORK for veteris.

Anton-Kr (approved):

PR #1621, opened 2022-03-21
combat medic gets med channel

Jeff-jefferson-jeffery:
discord asked for it

## Did You Test It?
no


ProbablyCarl:
We removed it because they're not really great when having it.
I gave it to them initially, but it turned out to be a mistake.

Anton-Kr:
Gonna agree here.

PR #1634, opened 2022-03-22
Initial Lace Rebalance

ProbablyCarl:
- - -
IMPORTANT
- - -
Laces, now rather than killing their user immediately, provide an avenue of RP, keep the player in the round and are no longer an incredible detriment. Mostly.
Destruction of the lace will remove all skills, temporarily stun you via seizure, give mild brain damage and force hallucinations for a hot minute. To compensate, it has been made far harder to destroy and still retains the 0.5 damage modifier.

To understand why, it's assumed that the user's conscious, stored normally in the lace to some degree, would sustain tremendous damage upon destruction of said device. You don't so much lose yourself as it is your brain being forcefully rewired, the implant being housed between the cerebellum and medulla oblongata and in a prime area for such.
Alternatively, refer to the in game description: "By way of violent neuron rearrangement, you're practically a husk."

This will likely need tweaks, but is absolutely more in line with what we're aiming for.

A unique sound for destruction has been provided by [ShadeAware](https://github.com/ShadeAware). Thank you for permitting me to use it.
- - -


Sparrowbird611:
So basically losing your lace lobotomizes you. That actually sounds funny af ngl.

code/modules/organs/internal/stack.dm
```diff
@@ -106,16 +113,45 @@
...
+/obj/item/organ/internal/stack/proc/smash()
+	to_chat(owner, "<span class = 'notice' font size='10'><B>What happened...?</B></span>")
+	sleep(5 SECONDS)
+	to_chat(owner, "<span class = 'notice' font size='10'><B>Am I awake...?</B></span>")
+	sleep(5 SECONDS)
+	to_chat(owner, "<span class = 'notice' font size='10'><B>Who am I...?</B></span>")
+	sleep(5 SECONDS)
+	to_chat(owner, "<span class = 'notice' font size='10'><B>Is this the void...?</B></span>")
+	sleep(1 SECONDS)
+	alert(owner, "You have lost your lace. By way of violent neuron rearrangement, you're practically a husk. \
```
ShadeAware, line 125: I am really, really, really, really against this in a way that is impossible to overstate; completely removing a person's identity and forcing them to become a lobotomite doesn't open up a lot of avenues for roleplay, it just removes a person's adjacency and forces them to basically become a drone. That isn't fun, its just going to make people cryo out since they're basically dead anyway and if they don't want to play braindead they're going to be banned.

I would heavily suggest either heavily toning this down into something less completely absolute, or just re-instating owner.death() since it is arguably the same thing.

Rayfod: I have to agree with this, that it does make the person in question 100% want to go into cryo. And then come back 15 mins later without taking care of all the numerous debuffs and having to play stupid for the 1-2 hours of remaining time after the fight.

I would rather have a more roleplay-focused feeling, where you might have lost part of your memory, and what has happened to you at the current shift.

N8-Toe (approved):
Right my review

I do think the text on lace loss is maybe a half notch to high. But I with the adjusted relative size lace destruction should not be a "common" thing, but still present enough that is a risk

To those who are concerned, remember code and descriptions are not chiseled in stone tablets handed down by God. If it causes issues, we're come back and review it in a week or two. Ontop of that the usage of a lace is an "opt in" mechanic, you are not required to utilize one. 

Code looks good, approved

PR #1636, opened 2022-03-22
Reduces the size of the KOSMAG rooms

BurpleBineapple:
## About The Pull Request
This PR aims to reduce the size of the missile area.
image
image

## Why It's Good For The Game
The addition of the KOSMAG added a long hall plus lower 'magazine' that took up about 30% of starboard maintenance on D4 and D3 combined. The worst bit is that, from player feedback, the extremely long conveyor fails to work most of the time anyway. Hilariously, it also leaves the KOSMAG wide upon to anyone in maint who crawls under the flaps.

To give maintenance back some personality, as well as reduce the size of this addition, I have gone ahead and made a few new little maintenance easter eggs in the place of the KOSMAG hall and lower missile storage. The missiles are now on the conveyor by default and the lockers for parts are in the firing room. This now mirrors the OFD which is also in maintenance.

Some screenshots:
image
The old hall (The red door is now a maint door.)

image
Missile storage underneath.

image
The actual KOSMAG entrance

## Did You Test It?
Yes.

## Authorship
PurplePineapple#0001

## Changelog

:cl:
maptweak: The KOSMAG has been reduced in size. Old areas are now maintenance areas.
/:cl:


Teragenic:
Right. Know it's a bit meh but reducing the size of KOSMAG gave me the impression you were decreasing KOSMAG shell explosion size.

This PR should be renamed reduced Missile Magazine size. 


Just my little tidbit, but yeah PR looks and seems to be actually good for QoL. Just clickbaited me.

BurpleBineapple:
No idea why the Bearcat is failing again.

gilvian:
> No idea why the Bearcat is failing again.

remove some filth spawn apperently

Anton-Kr:
_Since this is my mapping I'm leaving all the following comments as a random, and not as a maintainer._ 

Honestly pretty against this change.

Actual problems seen:
- Where is the missile equipment? All the crates used to store it, so I do not see it now.

Feedback:
I feel like this removes some integral map design of the missile magazines. 

- Multi-z elevator was designed to make loading missiles tedious, and require effort. It's also frankly better map design to utilize multi-z
- The extension into the hallway was designed for missile department to feel like well... a real department. Why should deck/munitions techs need to crawl through maint to reach the missiles? I get that it reduces a fair bit of maint, but majority of the area that has been relegated back into maint used to be the gym anyway (which I removed as my pool addition has a gym anyway). 

I wouldn't mind something to make the setup more compact, but I feel like this is a far more boring version of what I made.

BurpleBineapple:
> Actual problems seen:
> 
> * Where is the missile equipment? All the crates used to store it, so I do not see it now.

In the main room. Those white lockers are all missile equipment.

> * Multi-z elevator was designed to make loading missiles tedious, and require effort. It's also frankly better map design to utilize multi-z

This can already be bypassed by quickly shoving everything onto the elevator, going up, then unloading it all. Every other ship has their missiles loaded onto the conveyor anyway. I've also received feedback from people actually using this such as John_Bigless and Lamasmaster that the current setup bugs out frequently due to the long conveyor belt. Missiles frequently get stuck in the multiple flaps between the doors and security for this entire area is near zero due to the fact that all it takes to break into the missile area is to crawl under a single flap. That is not what I would call better map design, personally.

> * The extension into the hallway was designed for missile department to feel like well... a real department. Why should deck/munitions techs need to crawl through maint to reach the missiles? I get that it reduces a fair bit of maint, but majority of the area that has been relegated back into maint used to be the gym anyway (which I removed as my pool addition has a gym anyway).

While I understand what you mean, this isn't really its own department. Nor is it treated as such by players. It is a part of supply and accessible by command. It also mirrors the OFD and in most cases, being that it is also used for ship-versus-ship actions, feels like it mirrors its purpose as well.  Command already had to make their way through maintenance to access the OFD. It seems fitting that the missiles would share the same. I see it more as an addition to a refitted vessel as an afterthought like with the OFD.

miniusAreas:
will conflict with #1637. 

Personally, there's an easier way to fix some of these concerns, and although I was biased on the size of it previously (i wanted my free shaker before entering engineering lmfao)

Additionally, this space will be cut down in the future, specifically when we get around to more gunnery equipment. Parts of d4 and some on d3 may end up becoming it's own gunnery department. d4 is quite literally maintenance, supply and pods, with infantry and some civ stuff tucked away. From a mapper's perspective and from the words of Oranges and several others who helped teach me, Maint space is *not* just something there for flavor. Specifically, it's there for two reasons.

ONE - it gives space for future expansion to get placed (futureproofing.)

TWO - gives players and antags room to do fun expansion works or gimmicks. This one comes at a lower priority than one, as maint CAN always be expanded.

Remember that the ships hull can be changed, expanded, and shrunk to whatever needs are needed. Don't feel pressured to limit yourself to the current silhouette.

BurpleBineapple:
> Additionally, this space will be cut down in the future, specifically when we get around to more gunnery equipment.

Then I suggest expanding when these are ready instead of basically cutting the ship in half to fit a conveyor that does not even work. As it is now, the current setup is unnecessarily long.

Anton-Kr:
Should note that on testing the conveyor does in fact work. It has a fair deal of wonkiness with the sprites being 2x1 but still.

PR #1640, opened 2022-03-22
Re-adds Telecomms Relay

Rafflesia7793:
## About The Pull Request
This PR adds telecomms relays back to the game that were removed: https://github.com/BoHBranch/BoH-Bay/commit/e262fedcd5609832e3b37e4a9e3f092c682cde3c

## Why It's Good For The Game
This will allow exploration to stay in more direct contact with the Dagon, if RnD puts the time and effort in to construct the relay, and Engineering assists with the extra power draw the relay creates.

## Did You Test It?
image

## Authorship
Rafflesia, Baystation


Rafflesia7793:
Checks are failing because I didn't push Origin before merging PR, so indent fix got left out.

gilvian:
Exploration main here - As much as this thing seem helpful, I doubt it should be added onto the Dagon, maybe for other off-ships but not the main one.

Exploration should be a bit about exploring and going out into the danger, instead you just go now ;help planet bad and really most of the danger turns into just awaiting for dagon to teleport. Also makes the long ranges useless and dagon instead of worrying about you crashing you can just also go ;help pilot dead behind the helms control

Kaksisilma:
Skimmed briefly through changed files and this looks fine code-wise. 

Waiting for feedback, though I myself am not too keen of this change.

Rafflesia7793:
I don't mind changing the requirements to print the board off or build the machine, but I see no reason why it couldn't exist in code and be utilized, even if only by admin fiat or mapping.

Rafflesia7793:
In light of this screenshot I really cease to see any real reason to need the relay, or to not implement it, just do whatever you like: 
image

The relay takes 50kW, but just building a broadcaster and receiver with stock parts readily available from Tcomms storage only consumes 1kW.

Rafflesia7793:
image

gilvian:
> image

Proof we don't really need it so we can close down the PR :^)

but for real I think that should be just for off-ship people since their comms are really scuffed, like skrell and vox after sometime just turn off and don't function

Anton-Kr:
Eh I don’t think people will bother with this much anyway.

Anton-Kr (approved):

PR #1651, opened 2022-03-23
[TWEAK] Modifies the skill penalties of lace destruction on non-combat related skills. Makes the language of personality loss less overbearing.

ShadeAware:
## About The Pull Request
All **non-combat** skills have had their penalties from lace destruction modified from -5 to -2. CQC and Ranged Weapons skills unaffected, obviously. Values subject to change and should be investigated to ensure compliance with balance.

Language of lace destruction has been **heavily** toned down from straight character deletion to a more mild agonizing bout of temporary memory loss and damage. 

New sound effect for the above!
## Why It's Good For The Game
Laces as they stand are still too much of a detriment to really be worth it for anything other than assistant (and arguably, even that is shooting yourself in the foot.) This PR aims to at least tone down its mechanical effects.

Complete character wiping on lace destruction is awful in every sense of the word and should have never been allowed. If you're going to completely delete someone's personality AND skillset, you may as well just call a death proc since most people will cryo out anyway. That isn't good roleplay. Its death with extra steps. This PR aims to remove that aspect of laces and leave people's characters in-tact should their lace be destroyed.
## Did You Test It?
A few line changes.
## Authorship
ShadeAware
## Changelog

:cl:
/:cl:


Anton-Kr:
Agree with this

Kaksisilma:
Indifferent to the changes. Looks ok code-wise, but I suggest deferring this one to The Man™

@ProbablyCarl

SilverCelestine:
I'm going to have to hard disagree with this. You **are** your Lace. If you make a clone of yourself, it sits there with a vacant gaze in its eyes. Your skills, personality, memories, consciousness, everything that makes up who you are is preserved in the form of data on the lace. 

Through the technology of laces, we've arguably transcended our humanity. The bodies we inhabit are no longer ours and we're no longer entirely mortal, they're a shell from which we pilot. When the lace (which is **US**) is gone, all that's left is a shell. 

If you don't want your lace to be destroyed then one should take extra precautions into preserving it integrity, such as not charging into danger with reckless abandon because you can just be resleeved!

Pariah919:
Agreeing with @SilverCelestine here as a medical player (though I'm takign an exodus from medical until sutures)

Laces make combat a bit *too* forgiving for my tastes, brainmed already gives you a huge gap before dying, but my take is that you should move them to the chest area, make them tiny (size 5) and add redgrab moves for removing them with a knife or shooting them with a gun, but removal should still lead to effective death in most cases.

ShadeAware:
> I'm going to have to hard disagree with this. You **are** your Lace. If you make a clone of yourself, it sits there with a vacant gaze in its eyes. Your skills, personality, memories, consciousness, everything that makes up who you are is preserved in the form of data on the lace.
> 
> Through the technology of laces, we've arguably transcended our humanity. The bodies we inhabit are no longer ours and we're no longer entirely mortal, they're a shell from which we pilot. When the lace (which is **US**) is gone, all that's left is a shell.
> 
> If you don't want your lace to be destroyed then one should take extra precautions into preserving it integrity, such as not charging into danger with reckless abandon because you can just be resleeved!

1. That... isn't how brains, or memory work. Laces are extremely vague in their actual function and can be quite literally anything.
2. You're basically making things up to justify what is an unjustifiable feature of something that is OSTENSIBLY supposed to make it more difficult to die. 

What is the difference between just calling a death proc and wiping someones character? **Nobody** wants their fucking character wiped, and with how much hate laces have been getting I honestly struggle to see why we keep them around at this point. You shouldn't be fucking forced to play a lobotomy patient because, God forbid, you wanted to avoid a 3 hour round removal because Murderbone Mc.Why-Isn't-He-Banned-Yet is on yet another antagonist rampage, at **gunpoint** of being banned for not following this asinine rule.

And your thing about "just being re-sleeved" demonstrates that you really have _no idea what this PR even does._ The entire point of this PR is to make the consequences for your lace being destroyed _slightly less severe_ because at the moment there is absolutely NO reason why anyone should have one, combat-role or not.

You want laces to be an actual detriment to have and completely, totally useless? **Remove them.** Or make the lace kill people on destruction. At its current state, the only difference is that you have to cryo before round removal.

Rafflesia7793:
I like the changes overall. The skill system wasn't designed with skill loss in mind however and -2 may very well lock surgeons out of their job altogether. However a master level electrician isn't affected at all by -2 in electrical. Just as -2 in bureaucracy and finance does nothing.

It would be nice to see the skill penalties asymmetrically applied so that every job is heavily affected, but no one is unable to do their job entirely, assuming they weren't borderline incompetent at it already skill-wise.

Anton-Kr:
Following discussion with Carl and some other staff I believe the consensus was ultimately against this change. Thank you for the effort nonetheless.

PR #1652, opened 2022-03-23
No! Bad snek!

Cupax3:
## About The Pull Request
Sneks are no longer able to be antagonists other than changelings. And converted antagonists. One problem at a time.

## Why It's Good For The Game
GAS potators were fun the first time. Actually, no, they weren't.

## Did You Test It?
image

## Authorship
Me

## Changelog

:cl:
balance: Snek no badguy
/:cl:

Note: This only prevents being chosen as antagonists by ticker (late antag spawn) or by game start (lobby). This does not prevent conversion to cult - one problem at a time.

### DIFF
--- a/code/game/antagonist/antagonist.dm
+++ b/code/game/antagonist/antagonist.dm
@@ -113,19 +113,21 @@
 
 	// Prune restricted status. Broke it up for readability.
 	// Note that this is done before jobs are handed out.
-	for(var/datum/mind/player in mode.get_players_for_role(id))
-		if(ghosts_only && !(isghostmind(player) || isnewplayer(player.current)))
+	for (var/datum/mind/player in mode.get_players_for_role(id))
+		if (ghosts_only && !(isghostmind(player) || isnewplayer(player.current)))
 			log_debug("[key_name(player)] is not eligible to become a [role_text]: Only ghosts may join as this role!")
-		else if(config.use_age_restriction_for_antags && player.current.client.player_age < minimum_player_age)
+		else if (config.use_age_restriction_for_antags && player.current.client.player_age < minimum_player_age)
 			log_debug("[key_name(player)] is not eligible to become a [role_text]: Is only [player.current.client.player_age] day\s old, has to be [minimum_player_age] day\s!")
-		else if(player.special_role)
+		else if (player.special_role)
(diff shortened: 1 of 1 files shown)


Kaksisilma:
Looks OK code-wise, and I agree with the premise.

I will ultimately defer this one to Carl, however.

Sparrowbird611:
Seems overblown to me. There's far more powerful things available to normal crew and antags than GAS nab, and if the nab is a problem that should be getting nerfed instead of a band aid like this. The only major pain otherwise is that you cannot pull GAS so you just gotta grab them or tie them to a bed.

Kaksisilma:
> Seems overblown to me. There's far more powerful things available to normal crew and antags than GAS nab, and if the nab is a problem that should be getting nerfed instead of a band aid like this. The only major pain otherwise is that you cannot pull GAS so you just gotta grab them or tie them to a bed.

The PR premise is that there is a problem with GAS antags, and not with GAS nabs (Unless, only if by consequence), which I agree. 
For me, it usually results in the worst, most Low quality antagonists who typically power-game to hell and beyond and do things such as gun-wielding GAS. Serpentids have already little reason ICly to do antagonistic actions as per lore, and their vast scope of abilities incentivize murderbone-y playstyles.

We're not B*y, but there's a reason as to why they aren't allowed to be antags there even with a whitelist: The lore limitations and the less important but still relevant balance concerns. I personally don't think gutting their abilities is a worthwhile trade to make them antag-able, either. We should instead limit them from being antag while allowing them to keep their brisk attitude and their abilities, and this is exactly what this PR does.

Anton-Kr:
Personally against this because I think having murderous ambush snakes always be good is kinda lame

Don2137:
In a year of playing Hestia I've yet to see a single good GAS antag. It's literally always been powergamers and shitters.

The race was never designed for antagging, their design already allows them to self antag on the basis of some vague rules and lore, even on bay. Nabbing and such is the equivalent of magdumping someone, except a GAS antag can always come up with some bullshit excuse because apparently murder snakes with anxiety just kinda do that sometimes. Seriously, if you're in favor of them being antags, describe a single situation in which they made a round better.

SilverCelestine:
As the current GAS Maintainer (God help me), I fully support and approve the removal of them from the Antag system.

PR #1656, opened 2022-03-24
Commissary remap

Don2137:
## About The Pull Request
I swapped the Commissary with the Auxiliary Infirmary and made it somewhat bigger. Auxinf also gets a bootleg chem setup. That's about it.
## Why It's Good For The Game
RP areas are always good.
## Did You Test It?
Yes
## Authorship
Me
## Changelog

:cl:
maptweak: Remapped the Commissary and the Auxiliary Infirmary
/:cl:

### DIFF
--- a/commissary_finalest.dmm
+++ b/commissary_finalest.dmm
@@ -1,0 +1,59870 @@
+//MAP CONVERTED BY dmm2tgm.py THIS HEADER COMMENT PREVENTS RECONVERSION, DO NOT REMOVE
+"aa" = (
+/turf/space,
+/area/space)
+"ab" = (
+/obj/machinery/porta_turret/exterior/dagon,
+/turf/simulated/floor/airless,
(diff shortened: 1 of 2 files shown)


miniusAreas:
If this PR is about the commissary, then this is a downgrade to what is currently present. This moves the commissary on the other side of the ship from the merchant (funfact, the merchant SHOULD have access to those rooms to use them!), and the new setup for the southern airlock does not work, flat out (3 doors VS the 2 that should be present.)

Secondly, neither of these rooms see very much use. Auxinf is touched by maint looters at best, and the commissary is IDlocked to either Cargo, or the Merchant, and there's very little reason to mess with these because of this.

Don2137:
Merchant is never really gonna use the commissary, they already have their own ship they can customize so why go to the smaller, cramped room? It's also easier to steal from than their ship.

Don2137:
i hate github

PR #1662, opened 2022-03-25
Viewcones

ProbablyCarl:
Pulls viewcones from: https://github.com/Baystation12/Baystation12/pull/21764

That's about it. Tested it and it functions. Thought it'd be worth tossing it up as a PR to gauge reactions.


Jeff-jefferson-jeffery:
We're lifeweb now

ProbablyCarl:
Reopened here: https://github.com/BoHBranch/BoH-Bay/pull/1663

PR #1664, opened 2022-03-25
Adds Laser Carbines, Combat Knives and inducers to antag uplink

Pariah919:
## About The Pull Request
Laser carbines, combat knives and Inducers added
~~i'm dumb and an idiot so~~ these are costed based off the CRETH they cost.
Combat knives being equivalent to 
## Why It's Good For The Game
Laser carbines being accessible to traitors make sense, I asked around and people said it should cost a bit less than an STS, it has hitscan and burn damage but less DPS.
Combat knives make sense, they're a bit more than the switch blade.
Inducers can be useful, and they're 20 TC overall. That's FOUR syringes.
## Did You Test It?
Not sure how to test uplink stuff, if someone could give me some help with that, ping me on discord.
## Authorship
MJP#4043, Dudeman100- Ckey
## Changelog

:cl:
tweak: Added laser carbines (68tc), Inducers (20tc), Combat knives (5tc) to antag uplinks.
/:cl:

### DIFF
--- a/code/datums/uplink/devices and tools.dm
+++ b/code/datums/uplink/devices and tools.dm
@@ -164,3 +164,9 @@
 	item_cost = 4
 	path = /obj/item/clothing/gloves/nabber
 
+/datum/uplink_item/item/tools/inducer
+	name = "Inducer"
+	desc = "Use this on powered equipment to put power back into it, be it a laser gun or some APC on the road."
+	item_cost = 20
+	path = /obj/item/inducer
+
--- a/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
+++ b/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
@@ -74,6 +74,12 @@
 	item_cost = 48
 	path = /obj/item/weapon/gun/projectile/automatic/merc_smg
 
+/datum/uplink_item/item/visible_weapons/lasercarbine
+	name = "Laser Carbine"
+	desc = "A common laser carbine with only a lethal mode, and nothing else. Comes charged."
+	item_cost = 68
+	path = /obj/item/weapon/gun/energy/laser
+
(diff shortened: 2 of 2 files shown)


Yurek1:
> 

i think this is fine but combat knives don't seem like something worth adding, since they can be printed from any hacked autolathe

Rafflesia7793:
> > 
> 
> i think this is fine but combat knives don't seem like something worth adding, since they can be printed from any hacked autolathe

Yeah but that requires to you break into those departments to use their autolathe. If we're worried about bloating the uplink, there's TONS of dumb stuff with zero use that can be removed.

Yurek1:
> 

almost every department has a lathe, with the exception of civilians, service, and exploration. i guess if you put the cost to something like 3TC it might be more worth it? anyone who's military can spawn with them in most scenarios, with few exceptions, so the only people who this will apply to is civilians

N8-Toe (approved):
code and concept look good

Pariah919:
>almost every department has a lathe, with the exception of civilians, service, and exploration. i guess if you put the cost to something like 3TC it might be more worth it? anyone who's military can spawn with them in most scenarios, with few exceptions, so the only people who this will apply to is civilians
>

Switchblades are like worse, in every scenario statistically outside of TC cost and fashion. I can get the idea though. This is moreso a 'If I need some sort of stabbing implement on the cheap, NOW.' Kind of thing. I could make it plasteel, however.

Yurek1:
> 

i wasn't talking about switchblades, but that would be neat. a new knife, traitor only, that does a little more damage? that would be cool and unique enough to be worthy to be put into the uplink. 
as it stands, the regular steel combat knife can be spawned with from the tactical gear loadout menu, or printed from literally any hacked autolathe

PR #1670, opened 2022-03-26
Buffs shields.

Rafflesia7793:
## About The Pull Request
When shields were first introduced and these values set, you could accurately fire longarms while holding one. This is no longer the case, and a result, they have fallen into disuse, only used for the rule of cool. This PR slightly buffs the combat shield's block chance, the amount of damage it is capable of blocking, and moves the shield's slowdown to match it's current capabilities.

## Why It's Good For The Game
Adds more options to the stale combat meta.

## Did You Test It?
Yes.
## Authorship
Rafflesia
## Changelog

### DIFF
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -100,7 +100,7 @@
 	throw_range = 3
 	w_class = ITEM_SIZE_HUGE
 	matter = list(MATERIAL_PLASTEEL = 8500)
-	max_block = 55
+	max_block = 50
 	can_block_lasers = TRUE
 	slowdown_general = 1.5
 
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -100,9 +100,9 @@
 	throw_range = 3
 	w_class = ITEM_SIZE_HUGE
 	matter = list(MATERIAL_PLASTEEL = 8500)
-	max_block = 50
+	max_block = 55
 	can_block_lasers = TRUE
-	slowdown_general = 1
+	slowdown_general = 1.5
 
 /obj/item/weapon/shield/buckler
 	name = "buckler"
--- a/code/game/objects/items/weapons/shields.dm
+++ b/code/game/objects/items/weapons/shields.dm
@@ -100,7 +100,7 @@
 	throw_range = 3
 	w_class = ITEM_SIZE_HUGE
 	matter = list(MATERIAL_PLASTEEL = 8500)
-	max_block = 55
+	max_block = 50
 	can_block_lasers = TRUE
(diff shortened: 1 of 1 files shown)


Anton-Kr:
Seems like an interesting change that should make SMGs etc a bit more used/useful. Why not!

Anton-Kr (approved):

BurpleBineapple:
> Seems like an interesting change that should make SMGs etc a bit more used/useful. Why not!

~~SMGs are one handed, at least the 7mm ones available to security. This however will allow you to fire bulldogs and LMGs such with a shield which I'm not sure is a great idea.~~ Disregard previous, I'm tired. But anyway, the plasteel combat shield can currently stop projectiles up to rifles. The damage var is calculated only from the raw damage of the bullet, not armor penetration for ballistics.

After checking again, for lasers it *only* checks armor penetration. This still means it currently blocks all regular laser rifles excluding the x-ray rifle and laser DMR

Rafflesia7793:
>This however will allow you to fire bulldogs and LMGs such with a shield which I'm not sure is a great idea. 

This PR makes zero changes to what weapons can and cannot be fired accurately with the shield in offhand.

Anton-Kr:
I'll test this a bit further.

Anton-Kr (changes requested):
Following discussion, this seems a bit heavyhanded. I am fine with removing the slowdown, **OR** upping the block chance rate. I want the max_block however to remain where it is, to make more weapons have the ability to pen the shield, because at the moment even the Skrell railgun wouldn't penetrate the shield.

Rafflesia7793:
> Following discussion, this seems a bit heavyhanded. I am fine with removing the slowdown, **OR** upping the block chance rate. I want the max_block however to remain where it is, to make more weapons have the ability to pen the shield, because at the moment even the Skrell railgun wouldn't penetrate the shield.

Considering it's a defensive implement I'll go with making the block chance not nearly 50/50. In the future I'd like to make it so that aiming for feet/hands/head mitigates the shield entirely, hopefully when that happens, we can consider improving the slowdown.

Anton-Kr:
Apologies if my comment was ambiguous, but I'd still like `max_block `at 50 to allow high caliber weapons outside of just the AMR to penetrate. 

Thanks!

Rafflesia7793:
> Apologies if my comment was ambiguous, but I'd still like `max_block `at 50 to allow high caliber weapons outside of just the AMR to penetrate.
> 
> Thanks!

Dafuq? That was unintentional change. I'll fix it again. Probably something with github. Edit: Oh I know, I pulled when I should've pushed.

Rafflesia7793:
Should be good to go now.

Anton-Kr (approved):

PR #1690, opened 2022-03-28
rnd/ce buff (advanced tools are back, kind of)

lolman360:
## About The Pull Request

advanced tools. they're more compact than regular tools, look cooler, and in the case of jaws of life, can remove both braces and pry open powered airlocks (after a 7.5s delay)

CE starts with a belt, one starts in CE's locker, and sci can print them. no special materials required yet, but this is subject to change.

advanced welder itemsize large >> normal

## Why It's Good For The Game

bay removed `toolspeed` so until i totally rework the tool system (port neb's or something) the only advantage you get is compactness

## Did You Test It?
not yet

## changelog

:cl:
rscadd: advanced tools
balance: advanced welder now useable
/:cl:


code/game/objects/items/weapons/tools/screwdriver.dm
```diff
@@ -36,4 +36,44 @@
...
+/obj/item/weapon/screwdriver/power
+	name = "hand drill"
+	desc = "A simple powered hand drill. It's fitted with a screw bit."
+	icon_state = "drill_screw"
+	item_state = "drill"
+	matter = list(MATERIAL_STEEL = 500, MATERIAL_ALUMINIUM = 500, MATERIAL_PLASTIC = 500)
+	origin_tech = list(TECH_MATERIAL = 2, TECH_ENGINEERING = 2)
+	slot_flags = SLOT_BELT
+	force = 8
+	w_class = ITEMSIZE_SMALL
```
Anton-Kr:
```suggestion
	w_class = ITEM_SIZE_SMALL
```

Anton-Kr (changes requested):

code/datums/outfits/jobs/engineering.dm
```diff
@@ -14,6 +14,7 @@
 	name = OUTFIT_JOB_NAME("Chief engineer")
 	head = /obj/item/clothing/head/hardhat/white
 	uniform = /obj/item/clothing/under/rank/chief_engineer
+	belt = /obj/item/weapon/storage/belt/utility/chief
```
Anton-Kr, line 17: The toolbelt isn't placed in CE's locker.

code/game/machinery/doors/airlock.dm
```diff
@@ -1155,6 +1155,18 @@ About the new airlock wires panel:
 				else
 					to_chat(user, "<span class='warning'>You need to be wielding \the [C] to do that.</span>")
 
+	else if(istype(C, /obj/item/weapon/crowbar/brace_jack/jaws))
+		if(locked)
+			to_chat(user, "<span class='notice'>The airlock's bolts prevent it from being forced.</span>")
+		else if(!welded && !operating)
+			user.visible_message(SPAN_WARNING("[user.name] wedges the [C] into [src] and starts forcing it open!"), SPAN_WARNING("You wedge the [C] into [src] and start forcing it open!"))
```
Anton-Kr, line 1162: I'd prefer this to be destructive in nature (think EMAG) and for it to destroy the airlock.

code/game/machinery/doors/airlock.dm
```diff
@@ -1155,6 +1155,18 @@ About the new airlock wires panel:
...
+			if(!do_after(user, 7.5 SECONDS))
+				return
+			if(density)
```
Anton-Kr: airlock.dm provides defines here like AIRLOCK_CLOSED and AIRLOCK_OPEN. Please use those instead.

code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
```diff
@@ -21,7 +21,7 @@
 		/obj/item/device/multitool,
 		/obj/item/device/flash,
 		/obj/item/taperoll/engineering,
-		/obj/item/weapon/crowbar/brace_jack
+		/obj/item/weapon/storage/belt/utility/chief
```
Anton-Kr, line 24: Believe the actual locker is defined in Torch files and not here. May be wrong.

Anton-Kr (changes requested):

Anton-Kr (changes requested):
RE: Sprites. Both sprites have a very harsh single color outline that isn't very consistent with our (already inconsistent) art style. Could you remove the outline? If not @N8-Toe may be able to assist you.

code/game/machinery/doors/airlock.dm
lolman360: ideally it's an engineering tool so i'd like to have it not break airlocks

code/datums/outfits/jobs/engineering.dm
lolman360: `CE starts with a belt`

code/game/machinery/doors/airlock.dm
BurpleBineapple: At the moment, this is just acting as a normal crowbar. Destructive might be interesting if it's able to pry open bolted doors.

BurpleBineapple (commented):
I'd say mostly, I agree with ant here. Code-wise everything works fine, tested that on my local branch.

I don't feel particularly strongly about making the jaws of life break airlocks when prying them open. I think it would be interesting if it could do this to powered airlocks. At the moment, it's just a normal crowbar which is also fine.

code/game/machinery/doors/airlock.dm
lolman360: > At the moment, this is just acting as a normal crowbar. Destructive might be interesting if it's able to pry open bolted doors.

harm intent is needed to pry open powered doors, i might consider destructive if it can pry open bolted ones too at a longer delay

Anton-Kr:
@BurpleBineapple it overrides crowbar code for airlocks thus does in fact open powered doors.

PR #1703, opened 2022-03-29
Neutron Grenade Rebalance

ProbablyCarl:
- - -
Balance:
 - Drops the initial radiation from neutron grenades a significant amount, instead replacing most of it with the fact that it can contaminate areas. it's now a proper dirty bomb. 65 radiation at the epicenter, with walls and terrain taken into account.
- - -

### DIFF
--- a/code/game/objects/items/weapons/grenades/neutron.dm
+++ b/code/game/objects/items/weapons/grenades/neutron.dm
@@ -9,7 +9,7 @@
 	origin_tech = list(TECH_BLUESPACE = 5, TECH_MAGNET = 4, TECH_ENGINEERING = 5)
 	arm_sound = 'sound/effects/scanbeep.ogg'
 	var/explode_at
-	var/radiation = 64//with how this is set up, and how rads caps-out anyway, 500 is what it'll jump to with how I have this.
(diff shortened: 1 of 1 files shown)


anconfuzedrock:
How long do the rads last / can you get rid of them? I'm not sure we want to make lasily tossing an SM grenade into medbay even funnier.

ProbablyCarl:
> How long do the rads last / can you get rid of them? I'm not sure we want to make lasily tossing an SM grenade into medbay even funnier.

This is a significant nerf. Standard SM output for rads is roughly 500 I believe.
Rads dissipate fairly quickly, really. It's about one per second, IIRC. Just about, anyhow. So you'd need about a minute and a half at most for it to clear from Medical, and that'd be at only the impact point. Walls, doors and other such things heavily dampen the effect, as it takes that into account.

As is, without this nerf, it dumps 500 rads(cap for the object) into anyone within LoS, which was intentional to give it its gimmick. Considering rads are very easy to deal with.
This drops that to 64 for LoS, and irradiates objects as described above.

anconfuzedrock:
okie then, that sounds like it wouldn't permanently ruin medbay.

PR #1709, opened 2022-03-30
[WIP] Cyberpunk 2556: The Augment Update

BurpleBineapple:
## About The Pull Request
A large update that aims to port over the changes to the augment system from Baystation12 as well as fix simple augments from being unable to redeploy once retracted.

## Why It's Good For The Game
This PR fixes some issues with armblade type objects which causes them to be set to null once retracted. This is no longer the case.

This update also ports over many changes to the system from baystation as well as merging item and simple into one 'item' file. It also cleans up the formatting in several places.

Augments are now installed into augment slots rather than the limb. Each body part has a single augment slot. This prevents several of one type of augment being installed Example, you can't have both an armblade and a gun in your arm, but you can have cyberclaws in your hand and a gun in your arm.

This PR also updates HUD augments to now work and will be adding binocular augments, welding goggle augments, pop-out guns for arms, and a powerfist augment from baystation12.

Other more specialized augments may be added in future PRs. Many of these items will be available to traitors via a new 'augmentation' section in the uplink where autodoc-type items can be purchased to install the part into your limbs.

## Did You Test It?
WIP but yes.
## Authorship
PurplePineapple
Spookerton - Baystation12
Juxtaposed - Baystation12
SierraKomodo - Baystation12
Azlan - Baystation12
## Changelog

:cl:
/:cl:


BurpleBineapple:
PR is now working. Will be keeping it as a draft until I add some antagonist objects.

PR #1720, opened 2022-03-31
Banana pie redux

Thescri:
## About The Pull Request
the return of banana cream pie instead of default snack with banana in

## Why It's Good For The Game
We should at least comment out the pie recipe, otherwise. Nobody likes banana-flavoured default icon snacks.

## Did You Test It?
Recipe tested successfully

## Authorship
Bunnyears

## Changelog

:cl:
tweak: changes banana cream pie desc
bugfix: no more flickering magenta icon in the kitchen
/:cl:

### DIFF
--- a/code/modules/cooking/recipes/recipes_oven.dm
+++ b/code/modules/cooking/recipes/recipes_oven.dm
@@ -160,7 +160,7 @@
 	fruit = list("banana" = 1)
 	reagents = list(/datum/reagent/sugar = 5)
 	items = list(/obj/item/weapon/reagent_containers/food/snacks/sliceable/flatdough)
-	result = /obj/item/weapon/reagent_containers/food/snacks/pie
+	result = /obj/item/weapon/reagent_containers/food/snacks/bananapie
 
 /datum/recipe/cherrypie
 	appliance = OVEN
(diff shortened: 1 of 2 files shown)


code/modules/reagents/reagent_containers/food/snacks.dm
```diff
@@ -880,7 +880,7 @@
 
 /obj/item/weapon/reagent_containers/food/snacks/bananapie
```
BurpleBineapple: After looking through the code, this is actually a bug. Instead of changing the recipe's path, please change this to

```suggestion
/obj/item/weapon/reagent_containers/food/snacks/pie
```

This will fix the issue with pies being unable to splat as well.

code/modules/cooking/recipes/recipes_oven.dm
```diff
@@ -160,7 +160,7 @@
 	fruit = list("banana" = 1)
 	reagents = list(/datum/reagent/sugar = 5)
 	items = list(/obj/item/weapon/reagent_containers/food/snacks/sliceable/flatdough)
-	result = /obj/item/weapon/reagent_containers/food/snacks/pie
+	result = /obj/item/weapon/reagent_containers/food/snacks/bananapie
```
BurpleBineapple:
```suggestion
	result = /obj/item/weapon/reagent_containers/food/snacks/pie
```
Keep this how it is.

BurpleBineapple (changes requested):

code/modules/reagents/reagent_containers/food/snacks.dm
Thescri: Good call. Swapped them around so it should be better now.

BurpleBineapple (approved):

PR #1724, opened 2022-04-01
plasmaman burning no longer double-dips burn damage

lolman360:
## About The Pull Request

plasmamen are both set on fire and take constant burn damage which is absurd

this puts them more in line with tg plasmemes
:cl:
balance: plasmaman burn less
/:cl:


BurpleBineapple (approved):
Seems fine to me from a code standpoint. I agree with the concept as well, personally. Phorosians need some slack.

PR #1725, opened 2022-04-01
Replaces normal guns with smartguns

AMcilraith:
## About The Pull Request
Simply replaces **some** of the normal guns with smartguns, because it makes it more sci-fi like
## Why It's Good For The Game
It's good for the game because smartguns require you to swipe your ID but also are lasers making them extra deadly, sec now gets a bunch in their armoury, along with all the command staff.
## Did You Test It?
Yes
## Authorship
Dr.Farson
## Changelog

:cl:AMcIlraith
maptweak: replaces guns with better smartguns
/:cl:


BurpleBineapple:
Looks at the diffs, you've may have made a mistake when creating this PR. You're currently deleting a large portion of the map.

AMcilraith:
> Reference

Yeah, This seems to have fixed that

ProbablyCarl:
Stealth nerf to the Sec armory, by removing the plasma projector?

BurpleBineapple:
I recommend checking #1728 before merging this. If compacts are given the recharge back, I would prefer those put on the bridge instead of full sized smartguns.

Security's equipment is also fine for the moment. Full sized smartguns are better as an item that requires production from my experience playing brig chief. They're very powerful guns against internal threats.

AMcilraith:
> I recommend checking #1728 before merging this. If compacts are given the recharge back, I would prefer those put on the bridge instead of full sized smartguns.
> 
> Security's equipment is also fine for the moment. Full sized smartguns are better as an item that requires production from my experience playing brig chief. They're very powerful guns against internal threats.

Alright, fair enough, I will note that, and fix it, how does two full-sized and re-add the plasma projector sound?

AMcilraith:
> Stealth nerf to the Sec armory, by removing the plasma projector?

Plasma projector re-added, didn't realise I deleted it

PR #1726, opened 2022-04-01
Removes Skrell Railgun Hardstun

BurpleBineapple:
## About The Pull Request
This PR aims to remove the hardstun from the Skrellian 6 shot railgun.

## Why It's Good For The Game
At the moment, this is the fastest firing railgun in the game and, on top of being able to hardstun, it can lock a person in hardstun due to it firing faster than the stun lasts.

The damage and AP remain the same. However, the Skrellian railgun is no longer an instant win gun.
## Did You Test It?
Yes.

## Authorship
PurplePIneapple#0001
## Changelog

:cl:
tweak: Skrellian railgun now fires a new subtype projectile which has no hardstun
/:cl:


Rafflesia7793:
Based, if ppl feel like the Skrell are lacking in firepower down the road, we can always give them something else to boost them up that isn't stunlocking. Stun in general is a mechanic that should be used extremely sparingly, and most of the offships and onships have some form of it.

It works best when it's used for LTL weapons, but stun on a weapon that's incredibly lethal is just shortsighted to begin with. It doesn't create enjoyable combat for both sides.

PR #1594, opened 2022-03-19
[FUCK] Ports Taur Bodies and Weaversilk from Virgo

ShadeAware:
## About The Pull Request
As stated above, Genemodders now have access to tauric bodies. What this essentially means is, they can become centaurs - of various walks, including horses, snakes and spiders. These would be considered **Aberration Modders** in-lore. Additionally, these can only be used by the Aberration subvariant of genemodders.
image
Additionally, this also includes a new trait: **Weaver.** Weavers can make things from silk, and this silk can have its color changed; it is available to Genemodders via the Positive trait "Weaver" and is available to the Vass by default. I had to do some hacky bullshit to get it to work so there may need to be some bug squashing down the line, but it _should_ work and doesn't break anything major.
image

## Why It's Good For The Game
More customization is always good! Brings in a new trait that Genemodders might enjoy. Makes the Vass feel special.
image

## Did You Test It?
Yes, extensively. 
## Authorship
ShadeAware
## Changelog

:cl:
rscadd: Weaver trait and structures.
rscadd: Genemodders can now access a number of tauric bodies.
/:cl:

PR #1743, opened 2022-07-25
Bye bye taurs

gilvian:
should revert [pr ](https://github.com/BoHBranch/BoH-Bay/pull/1594)
Here are the tags supported by changelog:
* rscdel - horse dies


Cupax3:
Does it compile without issues and does it break anything?

gilvian:
ummm, i dont know how to use github

